PR #1, opened 2023-06-30
Agile Coretime

gavofyork:
This proposes a periodic, sale-based method for assigning Polkadot Coretime. The method takes into account the need for long-term capital expenditure planning for teams building on Polkadot, yet also provides a means to allow Polkadot to capture long-term value in the resource which it sells. It supports the possibility of building secondary markets to make resource allocation more efficient and largely avoids the need for parameterisation.

Implementation: https://github.com/paritytech/substrate/pull/14568


RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,265 @@
...
+
+## Summary
+
+This proposes a periodic, sale-based method for assigning and Polkadot Coretime. The method takes into account the need for long-term capital expenditure planning for teams building on Polkadot, yet also provides a means to allow Polkadot to capture long-term value in the resource which it sells. It supports the possibility of building secondary markets to make resource allocation more efficient and largely avoids the need for parameterisation.
+
+## Motivation
+
+### Present System
+
+The present system of allocating time for parachains on the cores of the Polkadot Ubiquitous Computer (aka "Polkadot") is through a process known as *slot auctions*. These are on-chain candle auctions which proceed for several days and result in a core being assigned to a single parachain for 6six months at a time up to 18 months in advance. Practically speaking, we only see two year periods being bid upon and leased.
```
lucasvo, line 21: Typo: 6six

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,265 @@
...
+
+### Specific functions of the Broker-chain
+
+Several functions of the Broker-chain SHALL be exposed through dispatchables and/or a `nonfungible` trait implementation integrated into XCM:
+
+#### 1. Transfer of ownership
+
+A `transfer(region: RegionId, new_owner: AccountId)` dispatchable SHALL have the effect of altering the current `owner` field of `region` in the `Regions` map from the signed origin to `new_owner`.
+
+An implementation of the `nonfungible` trait SHOULD include equivalent functionality. `RegionIndex` SHOULD be used for the `AssetInstance` value.
```
lucasvo, line 169: I don’t believe `RegionIndex` is defined and I’m not quite sure what is meant here. Is this `RegionId`

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,265 @@
...
+}
+struct RegionRecord {
+    owner: AccountId,
+    end: Timeslice,
+    price: Option<Balance>,
+    allocation: Option<Vec<ParaId>>, // begins set to `None`
+}
+```
+
+Notably, if a region is split or transferred, then the `price` is reset to `None`.
```
lucasvo, line 155: I think it would be easier to understand if you explicitly say that splitting a region removes the ability to renew at the same price which I think you are implicitly saying by specifying splitting sets the price to “none”. This also implies though that “split cores” are not eligible for priority renewal, correct? You also don’t seem to mention anything of the fact that existing owners of cores should be able to get priority renewal. Is that correct?

lucasvo (commented):
Initial review with minor comments and one question for clarity. I think the mechanism is simple and a much better model than the current slots. I will take another pass once we’ve discussed this a bit more.

RFC-0001-Agile Coretime.md
gavofyork: This is written above:

> Notably, if a region is split or transferred, then the `price` is reset to `None`.

Not enough?

Also, this is about split *regions*, i.e. taking the month-long piece and splitting in into smaller pieces.

lucasvo: I don’t understand if there is a specific reason that the price needs to be set to none. Does that mean a split region can’t be renewed the same way a “full core” region would?

gavofyork: Yeah - because it would then have two owners - which one could "renew"?

lucasvo: My comment initially was to clarify that split cores shouldn’t be able to get priority renewal. Carefully reading the spec does imply that but I think making this more explicit would be helpful for people to understand.

> I think it's ok when the core would be used consistently by the same paras in the same way from month to month, but it doesn't make sense when they're being carved up and, presumably, traded.

I think this does not go into the initial spec. It would should be possible to offer to buy split cores at some point at which point this can be added. I doubt the demand for them will be particularly high today and it shouldn’t delay a first implementation.

lucasvo:
I believe this RFC makes transfer of regions lose the priority on core renewal but changing the allocation of the slot from one parachain to another does not. This would mean the system can quite easily be gamed to transfer ownership of a region without it being prevented from being renewed (for example by holding the region in a pure proxy and simply transferring ownership). Wouldn’t we want to tie the ability to renew to the paraid that the core is running and not which account controls the core?

rphmeier (commented):
## Regions Architecture

The regions here have a number of similarities with the regions I've proposed in #3 but lack a couple significant details which will be important for implementing elastic scaling and other future mechanisms.

My proposal was focused on relay-chain coretime scheduling and is actually a replacement for all existing scheduling logic in the relay-chain. In my opinion, the mechanisms on the **broker-chain** should match the mechanisms on the **relay-chain** closely in order to avoid friction between those components.

Scheduling on the relay-chain, looking forward, needs to solve a few key problems:
  1. **Gracefully handle tens of thousands of parachains without significant runtime scheduling overhead**.
  1. **In the case that a parachain has scheduled assignments on several cores, the mapping between upcoming blocks on the parachain and specific cores must be unambiguous in the near-term future to the validators working on those cores**
  1. **When cores are highly shared, the time it takes to make up for a missed opportunity should be minimally dependent on the number of other applications also scheduled on the core**
  1. **Accommodate a variety of different scheduling frequencies and overlapping durations, all on the same core**

(more detailed description of all of the above and more in #3)

The regions RFC is my attempt to solve all these problems. When we discussed offline, @gavofyork raised the point that handling splitting/transferring on the relay-chain is likely to incur too much load, which is fair. However, the relay-chain `Region` primitive itself is still important. The broker-chain should be able to use its own primitives, but they ought to be compatible with the direction of relay-chain scheduling.

To solve (1), the proposal focuses on having parachain candidates tagged with a _deterministic and immutable_ region identifier which is submitted along with the candidate by the relay-chain block author. The relay-chain logic needs only lazily check that the region is in surplus and the parachain assignment is correct, which is a single load, modify, store operation per core per relay-chain block. This way, all scheduling overhead is pushed to the node-side. 

To solve (2), we can build upon the solution to (1). In the future, this region identifier for a particular candidate may be included inside the `CandidateReceipt` by the collator, as collator-selection algorithms working across many regions must already figure out how to utilize their regions and asking validators to re-run this allocation logic with less information is a redundancy that can be avoided. This solves the elastic scaling candidate-group problem, as validators will know unequivocally which backing group is intended to work on the candidate.

To solve (3) and (4), the regions proposal schedules core-time somewhat _probabilistically_. P2P network systems experience variance in practice, and the time it takes to back a block or make its data available does not always fall within prescribed bounds. Stated otherwise - if Polkadot were to set time limits on backing and availability timeouts such that they were always met, or were met 99.9% of the time, those bounds would likely be too conservative and we'd be leaving significant performance on the table. By giving regions a _single_ assignee, the probabilistic scheduling allows for cores and parachains to "make up" recently missed opportunities by accepting more than one candidate at a time, up to configured per-core and per-relay-chain-block maximum to avoid massive per-block loads. Chains can never access more core-time than they've been allocated in total. With this solution, even with 1000 chains scheduled on a single core with varying frequencies, they cause minimal friction on each others' timing and system load is both predictable and capped when averaged out over any period longer than a minute or two.

## Applying Regions to this RFC

I suggest that my regions proposal be altered, removing the ability to split, transfer, and reassign regions in an unpermissioned way on the Polkadot relay-chain. 

Instead, these actions would become permissioned, with the permission being held by the broker-chain. The broker-chain would then 'blit' the data structures it manages onto regions in the relay-chain to manage scheduling. This can be done with a single XCM to create the regions on the relay-chain, and the scheduling logic there would handle the rest.

The region records described in this RFC will be compatible, in that they could be transformed into single-assignee, frequency-based regions when blitting them up to the relay-chain, but I suggest we outline that intention in this RFC to commit to that as a plan. The regions in this RFC also have a few properties I'll comment on:
  1. The `RegionId` is not unique - every `BULK_PERIOD`, `RegionId`s will be reused, as they are dependent on only the core Index and the timeslice within the `BULK_PERIOD`. That may make it harder for logic living in other chains to do bookkeeping about which regions they own.
  3. The `allocation` containing a `Vec<ParaId>` rather than having multiple regions, each with their own single `ParaId` may discourage region-sharing among chains, as either they _all_ have to renew together or _none_ of them can. In my opinion, this is likely to take away from the value proposition of sharing regions altogether, as all parachains will want to stay on the `RENEWAL_PRICE_CAP` curve but some will be chained to sinking partners. That said, it does also give parachains very strong incentives to help their region companions survive. Let's discuss this property & alternatives.

Tight integration with the regions RFC would make updating the broker parachain to elastic scaling, or adding other mechanisms for accessing core-time technically trivial, economic design notwithstanding.

Since full implementation of the regions RFC will likely take a while, it'd probably make most sense to include a new call on the relay-chain that the broker can invoke via XCM::Transact: **BlitBrokerRegion**. This would take the Broker region format as an argument and transform it into whatever the scheduling mechanism that the relay chain currently uses is - whether that's the existing scheduling infrastructure quickly adapted for the purpose, or the new regions architecture when that lands.

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,291 @@
...
+
+It is also expected to receive information from the Relay-chain on total revenue from instantaneous coretime sales on a per-block basis.
+
+### Bulk Sales
+
+There is a periodic sale of Coretime happening at a period of `BULK_PERIOD`.  A number of instances of Coretime spanning from `LEADIN_PERIOD` in the future to `LEADIN_PERIOD + BULK_PERIOD` in the future of a single Polkadot UC Core is offered by the Polkadot System at a fixed price.
+
+These instances which are owned by purchaser are called *regions*. This sale happens on the Broker-chain. Regions are quantized into atomic periods of `TIMESLICE`, into which `BULK_PERIOD` divides a whole number of times. The `Timeslice` type is a `u16` which can be multiplied by `TIMESLICE` to give a `BlockNumber` value indicating the same period in times of (Relay-chain) blocks.
+
+The Broker-chain aims to sell some `BULK_TARGET` of Cores, up to some `BULK_LIMIT`. It makes this sale in a single batch `LEADIN_PERIOD` prior to the beginning of the period being sold. The Broker chain publishes a price for this batch of sales for the `BULK_TARGET` period prior to the sale execution.
```
BradleyOlson64, line 132: I find the last sentence here confusing. Particularly, "sales for the `BULK_TARGET` period". Mind clarifying?

rphmeier, line 132: Is there a formula or parameters for how the price adjusts from period to period, based on the actual, limit, and bulk?

I imagine something like a minimum price & an adjustment %. The algorithm should be specified in the RFC to explain how it meets goal (1)

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,291 @@
...
+In both cases, the `price` field SHALL become `None`.
+
+#### 2. Split into segments
+
+A `split(region: RegionId, pivot: Timeslice)` dispatchable SHALL have the effect of mutating the Regions entry of key `region` so that the `end` field becomes `pivot` and create a new Regions entry of the same `core` but a `begin` equal to `pivot` whose `RegionRecord` has the same `owner`, `end` and `allocation` fields as the origin value of `region`.
+
+`price` in both records is/becomes `None`.
+
+Also:
+- `owner` field of `region` must the equal to the Signed origin.
```
joepetrowski, line 185: Does it need to be `Signed`, or just something that maps to an `AccountId` (as in `new_owner` within `transfer`)? For example, it could be owned by a collective or parachain. Or it could be signed with a session key corresponding to a collator (in order to claim allocation).

joepetrowski: Perhaps session key doesn't work if it needs to be possible to authorise transfer.

RFC-0001-Agile Coretime.md
gavofyork: The price-change mechanism is possibly out of scope of this RFC. One possibility is for it to have `MAX_PRICE_INCREASE` and `MAX_PRICE_DECREASE` fractions, then a formula:

```
NEW_PRICE := IF CORES_SOLD < BULK_TARGET THEN
    RETURN OLD_PRICE - OLD_PRICE * MAX_PRICE_DECREASE * CORES_SOLD / BULK_TARGET
ELSE
    RETURN OLD_PRICE + OLD_PRICE * MAX_PRICE_INCREASE * (CORES_SOLD - BULK_TARGET) / (BULK_LIMIT - BULK_TARGET)
END IF
```

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Motivation
+
+### Present System
+
+The present system of allocating time for parachains on the cores of the Polkadot Ubiquitous Computer (aka "Polkadot") is through a process known as *slot auctions*. These are on-chain candle auctions which proceed for several days and result in a core being assigned to a single parachain for six months at a time up to 18 months in advance. Practically speaking, we only see two year periods being bid upon and leased.
+
+Funds behind the bids made in the slot auctions are merely locked, not consumed or paid and become unlocked and returned to the bidder on expirt of the lease period. A means of sharing the deposit trustlessly known as a *crowdloan* is available allowing token holders to contribute to the overall deposit of a chain without any counterparty risk.
```
lolmcshizz, line 23: > Funds behind the bids made in the slot auctions are merely locked, not consumed or paid

This wording and multiple other references to sales/ purchase suggest that Bulk coretime will be *paid for* in DOT, in contrast to currently where it's simply locked and the "cost" is the opportunity cost of not staking - it should be explicit if this is the case.

It would also be useful to understand - if this is the case - where those DOT are sent, i.e. who is paid? Is it validators? Is it burned? If validators, would high reward rates through demand for blockspace impact inflation that currently forms the vast majority of their revenue?

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,291 @@
...
+
+There is no ability to determine capital requirements for hosting a parachain beyond two years from the point of its initial deployment onto Polkadot. While it would be unreasonable to have perfect and indefinite cost predictions for any real-world platform, not having any clarity whatsoever beyond "market rates" two years hence can be a very off-putting prospect for teams to buy into.
+
+However, quite possibly the most substantial problem is both a perceived and often real high barrier to entry of the Polkadot ecosystem. By forcing innovators to either raise 7-figure sums through investors or appeal to the wider token-holding community, Polkadot makes it essentially difficult for a small band of innovators from deploying their technology into Polkadot. While not being actually permissioned, it is also far from the barrierless, permissionless ideal which an innovation platform such as Polkadot should be striving for.
```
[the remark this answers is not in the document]
lucasvo: I would agree and I believe this is probably for committed parachain builders an important part. I would add this to the problem statement.

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,291 @@
...
+There are six main requirements:
+
+1. The solution MUST provide an acceptable value-capture mechanism for the Polkadot network.
+2. The solution SHOULD allow parachains and other projects deployed on to the Polkadot UC to make long-term capital expenditure predictions.
+3. The solution SHOULD minimize the barriers to entry in the ecosystem.
+4. The solution SHOULD maximize the value which the Polkadot UC provides by allocating its limited resources optimally.
+5. The design MUST work with a limited set of resources (cores on the Polkadot UC) whose properties and number may evolve over time.
+6. The design MUST avoid creating additional dependency on functionality which the Relay-chain need not strictly provide for the delivery of the Polkadot UC. This includes any dependency on the Relay-chain hosting a DOT token.
+
+Furthermore, the design SHOULD be implementable and deployable in a timely fashion; three months from the acceptance of this RFC would seem reasonable.
```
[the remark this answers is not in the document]
lucasvo: Wouldn't the fellowship be able to whitelist this upgrade?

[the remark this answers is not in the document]
joepetrowski: Yes but the track is still 28 days. And even though it is less restrictive to pass earlier on that track, a lot of parachain teams have expressed that they prefer a set block number (i.e. `At` over `After`) for runtime upgrades so that they can prepare for any breaking changes.

lucasvo:
@gavofyork you write:
> The point of renewals isn't to attempt to give as many entities as possible a discount for the next month: it's to ensure that committed teams get some guarantees about price for predicting future costs.

I think this point perhaps wasn't clear enough. Thanks for clarifying that this is implemented by having `renew` always `allocate` the core to the same ParaId.  I would add a comment to `renew` section that points out that this is a specific design goal or highlight this in the problem statement.

RFC-0001-Agile Coretime.md
gavofyork: > Bulk coretime will be paid for in DOT

That's right.

Any DOT recuperated for sales of system resources (Coretime, in this case) would by default be placed in the treasury. Governance would be able to determine what to do with it.

RFC-0001-Agile Coretime.md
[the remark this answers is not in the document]
gavofyork: > There is no ability to determine capital requirements for hosting a parachain beyond two years from the point of its initial deployment onto Polkadot

It is not that slots may become unavailable. It is that the price may become too high. Thus this statement on the paragraph above is actually all that needs to be said.

RFC-0001-Agile Coretime.md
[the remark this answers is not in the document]
gavofyork: We'll see. As long as the Relay-chain support for core-assignment exists, then this really shouldn't be a big change.

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,291 @@
...
+### Broker-chain
+
+The Broker-chain is a new system parachain. It has the responsibility of providing the Relay-chain via UMP with scheduling information of:
+
+- How many cores which should be made available during the next session.
+- Which cores should be allocated to which para IDs.
+
+Any cores left unallocated are assumed by the Broker-chain to be used for Instantaneous Coretime sales.
+
+It is also expected to receive information from the Relay-chain on total revenue from instantaneous coretime sales on a per-block basis.
```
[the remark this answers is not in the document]
gavofyork: It's really outside of the scope of this RFC and is only provided as a hint to illustrate a full solution.

The Broker would be on a dedicated core, yes.

RFC-0001-Agile Coretime.md
gavofyork: Indeed. The `owner` must be strictly equivalent to be the inner of the `Signed` origin.

RFC-0001-Agile Coretime.md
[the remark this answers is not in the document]
KiChjang: Perhaps also worth bringing up is the non-permanence of a slot/coretime. This is certainly a comment that I heard from some prospective builders which predates our current arrangement with slot auctions: they are essentially put off by the fact that they are not guaranteed to have their blocks processed indefinitely once they've purchased their slot/coretime.

May not be a legitimate concern, but re-iterating the reasoning/justification here would better assuage these concerns.

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,291 @@
...
+}
+struct RegionRecord {
+    owner: AccountId,
+    end: Timeslice,
+    price: Option<Balance>,
+    allocation: Option<Vec<ParaId>>, // begins set to `None`
+}
+```
+
+This map functions essentially as a linked list. With one region's `end` field functioning as the next region's key's `begin` field. It is keyyed by the sale index in order to allow the following sale period's Coretime to be manipulated during the `LEADIN_PERIOD` prior to it becoming allocatable.
```
rphmeier, line 156:
```suggestion
This map functions essentially as a linked list. With one region's `end` field functioning as the next region's key's `begin` field. It is keyed by the sale index in order to allow the following sale period's Coretime to be manipulated during the `LEADIN_PERIOD` prior to it becoming allocatable.
```

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,291 @@
...
+type SaleIndex = u16;
+struct RegionId {
+    core: CoreIndex,  // A `u16`.
+    begin: Timeslice, // A `u16`.
+}
+struct RegionRecord {
+    owner: AccountId,
+    end: Timeslice,
+    price: Option<Balance>,
+    allocation: Option<Vec<ParaId>>, // begins set to `None`
```
rphmeier, line 152: What are the proposed semantics of this field? Are para-IDs ordered? Is there a limit on the length of the vector? Can a `ParaId` appear multiple times, e.g. `[A, A, B]` to accommodate a (2/3, 1/3) split between A and B?

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,291 @@
...
+
+An additional storage map is maintained to keep the "heads" of this linked list. It is called `NextRegion` and it maps `CoreIndex` to `Timeslice`, to indicate the earliest stored region of the given core.
+
+Notably, if a region is split or transferred, then the `price` is reset to `None`, which has the effect of disallowing a price-increase-capped renewal.
+
+The Broker-chain provides feedback to the Relay-chain on which `ParaId` sets should be serviced on which cores, and does so as they change. Each block, the Broker-chain checks if the period of a `Timeslice` has elapsed. If so, it checks to see if any cores have a newly active `RegionRecord` value in the `Regions` map. If they do, it MUST notify the Relay-chain of the new responsibilities of the relevant `core`. In this case, it MUST remove the item from the `Regions` map and update the `NextRegion` map so it maps the relevant core to the value of removed record's `end` field.
```
rphmeier, line 162: Given that parachains have bounded execution, it may be necessary to split this over a few blocks when dealing with large numbers of cores (e.g. 500-1000)

RFC-0001-Agile Coretime.md
rphmeier: To me that seems quite in-scope for the RFC - as the specifics of the algorithm are likely not configurable by governance, only parameters.

That algorithm looks like it'd work. At least, it does the right thing: price goes up when cores are oversold, and down when cores are undersold. Though we might want some convexity in the adjustment (i.e. price adjusts at a superlinear rate per tick, depending on distance from the target).

RFC-0001-Agile Coretime.md
rphmeier: If we want to accommodate more complex multiplicities, e.g. a 10-way split with 3/10 and 7/10, we would save space and time with a `Vec<(ParaId, u16)>`.

gavofyork: This is intended to be the approximate split between blocks on which the given core is used. For example `[A, B, C]` would be a simple three-way split between paras A, B and C.

`[A, A, B]` would indeed correspond to 2/3 A and 1/3 B.

`[A, A]` is strictly equivalent to `[A]`.

gavofyork: > If we want to accommodate more complex multiplicities, e.g. a 10-way split with 3/10 and 7/10, we would save space and time with a `Vec<(ParaId, u16)>`.

Yes, in the case that it's a 3:7 split, this would indeed occupy less space/time. Of course for the case of a single Para or equal sharing of the core it would increase the space/time.

The question is only what we want to optimise for.

RFC-0001-Agile Coretime.md
gavofyork: Indeed. I don't see any great problem with having multiple `reallocate_core_soon` messages which can be trivially split over multiple blocks if needed. Most of the time I'd expect the changes not be aligned with each other (i.e. only one or two cores changing at any given Timeslice boundary) and therefore the typical case would be pretty trivial.

RFC-0001-Agile Coretime.md
gavofyork: Indeed - that's why I'd prefer to keep it as a pluggable component which can be proposed and iterated on later.

RFC-0001-Agile Coretime.md
gavofyork: If we assume that the average case is biased sharing of a core, then we should definitely switch to a weighted vec. I really can't say for sure how I can imagine the typical usage but including a weight won't cost very much in terms of space and yet will save space in all circumstances where there is bias, so I guess we may as well do it.

RFC-0001-Agile Coretime.md
gavofyork: There is now a reference algorithm in the RFC as an illustration of how it could work.

RFC-0001-Agile Coretime.md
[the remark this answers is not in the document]
gavofyork: Again, they have their blocks processed as long as they can keep winning auctions. The issue is one strictly about price.

RFC-0001-Agile Coretime.md
gavofyork: Done now.

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,291 @@
...
+
+Also:
+- `owner` field of `region` must the equal to the Signed origin.
+- `pivot` must equal neither the `begin` nor `end` fields of the `region`.
+
+#### 3. Consumed in exchange for allocation
+
+A dispatchable `allocate(region: RegionId, target: Vec<ParaId>)` SHALL be provided corresponding to the `allocate` function described above.
+
+It MUST be called with a Signed origin equal to the `owner` field of the value of the Regions map for the `region` key. The `allocation` field of this value MUST be `None` (a region may not be re-allocated).
```
rphmeier, line 177: The description of `split` is somewhat ambiguous - it seems to allow regions with `allocation = Some(...)` to be split (and the new regions inherit the same allocation).

However, that doesn't seem to give any practical benefit, as the new regions wouldn't be able to be reallocated. Am I missing something?

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,291 @@
...
+    price: Balance,
+    sale: SaleIndex,
+}
+```
+
+#### 4. Renewals
+
+A dispatchable `renew(core: CoreIndex)` SHALL be provided. Any account may call `renew` to purchase Bulk Coretime and renew an active allocation for the given `core`.
+
+This MUST be called during the `LEADIN_PERIOD` prior to a Bulk sale (exactly like `purchase`) and has the same effect as `purchase` followed by `allocate` containing the same `Vec<ParaId>`, except that:
```
rphmeier, line 199: If we have a region `(start, end, owner, price, allocation: None)`, and `renew` isn't called, what happens to it? Let's say it's e.g. the last `7 * DAYS`. The first `21 * DAYS` part of it is renewed by its owner, and this is not. Who gets the 7-day chunk that wasn't renewed?

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,298 @@
...
+
+The Broker-chain aims to sell some `BULK_TARGET` of Cores, up to some `BULK_LIMIT`. It makes this sale in a single batch `LEADIN_PERIOD` prior to the beginning of the period being sold. The Broker chain publishes a price for this batch of sales for the `BULK_TARGET` period prior to the sale execution.
+
+Accounts may call a `purchase` function with the appropriate funds in place to have their funds reserved and signal an intention to purchase Bulk Coretime for the forthcoming period. One account may have only one pending purchase. Any number of accounts may attempt to `purchase` Bulk Coretime for the forthcoming period, but the order is recorded.
+
+If there are more purchase requests than available cores for purchase in this period, then requests are processed in incoming order. Any additional purchase orders are carried over but marked as such. A purchase is only cancellable if it was carried over.
```
Tomen, line 121: This might be a potential MEV vector in case where bulk coretime becomes in such high demand that more orders are submitted within the first possible block than can be fulfilled and the validator decides the order in which to accept them.

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,298 @@
...
+
+A regular security review SHOULD be conducted prior to deployment through a review by the Web3 Foundation economic research group.
+
+Any final implementation MUST pass a professional external security audit.
+
+The proposal introduces no new privacy concerns.
+
+## Future Directions and Related Material
+
+RFC-2 proposes a means of implementing the high-level allocations within the Relay-chain.
```
Tomen, line 288: I suspect this is referring to what is currently RFC-3 (Blockspace Regions) #3

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,298 @@
...
+Additionally, there is a renewal system which allows a core's Para assignment to be renewed unchanged with a known maximum price increase from month to month. Note that the intention behind renewals is only to ensure that committed Paras get some guarantees about price for predicting future costs. As such this price-capped renewal system only allows cores to be reused for their same tasks from month to month. In any other context, the regular sale system must be used and the regular price paid.
+
+### Relay-chain and Instantaneous Coretime
+
+Sales of Instantaneous Coretime are expected to happen on the Polkadot Relay-chain. The Relay-chain is expected to be responsible for:
+
+- holding non-transferable, non-refundable DOT balance information for collators.
+- setting and adjusting the price of Instantaneous Coretime based on usage.
+- allowing collators to consume their DOT balance in exchange for the ability to schedule one PoV for near-immediate usage.
+- ensuring the Broker-chain has timely book-keeping information on Coretime sales revenue. This should include a total instantaneous revenue amount for each block number.
```
Tomen, line 87: why is it necessary that the broker-chain receives this information?

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,298 @@
...
+
+However, quite possibly the most substantial problem is both a perceived and often real high barrier to entry of the Polkadot ecosystem. By forcing innovators to either raise 7-figure sums through investors or appeal to the wider token-holding community, Polkadot makes it essentially difficult for a small band of innovators from deploying their technology into Polkadot. While not being actually permissioned, it is also far from the barrierless, permissionless ideal which an innovation platform such as Polkadot should be striving for.
+
+## Requirements
+
+1. The solution SHOULD provide an acceptable value-capture mechanism for the Polkadot network.
+1. The solution SHOULD allow parachains and other projects deployed on to the Polkadot UC to make long-term capital expenditure predictions for the cost of  ongoing deployment.
```
Tomen, line 37: The definition of "Polkadot UC" is not given before it is being used. I suggest either removing it or explaining to the reader what it means, how it differentiates from "Polkadot network" and relay chain. The meaning can be inferred by an informed users, but might not be obvious to anyone who has missed the discussion of recent days.

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,298 @@
...
+    price: Option<Balance>,
+    allocation: Option<Vec<(ParaId, u16)>>, // begins set to `None`
+}
+```
+
+This map functions essentially as a linked list. With one region's `end` field functioning as the next region's key's `begin` field. It is keyed by the sale index in order to allow the following sale period's Coretime to be manipulated during the `LEADIN_PERIOD` prior to it becoming allocatable.
+
+An additional storage map is maintained to keep the "heads" of this linked list. It is called `NextRegion` and it maps `CoreIndex` to `Timeslice`, to indicate the earliest stored region of the given core.
+
+Notably, if a region is split or transferred, then the `price` is reset to `None`, which has the effect of disallowing a price-increase-capped renewal.
```
Tomen, line 145: does this affect the sharing of coretime between parachains?

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,298 @@
...
+
+For an efficient market to form around the provision of Bulk-purchased Cores into the pool of cores available for Instantaneous Coretime purchase, it is crucial to ensure that price changes for the purchase of Instantaneous Coretime are reflected well in the revenues of private Coretime providers during the same period.
+
+In order to ensure this, then it is crucial that Instantaneous Coretime, once purchased, cannot be held indefinitely prior to eventual use since, if this were the case, a nefarious collator could purchase Coretime when cheap and utilize it some time later when expensive and deprive private Coretime providers of their revenue. It SHOULD therefore be assumed that Instantaneous Coretime, once purchased, has a definite and short "shelf-life", after which it becomes unusable. This incentivizes collators to avoid purchasing Coretime unless they expect to utilize it imminently and thus helps create an efficient market-feedback mechanism whereby a higher price will actually result in material revenues for private Coretime providers who contribute to the pool of Cores available to service Instantaneous Coretime purchases.
```
eskimor, line 244: in the current implementation, purchased is equivalent to queued. Hence it will be served as soon as possible.

RFC-0001-Agile Coretime.md
jonasW3F: Maybe its better to burn the DOT instead of diverting them to Treasury. There are a few reasons for that:
1) In the absence of 2y locks on DOT, the system might benefit from a permanent sink for DOTs. We also might consider to increase the ideal staking rate. Non-interactive staking might serve well here, too.
2) Inflow from coretime usage might, especially in the short-term, be very volatile. The Treasury conceptually benefits from predictable inflow, allowing for long-term budgeting. Inflation is the best way to do that. We'd counter that with burning for coretime.
3) This mechanism would lead to high inflow in times with high coretime usage and low inflow in times of low usage. It seems to me that, if anything, it should be the opposite. With a steady inflow the Treasury would always have enough funds to respond to demand shocks in coretime when necessary (by funding good projects / initiatives).

RFC-0001-Agile Coretime.md
rphmeier: Note that, as I understand it, the broker chain doesn't attempt to do anything related to on-demand / Instantaneous except tell the relay-chain that certain cores (or fractions of cores) are to be allocated into the on-demand mechanism for some time period, and expects some information about how much revenue was accrued during that time period on the core to be communicated back to the broker chain or readable from it.

We're still discussing the exact interface that the broker chain would use to communicate with the relay chain, but it'd be something like

`RelayChain::allocate(core_index, start, Option<end>, allocation)` where the `allocation: Vec<ParaId>` can encode information about allocating some or all of the coretime to the on-demand pool.

RFC-0001-Agile Coretime.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Summary
+
+This proposes a periodic, sale-based method for assigning and Polkadot Coretime. The method takes into account the need for long-term capital expenditure planning for teams building on Polkadot, yet also provides a means to allow Polkadot to capture long-term value in the resource which it sells. It supports the possibility of building secondary markets to make resource allocation more efficient and largely avoids the need for parameterisation.
+
+## Motivation
+
+### Present System
+
+The present system of allocating time for parachains on the cores of the Polkadot Ubiquitous Computer (aka "Polkadot") is through a process known as *slot auctions*. These are on-chain candle auctions which proceed for several days and result in a core being assigned to a single parachain for six months at a time up to 18 months in advance. Practically speaking, we only see two year periods being bid upon and leased.
```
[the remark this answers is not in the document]
joepetrowski: I read the 18 as correct: You bid on a six month term, 18 months in advance, allowing you to see 24 months into the future. As in, the six month term starts in 18 months, so the end is in 6 + 18 = 24 months.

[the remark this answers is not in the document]
Tomen: hm, I think it's a bit ambiguous, but maybe it's because I am not a native speaker.

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,486 @@
...
+
+#### Bulk Sales
+
+A sale of Bulk Coretime occurs on the Broker Chain every `BULK_PERIOD` blocks.
+
+In every sale, a `BULK_LIMIT` of individual *Regions* are offered for sale at a particular Sale Price.
+
+The Regions offered for sale have the same span: they last exactly `BULK_PERIOD` blocks, and begin `LEADIN_PERIOD` blocks into the future at the time of the sale.
+
+The Regions offered for sale also have complete
```
Tomen, line 136: this sentence is incomplete

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,486 @@
...
+
+Each Region offered for sale has a different Core Index, ensuring that they each represent an independently allocatable resource on the Polkadot UC.
+
+After every sale, the Next Sale Price is set according to the Previous Sale Price and the number of Regions sold compared to the desired and maximum amount of Regions to be sold. See Price Setting for additional detail.
```
Tomen, line 140: This model works under the assumption of DOT price volatility within a certain range. If we see a month where DOT price surges drastically, we encounter a condition that can make the price of core for the upcoming month unaffordable for a number of parachains (e.g. because they acquire the DOT on demand and suddenly see themselves unable to pay the rent for the coming month).

In such a situation, we would see that BULK_TARGET is missed by a large margin. This would result in a price decrease for the coming month, but would still present a situation where parachains miss the opportunity to acquire bulk coretime in that month.

We could either account for this by making bulk coretime dynamic pricing more complex (e.g. reverse auction during the lead-in period), or ensure that in such a situation instantaneous coretime sales is robust enough to offer parachains a viable alternative for this month.

Not a trivial decision, but I think it will present parachain teams with an operational overkill/burnout potential, if they have to pull off the operations of securing instantaneous coretime for a month.

Additionally, as far as I see it we currently project that bulk coretime goes live before instantaneous coretime; so this might kill off parachains for a month if they miss their purchase.

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,486 @@
...
+
+The specific price setting mechanism is out of scope for this proposal and should be covered in some other work. The present proposal assumes the existence of a price-setting mechanism which could take into account three parameters; two mostly fixed and one which changes each `BULK_PERIOD`. These parameters are `BULK_TARGET`, `BULK_LIMIT` and `CORES_SOLD` which is the actual number of cores sold for the present `BULK_PERIOD` and is always an unsigned integer at most `BULK_LIMIT`.
+
+In general we would expect the price to increase the closer `CORES_SOLD` gets to `BULK_LIMIT` and to decrease the closer it gets to zero. If it is exactly equal to `BULK_TARGET`, then we would expect the price to remain the same.
+
+A simple example of this would be the formula:
+
+```
+NEW_PRICE := IF CORES_OLD < BULK_TARGET THEN
+    OLD_PRICE - OLD_PRICE / 2 * CORES_SOLD / BULK_TARGET
```
Tomen, line 252: the left side of the function is flipped: it goes from OLD_PRICE @ 0 to OLD_PRICE/2 @ BULK_TARGET

The correct left side of the function looks like this:
OLD_PRICE / 2 + OLD_PRICE / 2 * CORES_SOLD / BULK_TARGET

or

OLD_PRICE / 2 * (1 + CORES_SOLD / BULK_TARGET)

Tomen:
I submitted a proposal for dynamic pricing here:

https://forum.polkadot.network/t/dynamic-pricing-for-bulk-coretime-sales/3359

RFC-0001-Agile Coretime.md
gavofyork: That's fixed now.

RFC-0001-Agile Coretime.md
gavofyork: `renew` can only be called on regions which were the full `BULK_PERIOD` in length and the assignment but be exactly the same from month to month.

RFC-0001-Agile Coretime.md
gavofyork: To be able to make proper rewards.

RFC-0001-Agile Coretime.md
gavofyork: There's an intro now.

RFC-0001-Agile Coretime.md
gavofyork: This mechanism is changed in the latest revision and should be clearer.

RFC-0001-Agile Coretime.md
gavofyork: (See https://github.com/polkadot-fellows/RFCs/pull/5)

text/0001-agile-coretime.md
gavofyork: completed.

text/0001-agile-coretime.md
gavofyork: > because they acquire the DOT just in time and suddenly see themselves unable to pay the rent for the coming month

Then obviously the chains should be holding DOT to mitigate exchange risk.

gavofyork:
> I submitted a proposal for dynamic pricing here:
> 
> https://forum.polkadot.network/t/dynamic-pricing-for-bulk-coretime-sales/3359

Perhaps consider opening an RFC.

RFC-0001-Agile Coretime.md
gavofyork: True, but month to month this should not be much of an issue as the price gets regulated properly.

RFC-0001-Agile Coretime.md
BlackDemonZyT: I totally agree with @jonasW3F , DOT from Coretime sales should be instantly burnt in order to slow down inflation.

Tomen:
> Perhaps consider opening an RFC.

Okay: #6

RFC-0001-Agile Coretime.md
gavofyork: This has been modified to a hybrid auction system which should reduce the potential for MEV somewhat.

Tomen:
I have discovered an edge case that I think is not covered for yet.

Let's say `BULK_LIMIT = 50; BULK_TARGET = 35;`.

A normal situation might be that we have 33 parachains that run on renewals. We are under the target and prices keep falling.

Now let's assume that 5 new parachains join and will buy renewals from now on forward. The result is a situation where renewal price increases are capped and the regular price rises at a faster rate, pushing the regular price up continuously without anyone actually paying at that price.

Governance might steer against this trend by bumping up the target, but there could be situations where that is not desired.

@gavofyork

rphmeier:
Yeah, the pricing mechanism should account for the number of regions that are currently being priced lower than the "actual" price and avoid increasing in that case. We definitely don't want the "price" to skyrocket exponentially higher than what is actually being paid.

gavofyork:
In the case that nothing is being sold at the nominal price then we can just keep it constant from last month.

muharem (approved):
cool

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,627 @@
...
+
+We can define a very simple progression where the price decreases monotonically from double the Sale Price at the beginning of the Leadin Period.
+
+```
+FACTOR(T) := 2 - T
+```
+
+#### Instantaneous Price Progression
+
+This proposal assumes the existence of a Relay-chain-based price-setting mechanism for the Instantaneous Coretime Market which alters from block to block, taking into account several parameters: the last price, the size of the Instantaneous Coretime Pool (in terms of cores per block) and the amount of Instantaneous Coretime waiting for processing (in terms of Core-blocks queued).
```
bkchr, line 343:
```suggestion
This proposal assumes the existence of a Relay-chain-based price-setting mechanism for the Instantaneous Coretime Market which alters from block to block, taking into account several parameters: the last price, the size of the Instantaneous Coretime Pool (in terms of cores per relay-chain block) and the amount of Instantaneous Coretime waiting for processing (in terms of Core-blocks queued).
```

"core blocks queued", wouldn't "tasks queued" be better?

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,627 @@
...
+
+The ideal situation is to have the size of the Instantaneous Coretime Pool be equal to some factor of the Instantaneous Coretime waiting. This allows all Instantaneous Coretime sales to be processed with some limited latency while giving limited flexibility over ordering to the Relay-chain apparatus which is needed for efficient operation.
```
bkchr, line 345: If we have more coretime available then waiting, will the coretime be bought "randomly" from the pool to ensure that all coretime providers get a fair share?

bkchr: Or will it favor coretime that wasn't sold and is still owned by the relay chain be favored?

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,627 @@
...
+// 40-bit (5 bytes). Could be 32-bit with a more specialised type.
+enum CoreTask {
+    Off,
+    Assigned { target: ParaId },
+    InstaPool,
+}
+// 113-bit (14 bytes). Could be 14 bytes with a specialised 32-bit `CoreTask`.
+struct ScheduleItem {
+    mask: CoreMask, // 80 bit
+    task: CoreTask, // 33 bit
```
bkchr, line 404:
```suggestion
// 120-bit (15 bytes). Could be 14 bytes with a specialised 32-bit `CoreTask`.
struct ScheduleItem {
    mask: CoreMask, // 80 bit
    task: CoreTask, // 40 bit
```

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,627 @@
...
+7. Governance upgrade proposal(s).
+8. Monitoring of the upgrade process.
+
+## Performance, Ergonomics and Compatibility
+
+No specific considerations.
+
+Parachains already deployed into the Polkadot UC must have a clear plan of action to migrate to an agile Coretime market.
+
+While this proposal does not introduce documentable features per se, adequate documentation must be provided to potential purchasers of Polkadot Coretime. This SHOULD include any alterations to the Polkadot-SDK software collection, most likely Cumulus.
```
bkchr, line 592:
```suggestion
While this proposal does not introduce documentable features per se, adequate documentation must be provided to potential purchasers of Polkadot Coretime. This SHOULD include any alterations to the Polkadot-SDK software collection.
```

Maybe there are other SDKs as well besides the one from Parity.

bkchr (approved):
Thank you for this nice RFC :) It explains all the things in good detail and gives the reader a good understanding of the how this should work.

I left mainly nitpicks and some questions here and there.

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,627 @@
...
+
+Bulk Coretime is sold periodically on a specialised system chain known as the *Coretime-chain* and allocated in advance of its usage, whereas Instantaneous Coretime is sold on the Relay-chain immediately prior to usage on a block-by-block basis.
+
+This proposal does not fix what should be done with revenue from sales of Coretime and leaves it for a further RFC process.
+
+Owners of Bulk Coretime are tracked on the Coretime-chain and the ownership status and properties of the owned Coretime are exposed over XCM as a non-fungible asset.
+
+At the request of the owner, the Coretime-chain allows a single Bulk Coretime asset, known as a *Region*, to be used in various ways including transferal to another owner, allocated to a particular task (e.g. a parachain) or placed in the Instantaneous Coretime Pool. Regions can also be split out, either into non-overlapping sub-spans or exactly-overlapping spans with less regularity.
```
kianenigma, line 72: In essence, everything is bulk core time at first. Then, a subset is scheduled to the instantaneous market by the owner (and the owner could be the system itself).

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,627 @@
...
+
+Also:
+- `owner` field of `region` must the equal to the Signed origin.
+- `mask` must have some bits set AND must not equal the Core Mask of the old Region AND must only have bits set which are also set in the old Region's' Core Mask.
+
+#### 4. `assign`
+
+Regions may be assigned to a core.
+
+A `assign(region: RegionId, target: ParaId, finality: Finality)` dispatchable shall have the effect of placing an item in the workplan corresponding to the region's properties and assigned to the `target` task.
```
kianenigma, line 235: ParaId -> TaskId?

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,627 @@
...
+
+Though the Instantaneous Coretime Credit account records a balance for an account identifier (very likely controlled by a collator), it is *non-transferable* and *non-refundable*. It can only be consumed in order to purchase some Instantaneous Coretime with immediate availability.
+
+The Relay-chain reports this usage back to the Coretime-chain in order to allow it to reward the providers of the underlying Coretime, either the Polkadot System or owners of Bulk Coretime who contributed to the Instantaneous Coretime Pool.
+
+Specifically the Relay-chain is expected to be responsible for:
+
+- holding non-transferable, non-refundable DOT-denominated Instantaneous Coretime Credit balance information for collators.
+- setting and adjusting the price of Instantaneous Coretime based on usage.
+- allowing collators to consume their Instantaneous Coretime Credit at the current pricing in exchange for the ability to schedule one PoV for near-immediate usage.
```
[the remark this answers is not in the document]
rphmeier: Probably something like "purchase coretime and use within some short period". The timeframe doesn't matter that much. The two constraints in practice (beyond the scope of the RFC) are:
  1. The economic constraint. Scheduled Coretime needs to expire within a short time period where coretime price volatility is pretty low, or it can be "saved up" when cheap and spent when expensive. This is what happened with GasToken in Ethereum and it caused large problems. This constraint leads to necessary expiry times of hours or days.
  2. The resource constraint. Scheduled coretime shouldn't clog up the scheduler, so the turnaround from buying instantaneous coretime to using it needs to be spent in a timeframe that doesn't overload the scheduler. This constraint probably indicates an expiry time of 30s-1m, depending on what the relay-chain scheduler can handle.

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,627 @@
...
+
+The Sale Price varies during an initial portion of the Purchasing Period called the *Leadin Period* and then stays stable for the remainder. This initial portion is `LEADIN_PERIOD` blocks in duration. During the Leadin Period the price decreases towards the Sale Price, which it lands at by the end of the Leadin Period. The actual curve by which the price starts and descends to the Sale Price is outside the scope of this RFC, though a basic suggestion is provided in the Price Setting Notes, below.
+
+#### Renewals
+
+At any time when there are remaining Regions of Bulk Coretime to be sold, *including during the Interlude Period*, then certain Bulk Coretime assignmnents may be *Renewed*. This is similar to a purchase in that funds must be paid and it consumes one of the Regions of Bulk Coretime which would otherwise be placed for purchase. However there are two key differences.
+
+Firstly, the price paid is exactly `RENEWAL_PRICE_CAP` more than what the purchase/renewal price was in the previous sale.
```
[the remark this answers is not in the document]
rphmeier: Yeah, agreed - this mechanism would push renewals to go back onto the open market when the market price is falling. That could lead to worse guarantees of continuity, if demand happens to pick up again this period.

text/0001-agile-coretime.md
rphmeier: Yes, or other Coretime marketplaces that manage allocations on the broker chain

text/0001-agile-coretime.md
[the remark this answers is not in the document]
lucasvo: I do think this is quite a bit worse than what was originally proposed by @gavofyork - that stated renewals should cost the lower of the current sale or current core cost multiplied by `RENEWAL_PRICE_CAP`.

text/0001-agile-coretime.md
gavofyork: Pool shares (i.e. Coretime contributions) are fungible over a period. The revenue is aggregated over the period and distributed pro rata according to overall contributions over that period.

text/0001-agile-coretime.md
gavofyork: One task (i.e. one parachain ID) may have more than one instance in the pool, so it's not *tasks*, but one-block cores, or a single atomic unit of Coretime. It's the resource R in the sentence "Pay the Instantaneous Coretime price for your task to be executed by R".

text/0001-agile-coretime.md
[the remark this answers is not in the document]
gavofyork: So as a workload which is renewable (practically speaking for now, a parachain, of group thereof), you would have three options:

- Pay 2% (or whatever) more than last month and guarantee your price and your slot.
- Pay the nominal (post-lead-in) price, which may be lower but requires competing within the market and waiting the lead-in period.
- Pay a higher-than-nominal price at some point during the lead-in period.

I would argue that placing a fixed-rate price increase could be defensible given that the buyer gets absolute certainty over availability. Nobody else in the market gets this. If the price ends up being higher than the apparent spot price, and the buyer wishes to optimize their costs, they are always free to join the regular market like everyone else.

The renewal mechanism is market-warping and has the potential to be anti-competitive. It should be subject to stringent safeguards.

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,627 @@
...
+
+#### 3. `interlace`
+
+Regions may be decomposed into two Regions of the same span whose eventual assignments take turns on the core by virtue of having complementary Core Masks.
+
+An `interlace(region: RegionId, mask: CoreMask)` dispatchable shall have the effect of removing the Region identified by `region` and creating two new Regions. The new Regions will each have the same span and owner of the original Region, but one Region will have a Core Mask equal to `mask` and the other will have Core Mask equal to the XOR of `mask` and the Core Mask of the original Region.
+
+Also:
+- `owner` field of `region` must the equal to the Signed origin.
+- `mask` must have some bits set AND must not equal the Core Mask of the old Region AND must only have bits set which are also set in the old Region's' Core Mask.
```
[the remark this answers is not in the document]
gavofyork: That is strictly superfluous from the second and third constraints.

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,627 @@
...
+
+#### 4. `assign`
+
+Regions may be assigned to a core.
+
+A `assign(region: RegionId, target: ParaId, finality: Finality)` dispatchable shall have the effect of placing an item in the workplan corresponding to the region's properties and assigned to the `target` task.
+
+If the region's end has already passed (taking into account any advance notice requirements) then this operation is a no-op. If the region's begining has already passed, then it is effectively altered to become the next schedulable timeslice.
+
+`finality` may have the value of either `Final` or `Provisional`. If `Final`, then the operation is free, the `region` record is removed entirely from storage and renewal may be possible: if the Region's span is the entire `BULK_PERIOD`, then the Coretime-chain records in storage that the allocation happened during this period in order to facilitate the possibility for a renewal. (Renewal only becomes possible when the full Core Mask of a core is finally assigned for the full `BULK_PERIOD`.)
```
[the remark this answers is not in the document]
gavofyork: Yes.

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,627 @@
...
+
+This may be called during the Interlude Period as well as the regular Purchasing Period and has the same effect as `purchase` followed by `allocate`, except that:
+
+1. The price of the purchase is the previous `price` incremented by `RENEWAL_PRICE_CAP`.
+1. The Region is allocated exactly the given `core` is currently allocated for the present Region.
+
+Renewal is only valid where a Region's span is assigned to Tasks (not placed in the Instantaneous Coretime Pool) for the entire unsplit `BULK_PERIOD` over all of the Core Mask and with Finality. There are thus three possibilities of a renewal being allowed:
+
+1. Purchased unsplit Coretime with final assignment to tasks over the full Core Mask.
+1. Renewed Coretime.
```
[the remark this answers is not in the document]
gavofyork: Purchase (+ manipulation + trade) + assignment is a different process to Renewal, and comes with constraints which do not apply in the latter case.

text/0001-agile-coretime.md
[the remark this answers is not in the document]
lucasvo: My problem with this is that to this date not a single parachain has done an automated trustless renewal, i.e. built a mechanism by which the parachain permissionlessly renewed a slot. The auctions are just too hard to build a strategy for and you're too much at a disadvantage as a parachain (vs. an offchain bidder). If we create a mechanism again where it's very hard for an existing workloads to renew at the optimal price then we will end up in the same situation again. 

If you are running a large application with many many users depending on it being available, **you will need certainty of uptime at all cost**.  So what would I do if I tried to design a strategy for Centrifuge? By default I will always have to renew because I can't risk not getting a core during LEADIN or post LEADIN.

Say prices are actually going down (which they should in the long term) I would then have to buy a second core on the market the minute it's cheap enough that over a few renewal periods I will break even paying for a core twice for one era. I would not want to risk going offline. This will lead again to very complex acquisition strategies which are not feasible to automate. We have to build a system that makes workload renewal very simple to automate and paid for by the chains themselves. **Everything else puts a huge amount of complexity onto teams that want to build in Polkadot.**

> The renewal mechanism is market-warping and has the potential to be anti-competitive. It should be subject to stringent safeguards.

Why not at least give renewals the same price as anyone could get by buying in the LEADIN period? I think this is a reasonable compromise.

That would satisfy two needs for existing workloads:
* Guarantee renewal provided the workload can pay *a fee*
* Guarantee that renewals are not more than the *highest possible* price in the current purchase period.

I'm sure some users will want to bet on a lower price than set in LEADIN and wait until after to bid but that would then come with the real risk of going offline.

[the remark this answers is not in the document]
lolmcshizz: > Why not at least give renewals the same price as anyone could get by buying in the LEADIN period? I think this is a reasonable compromise.

Agree with this - I have already asked questions internally regarding how we could automate renewal for HydraDX. The benefit of the existing crowdloan model is we can almost forget about it for the best part of 2 years - suddenly requiring teams to be looking every 2-4 weeks for the optimal price is too much overhead.

eskimor (commented):
Very good read!

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,652 @@
...
+
+The ideal situation is to have the size of the Instantaneous Coretime Pool be equal to some factor of the Instantaneous Coretime waiting. This allows all Instantaneous Coretime sales to be processed with some limited latency while giving limited flexibility over ordering to the Relay-chain apparatus which is needed for efficient operation.
+
+If we set a factor of three, and thus aim to retain a queue of Instantaneous Coretime Sales which can be processed within three Relay-chain blocks, then we would increase the price if the queue goes above three times the amount of cores available, and decrease if it goes under.
+
+Let us assume the values `OLD_PRICE`, `FACTOR`, `QUEUE_SIZE` and `POOL_SIZE`. A simple definition of the `NEW_PRICE` would be thus:
+
+```
+NEW_PRICE := IF QUEUE_SIZE < POOL_SIZE * FACTOR THEN
```
eskimor, line 377: I think the very least we would need to use some average value of `POOL_SIZE`, given that the number of on-demand cores can essentially change very block, given for example that regions with only a single bit set can be assigned to the instantaneous market.

gavofyork: It's not intended to be anything other than an illustration.

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,652 @@
...
+type TotalMaskBits = u32;
+struct InstaPoolHistoryRecord {
+    total_contributions: TotalMaskBits,
+    maybe_payout: Option<Balance>,
+}
+/// Total InstaPool rewards for each Timeslice and the number of core Mask which contributed.
+type InstaPoolHistory = Map<Timeslice, InstaPoolHistoryRecord>;
+```
+
+`CoreMask` tracks unique "parts" of a single core. It is used with interlacing in order to give a unique identifier to each compnent of any possible interlacing configuration of a core, allowing for simple self-describing keys for all core ownership and allocation information. It also allows for each core's workload to be tracked and updated progressively, keeping ongoing compute costs well-bounded and low.
```
gavofyork, line 467:
```suggestion
`CoreMask` tracks unique "parts" of a single core. It is used with interlacing in order to give a unique identifier to each component of any possible interlacing configuration of a core, allowing for simple self-describing keys for all core ownership and allocation information. It also allows for each core's workload to be tracked and updated progressively, keeping ongoing compute costs well-bounded and low.
```

text/0001-agile-coretime.md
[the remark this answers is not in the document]
gavofyork: Probably a fair compromise.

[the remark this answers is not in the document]
gavofyork: Changed to mirror this.

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,652 @@
...
+
+**Renewal Price**
+
+The Renewal Price is:
+
+- If the workload being renewed came to be through the *Purchase and Assignment* of Bulk Coretime, then the price paid during the Purchase operation.
+- If the workload being renewed was previously renewed, then the price paid during this previous Renewal operation plus `RENEWAL_PRICE_CAP`.
+- If the workload being renewed is a migation from a legacy slot auction lease, then the nominal price for a Regular Purchase (outside of the Lead-in Period) of the Sale during which the legacy lease expired.
+
+#### 8. Instantaneous Coretime Credits
```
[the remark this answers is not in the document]
eskimor: Ok, makes sense. If we don't have a balances pallet on Polkadot, there is also no other way.

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,652 @@
...
+- Workload to communicate core (re-)assignments is well-bounded and low in weight.
+- All mandatory bookkeeping workload is well-bounded in weight.
+
+```rust
+type Timeslice = u32; // 80 block amounts.
+type CoreIndex = u16;
+type CoreMask = [u8; 10]; // 80-bit bitmap.
+
+// 128-bit (16 bytes)
+struct RegionId {
```
rphmeier, line 407: as an implementation note, out-of-scope for the RFC, but these datatypes depend on using a packed representation as opposed to standard alignments

gavofyork: I did make a test to check the size (at least on my 64-bit M1 architecture) and it was indeed 128-bit as expected without any explicit packing.

text/0001-agile-coretime.md
```diff
@@ -0,0 +1,652 @@
...
+- The price increase per Bulk Coretime period for renewals.
+- The price decrease graph in the Leadin period for Bulk Coretime sales.
+- The initial price of Instantaneous Coretime.
+- The price-change algorithm for Instantaneous Coretime sales.
+- The percentage of cores to be sold as Bulk Coretime.
+- The fate of revenue collected.
+
+## Prior Art and References
+
+None.
```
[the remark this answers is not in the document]
gavofyork: Added.

RFC-0001-Agile Coretime.md
Haroonshaukat: This is definitely going to be HUGE! as the team has been an inspiration in their innovative approaches towards achieving the project goals and visions, I am so glad to be part of this project
This is definitely going to be HUGE! as the team has been an inspiration in their innovative approaches towards achieving the project goals and visions, I am so glad to be part of this project
This is definitely going to be HUGE! as the team has been an inspiration in their innovative ap

PR #4, opened 2023-07-04
Remove unnecessary host functions allocator usage

tomaka:
I have followed the template in https://github.com/polkadot-fellows/RFCs/pull/2.


text/0000-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,202 @@
...
+- `ext_storage_next_key`
+- `ext_default_child_storage_next_key`
+- `ext_crypto_ed25519_public_keys`
+- `ext_crypto_sr25519_public_keys`
+- `ext_crypto_ecdsa_public_keys`
+- `ext_offchain_network_state`
+- `ext_offchain_local_storage_get`
+- `ext_offchain_http_response_wait`
+- `ext_offchain_http_response_headers`
+- `ext_offchain_http_response_read_body`
```
tomaka, line 34: Note that this is to the extent of my knowledge. The list of host functions that are used in production (as opposed to the ones that are just prototypes) is not super well tracked.

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,212 @@
...
+
+The API of these new functions was heavily inspired by API used by the C programming language.
+
+## Unresolved Questions
+
+No unresolved questions.
+
+## Future Possibilities
+
+The existence of the host-side allocator has become questionable over time. It is implemented in a very naive way, and for determinism and backwards compatibility reasons it needs to be implemented exactly identically in every client implementation. Furthermore, runtimes make substantial use of heap memory allocations, and each allocation needs to go twice through the runtime <-> host boundary (once for allocating and once for freeing). Moving the allocator to the runtime side, while it would increase the size of the runtime, would be a good idea. But before the host-side allocator can be deprecated, all the host functions that make use of it need to be updated to not use it.
```
rphmeier, line 201: How do you move the allocator to the runtime side?

Do you just give it a big block of memory and let it split it up as desired?

tomaka: I don't really see the problem? The runtime could just do whatever it wants with its memory.

If the question is how to make sure that the memory usage is bounded, my personal opinion would be that at initialization the runtime does `storage_get(":heappages")` and puts the value in its memory allocator, and the allocator crashes it it reaches the limit.

Alternatively, the wasm memory needs to be grown by the wasm code with a `memory.grow` opcode, so the host can detect when memory usage is over a certain threshold.

rphmeier: Not an issue, I'm just not familiar enough with Wasm and was curious about how this would be done in practice

tomaka: Moving the allocator back into the runtime would make the runtime more similar to a normal program. The fact that the allocator is outside of the runtime is what requires additional code/hacks. So the answer on how to do move the allocator back to the runtime is simply "remove the hacks".

rphmeier (commented):
Is my understanding of the RFC correct that it proposes to _introduce_ the new functions but not to remove the old ones?

This seems valid, but after being accepted/implemented, the old versions should be deprecated and then removed entirely (after some significant period of time, to allow backwards compatibility)

tomaka:
> Is my understanding of the RFC correct that it proposes to introduce the new functions but not to remove the old ones?

Yes!

> This seems valid, but after being accepted/implemented, the old versions should be deprecated and then removed entirely (after some significant period of time, to allow backwards compatibility)

They should be deprecated, yes, but removing them would mean that you could no longer sync older blocks.
I'm not necessarily against breaking the possibility to sync older blocks, but that's not in scope of this RFC.

pepyakin:
This is definitely a step into the right direction. I also agree on that the C API approach makes sense under current condition. I am a bit worried though that this proposal doesn't provide any numbers though.

I am thinking that in the future, we might need to consider the API design overhaul as if it's 2023 actually. But this is outside of the scope of this proposal.

Just for linking purposes, here are a relevant discussion on Substrate: https://github.com/paritytech/substrate/issues/11883

burdges:
Is there an obstacle to host functions actually doing borrowing aka just pass pointers?  We need not even copy memory in say: 
```
fn host_function(foo: &'a mut [u8], &'b [u8; 32]) -> &'a mut [u8];
```

Also..

A hash interface like `pub fn keccak_256(data: &[u8]) -> [u8; 32]` is inherently inefficient because you must build some `Vec<u8>` before doing the hashing.  Although RustCrypto/digest is imperfect, at least their [traits](https://docs.rs/digest/latest/digest/#traits) avoid this.

We could've wasm types designed for usage directly in the host, like demanding u64 or simd alignment, and then impl appropriate `digest::{Update, Reset, FixedOutput*, ExtendableOutput*, XofReader}` directly upon those. 
```
#[derive(Clone)]
#[repr(align(8))]
pub(crate) struct Sha3State {
    state: [u8; 26 * 8],
    //  [u64; PLEN] with PLEN = 25 plus a usize=u64
    // https://github.com/RustCrypto/hashes/blob/master/sha3/src/state.rs#L9
}

impl Update for {
    fn update(&mut self, data: &[u8]) { .. }
}
...
```
You'd then cargo patch in sp-sha3 over sha3 and then you've a faster hashing interface.

tomaka:
> This is definitely a step into the right direction. I also agree on that the C API approach makes sense under current condition. I am a bit worried though that this proposal doesn't provide any numbers though.

I have opened this RFC with the desire of making things move. I would be more than happy if someone else took over this effort and provided numbers. Since nobody else is proposing these changes, it will have to do with me and without numbers.

> Is there an obstacle to host functions actually doing borrowing aka just pass pointers? We need not even copy memory in say:

> A hash interface like pub fn keccak_256(data: &[u8]) -> [u8; 32] is inherently inefficient for variable sized data because you must build some Vec<u8> before doing the hashing. Although RustCrypto/digest is imperfect, at least their [traits](https://docs.rs/digest/latest/digest/#traits) avoid this.

This is very precisely what this RFC proposes fixes.

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,210 @@
...
+(func $ext_hashing_blake2_128_version_2
+    (param $data i64) (param $out i32))
+(func $ext_hashing_blake2_256_version_2
+    (param $data i64) (param $out i32))
+(func $ext_hashing_twox_64_version_2
+    (param $data i64) (param $out i32))
+(func $ext_hashing_twox_128_version_2
+    (param $data i64) (param $out i32))
+(func $ext_hashing_twox_256_version_2
+    (param $data i64) (param $out i32))
```
burdges, line 71: Are the twox hashes still being used anywhere?  I'd thought we deprecated them since they're obvious footguns.

burdges:
> This is very precisely what this RFC fixes.

It seems only for the case where a fixed length buffer is being hashed.  [`sp_io::Hashing`](https://github.com/paritytech/substrate/blob/master/primitives/io/src/lib.rs#L1193) inherently demands allocations when hashing a variable length buffer.  A variable length buffer would be broken down into fixed length parts:
```
#[runtime_interface]
pub trait Hashing {
        //  Existing fixed length buffer stuff

	fn keccak_256_update(state: &mut Sha3State, data: &[u8])
	fn keccak_256_finalize(state: &Sha3State) -> [u8; 32]

	fn keccak_512_update(state: &mut Sha3State, data: &[u8])
	fn keccak_512_finalize(state: &Sha3State) -> [u8; 64]

	fn sha2_256_update(state: &mut Sha2State, data: &[u8])
	fn sha2_256_finalize(state: &Sha2State) -> [u8; 32]

	fn blake2_256_update(state: &mut Blake2State, data: &[u8])
	fn blake2_256_finalize(state: &Blake2State) -> [u8; 32]
}
```

Anyways..

If we typically hash fixed length buffers then maybe this variable length buffer allocation is best to just keep, as the above interfaces demands quite a few more calls.

> But before the host-side allocator can be deprecated, all the host functions that make use of it need to be updated to not use it. 

I see, we need roughly this for backwards compatibility anyways.

tomaka:
> Just for the case where a fixed length buffer is being hashed. [sp_io::Hashing](https://github.com/paritytech/substrate/blob/master/primitives/io/src/lib.rs#L1193) inherently demands allocations when hashing a variable length buffer.

`sp_io::Hashing` is an implementation detail of Substrate, not part of the Polkadot specification.

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,210 @@
...
+    (param $child_storage_key i64) (param $out i32))
+(func $ext_crypto_ed25519_generate_version_2
+    (param $key_type_id i32) (param $seed i64) (param $out i32))
+(func $ext_crypto_sr25519_generate_version_2
+    (param $key_type_id i32) (param $seed i64) (param $out i32) (return i32))
+(func $ext_crypto_ecdsa_generate_version_2
+    (param $key_type_id i32) (param $seed i64) (param $out i32) (return i32))
+```
+
+The behaviour of these functions is identical to their version 1 or version 2 equivalent. Instead of allocating a buffer, writing the output to it, and returning a pointer to it, the new version of these functions accepts an `out` parameter containing the memory location where the host writes the output. The output is always of a size known at compilation time.
```
rphmeier, line 90: The expected behavior (trap?) for an out-of-bounds memory access ought to be specified.

tomaka: You're actually (I guess unintentionally) raising a good point.

If we take for example `ext_crypto_ed25519_sign_version_2`, on success the function returns 0 and writes to `out`, and on failure the function returns 1 and doesn't write anything. Should a trap happen if `out` is out of range even in the situation where 1 would be returned?

I've added some text mentioning that yes, the range of `out` must always be checked.

yjhmelody:
Hi, I want to know any plan to move the allocator  inside of wasm runtime.
It definitly will make substrate runtime be more generic.

tomaka:
Same remark as https://github.com/polkadot-fellows/RFCs/pull/8#issuecomment-1722447950

I would like to propose this RFC, but I don't know how to proceed.

bkchr:
To move forward with this pull request, I would like to see some basic benchmarks being done. @yjhmelody could you please help with this?

bkchr:
I would also like to see that this directly moves the entire allocator to the runtime. Not sure why we only should do this half way.

tomaka:
I can expand the RFC if you want.
When I opened it, I actually expected a swift and unanimous approval.

If RFCs inherently need a long time, I might as well make it as big as possible.

koute:
+1 for going all the way and moving the whole allocator inside of the runtime, so that for new blocks the allocator host functions are never called anymore.

bkchr:
> When I opened this RFC (two and a half months ago), I actually expected a swift and unanimous approval.

Yeah sorry on that, but it should actually also not be "swift". Let's do it properly.

tomaka:
I've updated the RFC.

As I've mentioned in the "unresolved questions" section, I expect the offchain worker functions to be slightly slower than right now, but this is IMO negligible since they are offchain.

To me, the biggest potential sources of performance degradation comes from the deprecation of `ext_storage_get_version_1`, as the runtime might have to call `ext_storage_read_version_2` multiple times instead of calling `ext_storage_get_version_1` once.
In theory it should be possible to progressively read storage items (in a streaming way) whose size isn't known ahead of time, rather than completely load it in one go, but I understand that this would be a pretty large change to the runtime code and I'm afraid that half-assing the change might lead to reading becoming quite slow.

The changes to `ext_storage_next_key` might also cause performance degradation if not implemented properly. This is also a moderately sized change to the runtime code (that will degrade performances if half-assed), but I have the impression that it shouldn't be too hard to do it properly.

tomaka:
> To move forward with this pull request, I would like to see some basic benchmarks being done. @yjhmelody could you please help with this?

Is something in progress?

While I agree that RFCs shouldn't necessarily be swift, if nothing is done then the ETA of this RFC is "infinity" and it would be great if it was finite.

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,328 @@
...
+    (param $child_storage_key i64) (param $out i32))
+(func $ext_crypto_ed25519_generate_version_2
+    (param $key_type_id i32) (param $seed i64) (param $out i32))
+(func $ext_crypto_sr25519_generate_version_2
+    (param $key_type_id i32) (param $seed i64) (param $out i32) (return i32))
+(func $ext_crypto_ecdsa_generate_version_2
+    (param $key_type_id i32) (param $seed i64) (param $out i32) (return i32))
+```
+
+The behaviour of these functions is identical to their version 1 or version 2 counterparts. Instead of allocating a buffer, writing the output to it, and returning a pointer to it, the new version of these functions accepts an `out` parameter containing the memory location where the host writes the output. The output is always of a size known at compilation time. The runtime execution stops with an error if `out` is outside of the range of the memory of the virtual machine.
```
burdges, line 97: How does one actually implement a host function that borrows like this?

burdges:
> The more we wait, the more other changes .. will conflict with this RFC.

Agreed, our unstable host calls for elliptic curves pass `Vec<u8>`s too for example.  I managed to get them down to only one or two allocations each, and in cases those are unavoidable, but..  They could be fixed with a PR now, given they're not stable or used anywhere.  I just do not know how a host call can avoid doing one allocation when passing a `&[u8]`.  This matters for halo2 usage.

text/0004-remove-unnecessary-allocator-usage.md
tomaka: This probably requires some changes to the Substrate procedural macro stuff (which I'm not familiar with).
This is why I mention in the "drawbacks" section that it might be difficult to implement.

koute: > This probably requires some changes to the Substrate procedural macro stuff (which I'm not familiar with). This is why I mention in the "drawbacks" section that it might be difficult to implement, but I have no idea how much.

I don't remember from the top of my head whether the macro machinery supports this right now, but yeah, we can do this; not a problem.

yjhmelody:
> > To move forward with this pull request, I would like to see some basic benchmarks being done. @yjhmelody could you please help with this?
> 
> Is something in progress?
> 
> While I agree that RFCs shouldn't necessarily be swift, if nothing is done then the ETA of this RFC is "infinity" and it would be great if it was finite. The more we wait, the more other changes (such as #31) will conflict with this RFC.
> 
> (and the reason why I originally submitted a stripped-out version is that I anticipated this happening)

Sorry,I haven't seen this message before.
I personally think there is no need to do benchmarking, it will take a lot of effort. And there is no reason not to support the runtime's own Allocator, which can greatly promote the substrate runtime ecology, and obviously it will not degrade performance very much unless you use a very poor memory allocator.

text/0004-remove-unnecessary-allocator-usage.md
tomaka: They are used to know where to store stuff. For example the pallet `Foo` stores stuff under the key `xxhash64("Foo")`.
While this is out of scope of this RFC, this is also very inefficient (literally every single storage access calculates the xxhash of the pallet) and should be moved to compile time.

tomaka: Don't quote me on that, but it is likely that nowadays the only input to the twox hash functions are strings hardcoded in the source code (pallet names and storage key names).

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,328 @@
...
+If the buffer in `out` is too small to fit the entire header name of value, only the bytes that fit are written and the rest are discarded.
+
+```wat
+(func $ext_offchain_submit_transaction_version_2
+    (param $data i64) (return i32))
+(func $ext_offchain_http_request_add_header_version_2
+    (param $request_id i32) (param $name i64) (param $value i64) (result i32))
+```
+
+The behaviour of these functions is identical to their version 1 counterpart. Instead of allocating a buffer, writing `1` or `0` in it, and returning a pointer to it, the version 2 of these functions simply return 1 or 0.
```
bkchr, line 250:
```suggestion
The behaviour of these functions is identical to their version 1 counterpart. Instead of allocating a buffer, writing `1` or `0` in it, and returning a pointer to it, the version 2 of these functions simply return `1` or `0`. While `0` means that the operation was successful. 
```

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,328 @@
...
+```
+
+This function writes [the `PeerId` of the local node](https://spec.polkadot.network/chap-networking#id-node-identities) to the memory location indicated by `out`. A `PeerId` is always 38 bytes long.
+The runtime execution stops with an error if `out` is outside of the range of the memory of the virtual machine.
+
+```wat
+(func $ext_input_size_version_1
+    (return i64))
+(func $ext_input_read_version_1
+    (param $offset i64) (param $out i64))
```
bkchr, line 277: Why not just have the `read` function and follow the way the other `read` functions are implemented?

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,328 @@
...
+
+## Unresolved Questions
+
+The changes in this RFC would need to be benchmarked. This involves implementing the RFC and measuring the speed difference.
+
+It is expected that most host functions are faster or equal speed to their deprecated counterparts, with the following exceptions:
+
+- `ext_input_size_version_1`/`ext_input_read_version_1` is inherently slower than obtaining a buffer with the entire data due to the two extra function calls and the extra copying. However, given that this only happens once per runtime call, the cost is expected to be negligible.
+
+- The `ext_crypto_*_public_keys`, `ext_offchain_network_state`, and `ext_offchain_http_*` host functions are likely slightly slower than their deprecated counterparts, but given that they are used only in offchain workers this is acceptable.
```
bkchr, line 319: `ext_offchain_network_state` doesn't exist anymore?

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,328 @@
...
+This function supercedes the `ext_offchain_local_storage_get_version_1` host function, and uses an API and logic similar to `ext_storage_read_version_2`.
+
+It reads the offchain local storage key indicated by `kind` and `key` starting at the byte indicated by `offset`, and writes the value to the [pointer-size](https://spec.polkadot.network/chap-host-api#defn-runtime-pointer-size) indicated by `value_out`.
+
+The function returns the number of bytes that were written in the `value_out` buffer. If the entry doesn't exist, a value of `-1` is returned. Given that the host must never write more bytes than the size of the buffer in `value_out`, and that the size of this buffer is expressed as a 32 bits number, a 64bits value of `-1` is not ambiguous.
+
+The runtime execution stops with an error if `value_out` is outside of the range of the memory of the virtual machine, even if the size of the buffer is 0 or if the amount of data to write would be 0 bytes.
+
+```wat
+(func $ext_offchain_network_peer_id_version_1
```
bkchr, line 266: There is no way to read the external addresses? (Not really sure we still need them)

bkchr (approved):
Sorry for the long delay. I left some nitpicks, generally looking good and we could continue with this RFC from my side.

text/0004-remove-unnecessary-allocator-usage.md
tomaka: I think that doing it this way (first getting the size then getting the content) is way more clear and simple, and more importantly avoids the risk of allocating a buffer too small and having to re-allocate it again later and copy or throw away all the data that was already read.

When reading the storage, knowing the size of the storage value would require a database access, so if reading a storage value required two function calls we would need to access the database twice, which seems not great.

When reading the input, I assume that `ext_input_size_version_1` just returns something like `self.input.len()`, so there's no problem. Also, you typically read the input only once.

text/0004-remove-unnecessary-allocator-usage.md
tomaka: I didn't realize that.

tomaka: I mention it here: https://github.com/polkadot-fellows/RFCs/pull/4#issuecomment-1730995901

I think that this function should be deprecated entirely.

text/0004-remove-unnecessary-allocator-usage.md
tomaka: I also mention it here: https://github.com/polkadot-fellows/RFCs/pull/4#issuecomment-1730995901 and I think it should be deprecated.

It's completely unclear to me how to define "an external address".

tomaka:
I've run a quick small experiment and have tried to extract the allocator out of smoldot, in other words do to smoldot the opposite of what this RFC wants to do to the runtime, and then measure the size of smoldot compiled to Wasm.
Smoldot uses [`dlmalloc`](https://crates.io/crates/dlmalloc), which is the allocator used by default by the Rust standard library.

With the allocator outside of smoldot: 2,789.9kB
With the allocator inside of smoldot: 2,794.3kB

So basically the size of the code of the allocator is 5 kiB, and that's uncompressed.

text/0004-remove-unnecessary-allocator-usage.md
tomaka: Changed

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,328 @@
...
+The same explanations apply for `ext_crypto_sr25519_public_key_version_1` and `ext_crypto_ecdsa_public_key_version_1`.
+
+Host implementers should be aware that the list of public keys (including their ordering) must not change while the runtime is running. This is most likely done by copying the list of all available keys either at the start of the execution or the first time the list is accessed.
+
+```wat
+(func $ext_offchain_http_request_start_version_2
+  (param $method i64) (param $uri i64) (param $meta i64) (result i32))
+```
+
+The behaviour of this function is identical to its version 1 counterpart. Instead of allocating a buffer, writing the request identifier in it, and returning a pointer to it, the version 2 of this function simply returns the newly-assigned identifier to the HTTP request. On failure, this function returns 0. An identifier of 0 is invalid and is reserved to indicate failure.
```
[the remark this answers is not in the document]
tomaka: Done

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,328 @@
...
+(func $ext_crypto_ecdsa_num_public_keys_version_1
+    (param $key_type_id i32) (return i32))
+(func $ext_crypto_ecdsa_public_key_version_2
+    (param $key_type_id i32) (param $key_index i32) (param $out i32))
+```
+
+The functions superceded the `ext_crypto_ed25519_public_key_version_1`, `ext_crypto_sr25519_public_key_version_1`, and `ext_crypto_ecdsa_public_key_version_1` host functions.
+
+Instead of calling `ext_crypto_ed25519_public_key_version_1` in order to obtain the list of all keys at once, the runtime should instead call `ext_crypto_ed25519_num_public_keys_version_1` in order to obtain the number of public keys available, then `ext_crypto_ed25519_public_key_version_2` repeatedly.
+The `ext_crypto_ed25519_public_key_version_2` function writes the public key of the given `key_index` to the memory location designated by `out`. The `key_index` must be between 0 (included) and `n` (excluded), where `n` is the value returned by `ext_crypto_ed25519_num_public_keys_version_1`.
```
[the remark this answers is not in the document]
tomaka: Clarified

text/0004-remove-unnecessary-allocator-usage.md
bkchr: > Don't quote me on that, but it is likely that nowadays the only input to the twox hash functions are strings hardcoded in the source code (pallet names and storage key names).

Yes this. I mean it is up to the user, but for maps the default is blake2.





> should be moved to compile time.

This is also already done.

burdges: We'll avoid any pallet dependencies with bizarre hard coded storage names then I guess.  lol

text/0004-remove-unnecessary-allocator-usage.md
bkchr: Fine by me. Maybe should be mentioned a little bit more obvious in the RFC.

text/0004-remove-unnecessary-allocator-usage.md
bkchr: I mean you could pass `out` with a zero length to only get the length of the value.

text/0004-remove-unnecessary-allocator-usage.md
tomaka: I mention it here: https://github.com/tomaka/RFCs-1/blob/89a4fcc5586ffd43741600c1105870fb85dc156f/text/0004-remove-unnecessary-allocator-usage.md#L293-L299

radkomih:
Sorry if this isn't the right place for this, but I would like to share our insights on the topic:

I'm part of a team developing runtimes in Go, and we've had to modify the compiler to ensure compatibility with the external allocator. We've implemented a custom garbage collector that integrates seamlessly with this external allocator, but this is one of the reasons we're not using the standard Go compiler (another is related to the Wasm target). 
For us, integrating the host allocator into the runtime is not solely a matter of speed, it's also about facilitating easier adoption and generalization for writing Polkadot runtimes in various languages.

+1 this change to happen.

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,326 @@
...
+The behaviour of these functions is identical to their version 1 or version 2 counterparts. Instead of allocating a buffer, writing the output to it, and returning a pointer to it, the new version of these functions accepts an `out` parameter containing the memory location where the host writes the output. The output is always of a size known at compilation time. The runtime execution stops with an error if `out` is outside of the range of the memory of the virtual machine.
+
+```wat
+(func $ext_default_child_storage_root_version_3
+    (param $child_storage_key i64) (param $out i32))
+(func $ext_storage_root_version_3
+    (param $out i32))
+```
+
+The behaviour of these functions is identical to their version 1 and version 2 counterparts. Instead of allocating a buffer, writing the output to it, and returning a pointer to it, the new versions of these functions accepts an `out` parameter containing the memory location where the host writes the output. The output is always of a size known at compilation time. The runtime execution stops with an error if `out` is outside of the range of the memory of the virtual machine.
```
yjhmelody, line 106: For writing variable-length host functions, is it really necessary to revise it?
In fact, the host always needs an Allocator to support the copy of the entrypoint parameter.
So if the new version cannot improve performance, it will only increase the complexity of use.

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,326 @@
...
+
+It is expected that most host functions are faster or equal speed to their deprecated counterparts, with the following exceptions:
+
+- `ext_input_size_version_1`/`ext_input_read_version_1` is inherently slower than obtaining a buffer with the entire data due to the two extra function calls and the extra copying. However, given that this only happens once per runtime call, the cost is expected to be negligible.
+
+- The `ext_crypto_*_public_keys`, `ext_offchain_network_state`, and `ext_offchain_http_*` host functions are likely slightly slower than their deprecated counterparts, but given that they are used only in offchain workers this is acceptable.
+
+- It is unclear how replacing `ext_storage_get` with `ext_storage_read` and `ext_default_child_storage_get` with `ext_default_child_storage_read` will impact performances.
+
+- It is unclear how the changes to `ext_storage_next_key` and `ext_default_child_storage_next_key` will impact performances.
```
yjhmelody, line 321: For writing variable-length host functions, is it really necessary to revise it?
In fact, the host always needs an Allocator to support the copy of the entrypoint parameter.
So if the new version cannot improve performance, it will only increase the complexity of use.

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,326 @@
...
+The runtime execution stops with an error if `offset` is strictly superior to the size of the input data, or if `out` is outside of the range of the memory of the virtual machine, even if the amount of data to copy would be 0 bytes.
+
+### Other changes
+
+In addition to the new host functions, this RFC proposes two changes to the runtime-host interface:
+
+- The following function signature is now also accepted for runtime entry points: `(func (result i64))`.
+- Runtimes no longer need to expose a constant named `__heap_base`.
+
+All the host functions that are being superceded by new host functions are now considered deprecated and should no longer be used.
```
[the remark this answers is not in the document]
tomaka: It means what is written: they should no longer be used. Runtimes are what use host functions. Therefore, runtimes should no longer use these host functions.

> Do you mean increasing the version number?

This text simply says that if the RFC adds `foo_v2`, then `foo_v1` should no longer be used. You can see it as increasing the version number, yes, but given that there's no semver involved it's better to think as `foo_v1` and `foo_v2` as two completely unrelated functions.

text/0004-remove-unnecessary-allocator-usage.md
yjhmelody: It's just what you are talking about:  `input to the runtime call`.

koute:
This RFC is kind of dead, so unless anyone else steps up I volunteer to push it forward.

I will implement this in Substrate and report back; let's get this over the finish line.

burdges:
As discussed above, it appears twox is sufficently depricated to ignore here.

A polkavm-ish question:  Is there some minimum overhead for a host function call?  If small enough then like [I described above](https://github.com/polkadot-fellows/RFCs/pull/4#issuecomment-1657344467) ..

One could open up the hash functions guts and provide host functions for `digest::Update::update` and `digest::FixedOutput::finalize_into`.  We'd avoid doing so many allocations even inside the runtime this way.

burdges:
These calls are purely for performance, and have no reason to transfer control to the host.  We could've machine code which gets linked in directly maybe?  I guess that'd bring adifferent auditing problem.. 

> Nevertheless, I'm not sure if we need this?

Nah.  It helps avoids some performance footguns, but whatever.

IRFT standards where those footguns look likely, like hash-to-curve and HKDF, were designed to make this not a problem, not that we support any of their hash functions anyways.

yjhmelody:
> This RFC is kind of dead, so unless anyone else steps up I volunteer to push it forward.
> 
> I will implement this in Substrate and report back; let's get this over the finish line.

I'm not free recently, but if you implement it, I can help review it :)

text/0004-remove-unnecessary-allocator-usage.md
```diff
@@ -0,0 +1,326 @@
...
+
+## Explanation
+
+### New host functions
+
+This section contains a list of new host functions to introduce.
+
+```wat
+(func $ext_storage_read_version_2
+    (param $key i64) (param $value_out i64) (param $offset i32) (result i64))
```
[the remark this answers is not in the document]
tomaka: > result could be an i32 here; even with an i32 a return value of -1 is not ambiguous as you'll physically never be able to actually allocate a buffer of such size.

I went with `i64` because I'm not really a fan of "`-1i32` is not ambiguous". If we ever switch to a 64bits runtime, then it's no longer impossible for 4 GiB to be allocated. Sure, a 4 GiB storage item would be unusually big, but it's also not in the real of completely absurd.

> correct?

Yes, but I think that your explanation builds upon a wrong assumption.

> if m == n (so either the buffer was exactly the right size or we still have some data left) then the guest enlarges the buffer and calls the host function again with a bigger offset and repeats this until the whole value is read (and if the guest guesses wrong then this could be called many times),

The idea when I designed this function is that the runtime is supposed to know what the size of the storage item is. When you're for example reading a `u32` from storage, you know it's always 4 bytes. When you're reading a `Vec<something>`, you can first read the length prefix in front of the `Vec` (which realistically shouldn't be more than 5 bytes) and then allocate a `Vec` of exactly the correct size.

Allocating a buffer whose size is equal to the size of the runtime storage item builds upon the assumption that you can't decode SCALE-encoded items in a streaming way. This assumption is true in practice now, but in the absolute it isn't.

To give an example, if you have a storage item whose format is `Vec<Vec<u8>>`, right now you have to allocate a big buffer that contains the entire storage item, then when decoding you copy the data within each inner `Vec`. It seems more optimal to me (but I could be wrong) to call `storage_read` once to get the size of the outer `Vec`, then twice for each inner `Vec`, and read the data of each inner `Vec<u8>`s directly its their destination buffer.

athei:
I want to raise another motivation for this RFC that wasn't mentioned here so far if I am not overlooking something: The current host side allocator has a pessimistic worst case when it comes to how much of the available memory can be allocated.  Our auditors recommended a security margin of 4x (which means we can only assume to be able to safely use 1/4 of the memory). 

For memory intense pallets like contracts this is a problem. There might be other use cases which can benefit from more memory.

A memory allocator with a better worst case would allow us to use more memory.

s0me0ne-unkn0wn:
Closing as superseded by #145

PR #5, opened 2023-07-06
Coretime interface

gavofyork:
In the Agile Coretime model of the Polkadot Ubiquitous Computer, as proposed in RFC-1 and RFC-3, it is necessary for the allocating parachain (envisioned to be one or more pallets on a specialised Brokerage System Chain) to communicate the core assignments to the Relay-chain, which is responsible for ensuring those assignments are properly enacted.

This is a proposal for the interface which will exist around the Relay-chain in order to communicate this information and instructions.


text/0005-coretime-interface.md
```diff
@@ -0,0 +1,176 @@
...
+Prototype:
+
+```
+fn notify_revenue(
+    until: BlockNumber,
+    revenue: Balance,
+)
+```
+
+Provide the amount of revenue accumulated from Instantaneous Coretime Sales from Relay-chain block number `last_until` to `until`, not including `until` itself. `last_until` is defined as being the `until` argument of the last `notify_revenue` message sent.
```
rphmeier, line 150: (or 0, for the first call)

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,176 @@
...
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 06 July 2023                                                                                |
+| **Description** | Interface for manipulating the usage of cores on the Polkadot Ubiquitous Computer. |
+| **Authors**     | Gavin Wood                                                                                  |
+
+
+## Summary
+
+In the Agile Coretime model of the Polkadot Ubiquitous Computer, as proposed in RFC-1 and RFC-4, it is necessary for the allocating parachain (envisioned to be one or more pallets on a specialised Brokerage System Chain) to communicate the core assignments to the Relay-chain, which is responsible for ensuring those assignments are properly enacted.
```
rphmeier, line 12:
```suggestion
In the Agile Coretime model of the Polkadot Ubiquitous Computer, as proposed in RFC-1 and RFC-3, it is necessary for the allocating parachain (envisioned to be one or more pallets on a specialised Brokerage System Chain) to communicate the core assignments to the Relay-chain, which is responsible for ensuring those assignments are properly enacted.
```

#4 is about host functions

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,176 @@
...
+
+Instructs the Relay-chain to ensure that the core indexed as `core` is utilised for a number of assignments in specific ratios given by `assignment` starting as soon after `begin` as possible. Core assignments take the form of a `CoreAssignment` value which can either task the core to a `ParaId` value or indicate that the core should be used in the Instantaneous Pool. Each assignment comes with a ratio value, represented as the numerator of the fraction with a denominator of 57,600.
+
+If `end_hint` is `Some` and the inner is greater than the current block number, then the Relay-chain should optimize in the expectation of receiving a new `assign_core(core, ...)` message at or prior to the block number of the inner value. Specific functionality should remain unchanged regardless of the `end_hint` value.
+
+On the choice of denominator: 57,600 is a highly composite number which factors into: 2, 2, 2, 2, 2, 2, 2, 2, 3, 3, 5, 5. By using it as the denominator we allow for various useful fractions to be perfectly represented including thirds, quarters, fifths, tenths, 80ths, percent and 256ths.
```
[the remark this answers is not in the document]
gavofyork:
```suggestion
On the choice of denominator: 57,600 is a very composite number which factors into: 2, 2, 2, 2, 2, 2, 2, 2, 3, 3, 5, 5. By using it as the denominator we allow for various useful fractions to be perfectly represented including thirds, quarters, fifths, tenths, 80ths, percent and 256ths.
```

text/0005-coretime-interface.md
gavofyork:
```suggestion
Provide the amount of revenue accumulated from Instantaneous Coretime Sales from Relay-chain block number `last_until` to `until`, not including `until` itself. `last_until` is defined as being the `until` argument of the last `notify_revenue` message sent, or zero for the first call.
```

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,176 @@
...
+
+Instructs the Relay-chain to ensure that the core indexed as `core` is utilised for a number of assignments in specific ratios given by `assignment` starting as soon after `begin` as possible. Core assignments take the form of a `CoreAssignment` value which can either task the core to a `ParaId` value or indicate that the core should be used in the Instantaneous Pool. Each assignment comes with a ratio value, represented as the numerator of the fraction with a denominator of 57,600.
+
+If `end_hint` is `Some` and the inner is greater than the current block number, then the Relay-chain should optimize in the expectation of receiving a new `assign_core(core, ...)` message at or prior to the block number of the inner value. Specific functionality should remain unchanged regardless of the `end_hint` value.
+
+On the choice of denominator: 57,600 is a very composite number which factors into: 2, 2, 2, 2, 2, 2, 2, 2, 3, 3, 5, 5. By using it as the denominator we allow for various useful fractions to be perfectly represented including thirds, quarters, fifths, tenths, 80ths, percent and 256ths.
```
gavofyork, line 123:
```suggestion
On the choice of denominator: 57,600 is a very composite number which factors into: 2 ** 8, 3 ** 2, 5 ** 2. By using it as the denominator we allow for various useful fractions to be perfectly represented including thirds, quarters, fifths, tenths, 80ths, percent and 256ths.
```

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+Indicate that from this block onwards, the range of acceptable values of the `core` parameter of `assign_core` message is `[0, count)`. `assign_core` will be a no-op if provided with a value for `core` outside of this range.
+
+#### `notify_revenue_info`
+
+Prototype:
+
+```
+fn notify_revenue(
```
gavofyork, line 144:
```suggestion
fn notify_revenue_info(
```

burdges:
We'll need upward messages for full parachains who adjust their core usage but..

Afaik there is no reason to use upward messages in the scheduling system for on-demand parachains.  A payment on the relay chain can increase the balance, and enqueue a message for the parachain, then whenever the parachain makes a block it can pull from this balance and do whatever it does with the message.

gavofyork:
We're planning to remove as much functionality as possible from the Relay-chain, including currency.

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,176 @@
...
+type PartsOf57600 = u16;
+enum CoreAssignment {
+    InstantaneousPool,
+    Task(ParaId),
+}
+fn assign_core(
+    core: CoreIndex,
+    begin: BlockNumber,
+    assignment: Vec<(CoreAssignment, PartsOf57600)>,
+    end_hint: Option<BlockNumber>,
```
rphmeier, line 103: Noting here - the main purpose of the `None` end-hint is that it allows the broker chain to skip sending this message sometimes. 

Some users of #1 will likely split their regions up and defer their `allocate` calls until the last possible moment. When the next-up region has the same assignment as the previous, there's no need to send a new `assign_core` message

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Summary
+
+In the Agile Coretime model of the Polkadot Ubiquitous Computer, as proposed in RFC-1 and RFC-3, it is necessary for the allocating parachain (envisioned to be one or more pallets on a specialised Brokerage System Chain) to communicate the core assignments to the Relay-chain, which is responsible for ensuring those assignments are properly enacted.
+
+This is a proposal for the interface which will exist around the Relay-chain in order to communicate this information and instructions.
+
+## Motivation
+
+The overall motivation for splitting out functions from the Relay-chain onto System parachains is well understood. An well-understood interface is necessary for ensuring multiple chains are able to coordinate their efforts.
```
rphmeier, line 18:
```suggestion
The background motivation for this interface is splitting out coretime allocation functions and secondary markets from the Relay-chain onto System parachains. A well-understood and general interface is necessary for ensuring the Relay-chain receives coretime allocation instructions from one or more System chains without introducing dependencies on the implementation details of either side.
```

Added slightly more motivation details for posterity / the general technical audience

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,176 @@
+# RFC-5: Coretime Interface
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 06 July 2023                                                                                |
+| **Description** | Interface for manipulating the usage of cores on the Polkadot Ubiquitous Computer. |
+| **Authors**     | Gavin Wood                                                                                  |
```
rphmeier, line 7: We spent quite a lot of time on `fn assign_core` and `request` core parts of things together - formal request for an author credit here

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,176 @@
...
+
+In the Agile Coretime model of the Polkadot Ubiquitous Computer, as proposed in RFC-1 and RFC-3, it is necessary for the allocating parachain (envisioned to be one or more pallets on a specialised Brokerage System Chain) to communicate the core assignments to the Relay-chain, which is responsible for ensuring those assignments are properly enacted.
+
+This is a proposal for the interface which will exist around the Relay-chain in order to communicate this information and instructions.
+
+## Motivation
+
+The overall motivation for splitting out functions from the Relay-chain onto System parachains is well understood. An well-understood interface is necessary for ensuring multiple chains are able to coordinate their efforts.
+
+## Requirements
```
rphmeier, line 20: style: with the template in #2 , this would be a sub-section of "Motivation", as in the motivation for this RFC is to solve a problem with these requirements.

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,176 @@
...
+```
+type PartsOf57600 = u16;
+enum CoreAssignment {
+    InstantaneousPool,
+    Task(ParaId),
+}
+fn assign_core(
+    core: CoreIndex,
+    begin: BlockNumber,
+    assignment: Vec<(CoreAssignment, PartsOf57600)>,
```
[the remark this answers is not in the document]
rphmeier: Re: the requirement that "Worst case usage of the interface MUST NOT cause the Polkadot system to fail",

very heavy and frequent `assign_core` calls would either have to be ignored, buffered, or bounded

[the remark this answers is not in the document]
gavofyork: Added a section on reasonable limits.

text/0005-coretime-interface.md
```diff
@@ -71,6 +70,8 @@ fn request_revenue_at(
 
 Requests that the Relay-chain send a `notify_revenue` message back at or soon after Relay-chain block number `when` whose `until` parameter is equal to `when`.
 
+The period in to the past which `when` is allowed to be may be limited; if so the limit should be understood on a channel outside of this proposal. In the case that the request cannot be servied because `when` is too old a block then a `notify_revenue` message must still be returned, but its `revenue`` field may be `None`.
```
antonva, line 73:
```suggestion
The period in to the past which `when` is allowed to be may be limited; if so the limit should be understood on a channel outside of this proposal. In the case that the request cannot be serviced because `when` is too old a block then a `notify_revenue` message must still be returned, but its `revenue` field may be `None`.
```

gavofyork:
blake2_256 hash of the 33d45a8 text is `0x1a8f351c8f9cc3f5d6587129324bba6d7d0d086af1275c875f4a18af4651f792`.

gavofyork:
blake2_256 hash of `System.remark("APPROVE_RFC(0005,1a8f351c8f9cc3f5d6587129324bba6d7d0d086af1275c875f4a18af4651f792)")` is `0x8a5685393b5fc0845a5356b772ac8320cfbc2f05aed75dcd60355eab14aa081e`; preimage length is 86 bytes.

image

gavofyork:
Vote here: https://collectives.subsquare.io/fellowship/referendum/11

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,183 @@
...
+    end_hint: Option<BlockNumber>,
+)
+```
+
+Requirements:
+
+```
+assert!(core < core_count);
+assert!(targets.iter().map(|x| x.0).is_sorted());
+assert_eq!(targets.iter().map(|x| x.0).unique().count(), targets.len());
```
bkchr, line 113: But doesn't this prevent interleaved execution of two parachains? Because then we would have multiple times the same `CoreAssignment::Task` in this list?

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,183 @@
...
+assert_eq!(targets.iter().map(|x| x.1).sum(), 57600);
+```
+
+Where:
+- `core_count` is assumed to be the sole parameter in the last received `notify_core_count` message.
+
+Instructs the Relay-chain to ensure that the core indexed as `core` is utilised for a number of assignments in specific ratios given by `assignment` starting as soon after `begin` as possible. Core assignments take the form of a `CoreAssignment` value which can either task the core to a `ParaId` value or indicate that the core should be used in the Instantaneous Pool. Each assignment comes with a ratio value, represented as the numerator of the fraction with a denominator of 57,600.
```
bkchr, line 120: I don't get how `assignment` should work? How does the assignment and fraction map to scheduling the assignment to the core? 

I mean for things like `[(Task(1), 0.5), (Task(2), 0.5)]` task 1 and task 2 will be running interleaved all the time?

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,183 @@
...
+    assignment: Vec<(CoreAssignment, PartsOf57600)>,
+    end_hint: Option<BlockNumber>,
+)
+```
+
+Requirements:
+
+```
+assert!(core < core_count);
+assert!(targets.iter().map(|x| x.0).is_sorted());
```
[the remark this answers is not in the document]
rphmeier: They don't strictly need to be, but it's better to start with a more constrained interface

text/0005-coretime-interface.md
rphmeier: Interleaving would be something like

`targets: Vec<(Task(1), 1/2), (Task(2), 1/2)>`

text/0005-coretime-interface.md
rphmeier: > How does the assignment and fraction map to scheduling the assignment to the core?

It means that the relay-chain should allow the task/assignment to use that proportion of the core's resources. It's better to leave it general like this, because this definition also allows tasks to use 1/2 of the core every relay-chain block, when we get support for having multiple candidates per core at a time.

And scheduling tasks to _specific_ relay-chain blocks is actually really bad - this formulation does work nicely with e.g. #3 , which is a probabilistic scheduler

> I mean for things like [(Task(1), 0.5), (Task(2), 0.5)] task 1 and task 2 will be running interleaved all the time?

Yes, or `(Task(1), 0.25), (Task(2), 0.5), (Instantaneous, 0.25)` would mean that:
  * Task 1 gets to use 1/4 of the core's resources for this timeframe (1 block out of every 4 relay chain blocks)
  * Task 2 gets to use 1/2 of the core's resources (2 blocks out of every 4)
  * Instantaneous gets to allocate 1 block every 4 relay chain blocks.

bkchr: > It's better to leave it general like this, because this definition also allows tasks to use 1/2 of the core every relay-chain block, when we get support for having multiple candidates per core at a time.

But how would that look like? Currently the fraction expresses how much blocks a task can build for a given time range. I assume this time range will be `TIMESLICE` and both sides relay chain/brocker chain will be aware of this? 

We will need something to tell the relay chain that a certain task actually needs only 1/2 of a core and not 1/2 of the time range.

text/0005-coretime-interface.md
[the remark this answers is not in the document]
bkchr: I mean it will depend on the implementation of how to sort `CoreAssignment`, but this will probably mean that lower task numbers will be scheduled earlier?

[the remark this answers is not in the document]
bkchr: I really don't get the need for sorting it. Especially when we want to express something like `Task(1), Task(2)` should both use 1/2 of the same core at the same block.

text/0005-coretime-interface.md
```diff
@@ -0,0 +1,183 @@
...
+```
+fn credit_account(
+    who: AccountId,
+    amount: Balance,
+)
+```
+
+Instructs the Relay-chain to add the `amount` of DOT to the Instantaneous Coretime Market Credit account of `who`.
+
+It is expected that Instantaneous Coretime Market Credit on the Relay-chain is NOT transferrable and only redeemable when used to assign cores in the Instantaneous Coretime Pool.
```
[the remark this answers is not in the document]
gavofyork: you can't

text/0005-coretime-interface.md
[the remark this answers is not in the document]
gavofyork: i have no opinion at all here.

text/0005-coretime-interface.md
gavofyork: It's designed to be probabilistic - over a sufficiently long period it would be scheduled 1/2 of the time, and the Relay-chain is expected to be as fair as it can be, but practical constraints preclude the ability to state anything that is "perfectly fair all the time". We don't explicitly specify a timerange; the relay-chain is expected to just maximise the fairness over a minimal time range.

rphmeier: > But how would that look like? Currently the fraction expresses how much blocks a task can build for a given time range. I assume this time range will be TIMESLICE and both sides relay chain/brocker chain will be aware of this?

While it would be valid for the relay-chain side to interpret this interface as giving each chain 1/2 of the timeslice, that would be a pretty bad scheduler. Scheduler implementations (like #3) should probably aim for probabilistic interleaving.

text/0005-coretime-interface.md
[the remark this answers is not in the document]
bkchr: Given the comments from @rphmeier below, I see why the sorting doesn't has any effect

bkchr (approved):

PR #7, opened 2023-07-11
System Collator Set Selection

joepetrowski:
As discussed in the [roadmap](https://github.com/paritytech/roadmap/issues/34) with corresponding implementation issues, converting this to an RFC. This proposes a means of selecting collator sets on system chains.


text/RFC-0007-System Collator Selection.md
```diff
@@ -0,0 +1,162 @@
...
+Unlike the fixed-bond mechanism that fills up its Candidates, the election mechanism ensures that
+anyone can join the collator set by placing the `Nth` highest bond.
+
+### Set Size
+
+In order to achieve the requirements listed under _Motivation_, it is reasonable to have
+approximately:
+
+- 20 collators per system chain,
+- of which 15 are Invulnerable, and
```
xlc, line 104: any explanations of those numbers?
I would expect low number of invulnerable as it needs to be managed by governance and therefore high bar of entry.

joepetrowski: The numbers were recommendations from SR Labs, however I find them quite reasonable. Of course it will take time to get this many collators, and there may be situations where it makes sense to have more collators on a particular chain. But I think these are good general purpose targets.

With respect to the high bar of entry, that's the point. Due to the high number (supply) of system parachains, and low profitability (demand) for being a collator, the open election would likely be a very low bar of entry, making it vulnerable to malicious takeover. 

As the Motivation section says,

> "Another problem with economic scalability relates to the increasing number of system chains, and corresponding increase in need for collators (i.e., increase in collator slots). "Good" (highly available, non-censoring) collators will not want to compete in elections on many chains when they could use their resources to compete in the more profitable validator election. Such dilution decreases the required bond on each chain, leaving them vulnerable to takeover by hostile collator groups."

xlc: ok I see the reasons now. However this to me is more of a short term workaround that there is not enough incentive to be system chain collector. Any plan to fix that? Relying on governance to gate keep things should be a last resort.

joepetrowski: I don't see governance as a _last_ resort, just as one of many tools. Our initial discussion was around creating incentives (see [forum](https://forum.polkadot.network/t/economic-model-for-system-para-collators/1010)), but it's really not scalable. We expect many more collators than validators in the system, and as you know validators are expensive (e.g. 50 system chains x 20 collators = 1000 collators). Likewise, collation is not a secure activity and doesn't warrant large incentives.

The incentives for governance-selected collators is often around building reputation as, for example, a validator.

Also from the Motivation:

> "While the network as a whole uses staking (and inflationary rewards) to attract validators, collators face different challenges in scale and have lower security assumptions than validators. Regarding scale, there exist many system chains, and it is economically expensive to pay collators a premium. Likewise, any staked DOT for collation is _not_ staked for validation. Since collator sets do not need to meet Byzantine Fault Tolerance criteria, staking as the primary mechanism for collator selection would remove stake that is securing BFT assumptions, making the network less secure."

pepyakin:
Would be great if the RFC file is renamed according to the template.

joepetrowski:
> Would be great if the RFC file is renamed according to the template.

:+1: done

pmensik:
I am missing a description of a mechanism by which the Invulnerable collators will be selected in the future - there were various ideas in the collator group but we ended up using a quite simple solution based on fairness and gentleman's agreement. Is this mechanism going to stay the same in the future or should we get general governance involved when new system chains come out (like Coretime)?

joepetrowski:
There is no way to enforce on-chain what someone proposes to governance. Any individual or group can use any mechanism to propose to add or remove someone as an Invulnerable, and the referendum process will accept or reject the proposal. "Fairness" is quite subjective. Therefore, it is not part of the RFC.

text/0007-system-collator-selection.md
```diff
@@ -0,0 +1,162 @@
...
+will be selected as part of the collator set every session. Operations relating to the management
+of the Invulnerables are done through privileged, governance origins. The implementation should
+maintain an API for adding and removing Invulnerable collators.
+
+In addition to Invulnerables, there are also open slots for "Candidates". Anyone can register as a
+Candidate by placing a fixed bond. However, with a fixed bond and fixed number of slots, there is
+an obvious selection problem: The slots fill up without any logic to replace their occupants.
+
+This RFC proposes that the collator selection protocol allow Candidates to increase (and decrease)
+their individual bonds, sort the Candidates according to bond, and select the top `N` Candidates.
```
bkchr, line 78: Why are we not choosing randomly? Why based on the highest bond? What is the reason behind this?

bkchr (commented):
Looking good, I personally would not select on highest bound and instead select them randomly.

text/0007-system-collator-selection.md
joepetrowski: I don't really expect bonds to be high since there is not much incentive, but in case someone does have a strong incentive (e.g. thinks their transactions are being censored), they should have a path to getting selected.

With fixed bond and random selection, there are two problems (unless you see a better solution):

1. We can limit the number of candidates like we do now and select randomly. But if the set is full, how does a new candidate join? With a fixed bond, there's no comparison to justify kicking someone out.
2. We can have an unlimited number of candidates, but then people can just use multiple accounts and place the fixed bond many times, increasing their probability of being selected.

(2) could be an option, the selection rate would map to the bond amount (represented by number of accounts). It seems less efficient though as it requires running nodes for all the accounts, whereas top N requires only one node per candidate entity.

bkchr: > 2\. We can have an unlimited number of candidates, but then people can just use multiple accounts and place the fixed bond many times, increasing their probability of being selected.

Yeah that is a valid point. 

So, yeah scratch my idea.

joepetrowski:
/rfc process 0x37ae0558f4b008f9e7e421488b3466bfb2ea3510a1a58a2d40beff681e84cc42

joepetrowski:
/rfc process 0x9ad960de812071c892cc545743b4ffb85fcdaf969f1b77429ce9c1d4219f2f4b

rzadp:
> Error: Merge commits are not allowed on this repository.

The bot doesn't specify the merge option - apparently it defaults to a merge commit which is disabled in this repo.

rzadp:
/rfc process 0x9ad960de812071c892cc545743b4ffb85fcdaf969f1b77429ce9c1d4219f2f4b

PR #8, opened 2023-07-14
Store parachain bootnodes in relay chain DHT

tomaka:
[Rendered](https://github.com/tomaka/RFCs-1/blob/rfc-8/text/0008-parachain-bootnodes-dht.md).

### DIFF
--- a/text/0008-parachain-bootnodes-dht.md
+++ b/text/0008-parachain-bootnodes-dht.md
@@ -1,0 +1,146 @@
+# RFC-0008: Store parachain bootnodes in relay chain DHT
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 2023-07-14                                                                                  |
+| **Description** | Parachain bootnodes shall register themselves in the DHT of the relay chain                 |
+| **Authors**     | Pierre Krieger                                                                              |
+
+## Summary
+
+The full nodes of the Polkadot peer-to-peer network maintain a distributed hash table (DHT), which is currently used for full nodes discovery and validators discovery purposes.
+
+This RFC proposes to extend this DHT to be used to discover full nodes of the parachains of Polkadot.
+
+## Motivation
+
+The maintenance of bootnodes has long been an annoyance for everyone.
+
+When a bootnode is newly-deployed or removed, every chain specification must be updated in order to take the update into account. This has lead to various non-optimal solutions, such as pulling chain specifications from GitHub repositories.
+When it comes to RPC nodes, UX developers often have trouble finding up-to-date addresses of parachain RPC nodes. With the ongoing migration from RPC nodes to light clients, similar problems would happen with chain specifications as well.
+
+Since the PeerId and addresses of bootnodes needs to be stable, extra maintenance work is required from the chain maintainers. For example, they need to be extra careful when migrating nodes within their infrastructure. In some situations, bootnodes are put behind domain names, which also requires maintenance work.
+
+Because the list of bootnodes in chain specifications is so annoying to modify, the consequence is that the number of bootnodes is rather low (typically between 2 and 15). In order to better resist downtimes and DoS attacks, a better solution would be to use every node of a certain chain as potential bootnode, rather than special-casing some specific nodes.
+
+While this RFC doesn't solve these problems for relay chains, it aims at solving it for parachains by storing the list of all the full nodes of a parachain on the relay chain DHT.
+
+Assuming that this RFC is implemented, and that light clients are used, deploying a parachain wouldn't require more work than registering it onto the relay chain and starting the collators. There wouldn't be any need for special infrastructure nodes anymore.
+
+## Stakeholders
+
+This RFC has been opened on my own initiative because I think that this is a good technical solution to a usuability problem that many people are encountering and that they don't realize can be solved.
+
+## Explanation
+
+The content of this RFC only applies for parachains and parachain nodes that are "Substrate-compatible". It is in no way mandatory for parachains to comply to this RFC.
+
+Note that "Substrate-compatible" is very loosely defined as "implements the same mechanisms and networking protocols as Substrate". The author of this RFC believes that "Substrate-compatible" should be very precisely specified, but there is controversy on this topic.
+
+While a lot of this RFC concerns the implementation of parachain nodes, it makes use of the resources of the Polkadot chain, and as such it is important to describe them in the Polkadot specification.
+
+This RFC adds two mechanisms: a registration in the DHT, and a new networking protocol.
+
+### DHT provider registration
+
+This RFC heavily relies on the functionnalities of the Kademlia DHT already in use by Polkadot.
+You can find a link to the specification [here](https://github.com/libp2p/specs/tree/master/kad-dht).
+
+Full nodes of a parachain registered on Polkadot should register themselves onto the Polkadot DHT as the providers of a key corresponding to the parachain that they are serving, as described in [the `Content provider advertisement` section](https://github.com/libp2p/specs/tree/master/kad-dht#content-provider-advertisement) of the specification. This uses the `ADD_PROVIDER` system of libp2p-kademlia.
+
+This key is: `sha256(concat(scale_compact(para_id), randomness))` where the value of `randomness` can be found in the `randomness` field when calling the `BabeApi_currentEpoch` function.
+For example, for a `para_id` equal to 1000, and at the time of writing of this RFC (July 14th 2023 at 09:13 UTC), it is `sha(0xa10f12872447958d50aa7b937b0106561a588e0e2628d33f81b5361b13dbcf8df708)`, which is equal to `0x483dd8084d50dbbbc962067f216c37b627831d9339f5a6e426a32e3076313d87`.
+
+In order to avoid downtimes when the key changes, parachain full nodes should also register themselves as a secondary key that uses a value of `randomness` equal to the `randomness` field when calling `BabeApi_nextEpoch`.
+
+Implementers should be aware that their implementation of Kademlia might already hash the key before XOR'ing it. The key is not meant to be hashed twice.
+
+The compact SCALE encoding has been chosen in order to avoid problems related to the number of bytes and endianness of the `para_id`.
+
+### New networking protocol
+
+A new request-response protocol should be added, whose name is `/91b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3/paranode` (that hexadecimal number is the genesis hash of the Polkadot chain, and should be adjusted appropriately for Kusama and others).
+
+The request consists in a SCALE-compact-encoded `para_id`. For example, for a `para_id` equal to 1000, this is `0xa10f`.
+
+Note that because this is a request-response protocol, the request is always prefixed with its length in bytes. While the body of the request is simply the SCALE-compact-encoded `para_id`, the data actually sent onto the substream is both the length and body.
+
+The response consists in a protobuf struct, defined as:
+
+```
+syntax = "proto2";
+
+message Response {
+    // Peer ID of the node on the parachain side.
+    bytes peer_id = 1;
+
+    // Multiaddresses of the parachain side of the node. The list and format are the same as for the `listenAddrs` field of the `identify` protocol.
+    repeated bytes addrs = 2;
+
+    // Hash of the root node of the trie of the parachain.
+    bytes storageTrieRootHash = 3;
+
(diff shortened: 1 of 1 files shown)


text/0008-parachain-bootnodes-dht.md
```diff
@@ -0,0 +1,148 @@
...
+
+Because the list of bootnodes in chain specifications is so annoying to modify, the consequence is that the number of bootnodes is rather low (typically between 2 and 15). In order to better resist downtimes and DoS attacks, a better solution would be to use every node of a certain chain as potential bootnode, rather than special-casing some specific nodes.
+
+While this RFC doesn't solve these problems for relay chains, it aims at solving it for parachains by storing the list of all the full nodes of a parachain on the relay chain DHT.
+
+Assuming that this RFC is implemented, and that light clients are used, deploying a parachain wouldn't require more work than registering it onto the relay chain and starting the collators. There wouldn't be any need for special infrastructure nodes anymore.
+
+## Stakeholders
+
+This RFC has been opened on my own initiative because I think that this is a good technical solution to a usuability problem that many people are encountering and that they don't realize can be solved.
```
rphmeier, line 34:
```suggestion
This RFC has been opened on my own initiative because I think that this is a good technical solution to a usability problem that many people are encountering and that they don't realize can be solved.
```

text/0008-parachain-bootnodes-dht.md
```diff
@@ -0,0 +1,148 @@
...
+
+Note that "Substrate-compatible" is very loosely defined as "implements the same mechanisms and networking protocols as Substrate". The author of this RFC believes that "Substrate-compatible" should be very precisely specified, but there is controversy on this topic.
+
+While a lot of this RFC concerns the implementation of parachain nodes, it makes use of the resources of the Polkadot chain, and as such it is important to describe them in the Polkadot specification.
+
+This RFC adds two mechanisms: a registration in the DHT, and a new networking protocol.
+
+### DHT provider registration
+
+This RFC heavily relies on the functionnalities of the Kademlia DHT already in use by Polkadot.
```
rphmeier, line 48:
```suggestion
This RFC heavily relies on the functionalities of the Kademlia DHT already in use by Polkadot.
```

text/0008-parachain-bootnodes-dht.md
```diff
@@ -0,0 +1,148 @@
...
+
+Full nodes of a parachain registered on Polkadot should register themselves onto the Polkadot DHT as the providers of a key corresponding to the parachain that they are serving, as described in [the `Content provider advertisement` section](https://github.com/libp2p/specs/tree/master/kad-dht#content-provider-advertisement) of the specification. This uses the `ADD_PROVIDER` system of libp2p-kademlia.
+
+This key is: `sha256(concat(scale_compact(para_id), randomness))` where the value of `randomness` can be found in the `randomness` field when calling the `BabeApi_currentEpoch` function.
+For example, for a `para_id` equal to 1000, and at the time of writing of this RFC (July 14th 2023 at 09:13 UTC), it is `sha(0xa10f12872447958d50aa7b937b0106561a588e0e2628d33f81b5361b13dbcf8df708)`, which is equal to `0x483dd8084d50dbbbc962067f216c37b627831d9339f5a6e426a32e3076313d87`.
+
+In order to avoid downtimes when the key changes, parachain full nodes should also register themselves as a secondary key that uses a value of `randomness` equal to the `randomness` field when calling `BabeApi_nextEpoch`.
```
rphmeier, line 56:
```suggestion
In order to avoid downtime when the key changes, parachain full nodes should also register themselves as a secondary key that uses a value of `randomness` equal to the `randomness` field when calling `BabeApi_nextEpoch`.
```

tomaka:
I've found a pretty big issue with this proposal, which is that the `storageTrieRootHash` value that can be queried can't be trusted, and thus that the genesis hash of the chain can't be trusted.

I initially just wanted to add a note in the RFC saying that the calculated genesis hash must not be used for purposes other than determining the networking protocol names.
Unfortunately, the JSON-RPC API has functions to query the genesis hash, and so if a light client connects to a parachain in a chain-spec-less way (through just a relay chain and `para_id`), it has no way to implement the JSON-RPC API properly.

We can maybe simply remove these JSON-RPC functions, except that when you submit a transaction you need to include the genesis hash in it.

I can see two ways of solving this problem, I'm not sure right now which way is best:

- Give up for now on the idea of connecting to a parachain without a chain spec. Remove the `stateTrieRootHash` and `forkId` field from the new networking protocol, as they would be loaded from the chain spec. The RFC would simply make it possible to find the bootnodes of a parachain, but not to connect to said parachain without a chain spec.

- Double down on the idea of connecting to a parachain without a chain spec, by remove the JSON-RPC functions that allow knowing the genesis hash, and instead JSON-RPC clients should read `System::BlockHash[0]` in order to know the genesis hash. Given that the runtime verifies the genesis hash when submitting a transaction, we know that this genesis hash is always available in storage and just needs to be queried in a way or the other.

tomaka:
I ended up going for the first option, but I have also opened https://github.com/paritytech/json-rpc-interface-spec/pull/61 in order to make it future-proof.

rphmeier:
> Give up for now on the idea of connecting to a parachain without a chain spec. Remove the stateTrieRootHash and forkId field from the new networking protocol, as they would be loaded from the chain spec. The RFC would simply make it possible to find the bootnodes of a parachain, but not to connect to said parachain without a chain spec.

It seems that light clients will always need some form of chain spec, but that it only needs a few piecse of information:
  * Para-ID
  * Genesis Hash
  * Fork ID
 
Given the weakly synchronous model it would also make sense for light clients to have additional information such as a recent (~1 month) finalized block hash, for maximum deterrence of long range attacks.

tomaka:
> It seems that light clients will always need some form of chain spec, but that it only needs a few piecse of information:
>
>    Para-ID
>    Genesis Hash
>    Fork ID

For now, yes.
Also needed is the `properties` field (which PolkadotJS uses to know the token name and number of decimals (I've always thought that this should be moved on chain)).

> Given the weakly synchronous model it would also make sense for light clients to have additional information such as a recent (~1 month) finalized block hash, for maximum deterrence of long range attacks.

I don't understand why? Isn't it enough to look into the relay chain which parablock is finalized?

gavofyork:
Properties could definitely be placed in on-chain metadata.

tomaka:
> Para ID could too, no?

The light client looks into the relay chain (and thus needs to know the para ID) in order to know which of the parachain blocks is canonical.
But if we put the para ID on chain, then in order to read the para ID from the storage, the light client first needs to known which parachain block is canonical.
In other words it's a chicken and egg problem.

rphmeier:
> I don't understand why? Isn't it enough to look into the relay chain which parablock is finalized?

Technically, validator sets can "go back in time" after their stake is withdrawn and create an alternative chain. This puts an expiration date on finality's trustworthiness, and is the case in all PoS blockchains. 

Most of the time this is not a problem in practice, but in theory the correct thing to do is make sure that the node has some reference block for finality (which it either gets by synchronizing at least once every month, or has hardcoded in its chain spec).

Earlier I said it'd be a parachain block hash, but it could just as well be a relay chain block hash to use as a starting point for finality.

tomaka:
> Earlier I said it'd be a parachain block hash, but it could just as well be a relay chain block hash to use as a starting point for finality.

That's what I'm currently going for in smoldot. It seems easier to me to update the relay chain specification with an up-to-date block from time to time, rather than update the chain specs of every parachains.

Substrate-connect hardcodes the 4 relay chains (Polkadot, Kusama, Westend, Rococo), and is meant to publish a new version at a regular interval containing an up-to-date chain spec for these four relay chains (although the publishing doesn't really happen in practice because it's a semi-manual process and nobody is doing it). On the other hand, the parachains aren't hardcoded. With this solution, dApps/uApps don't need to deal with this problem.

> I suppose we can agree that light clients need some minimal chain spec?
> 
>     * Para ID
> 
>     * Genesis Hash
> 
>     * Relay Chain Spec
> 
>     * Properties (could be moved on-chain for later)
> 
>     * Finalized relay-chain block hash (nice-to-have)

The finalized relay chain block hash would be contained the relay chain spec, and you're missing the `forkId`, but otherwise yes.

bkchr:
Could we not store the `genesis_hash` in the `response` as well? We only need it to connect to the networking protocols. The `genesis_hash` is case of `FRAME` is stored on chain and we can fetch it from there and ensure that it is the one we got in the initial `response`.

tomaka:
> Could we not store the `genesis_hash` in the `response` as well? We only need it to connect to the networking protocols. The `genesis_hash` is case of `FRAME` is stored on chain and we can fetch it from there and ensure that it is the one we got in the initial `response`.

What you suggest is correct, but in practice it's very complicated to implement.

The problem is: what to do if multiple parachain bootnodes return multiple different `genesis_hash`es?
It is in principle possible to write code so that it uses a separate networking protocol name for each parachain bootnode, which avoids the problem, but that's pretty complicated.

The way I was going with this RFC is to start with a minimal version, and then maybe add the `genesis_hash` and `fork_id` to the response later, but I'm not super confident in that decision and we could indeed maybe just add them now.

tomaka:
Also, while the `genesis_hash` can be verified at some point, the `fork_id` can't.

This means that it's possible for multiple different parachain bootnodes to return multiple different `fork_id`s, and to all be successfully reachable through their own individual `fork_id`.
This complicates the code, but isn't a problem by itself. What is a problem, however, is which `fork_id` do you use when you discover the other parachain nodes through Kademlia? The answer is obviously to use the same `fork_id` as the node that was queried, but since the same node can be discovered multiple times, that means that the same node can have multiple different `fork_id`s, and it basically becomes a mess.

tomaka:
I've pushed another commit that adds back `genesis_hash` and `fork_id`, as the cost of adding them seems minimal to me, and adding them now removes the cost of going through another RFC and implementation+upgrade phase in the future (in particular, parachain builders can be pretty slow to upgrade their Substrate version).

I would like to submit this for voting, but there's still the unresolved question of what to do when it comes to obtaining randomness. Should we use the `BabeApi_currentEpoch` and `BabeApi_nextEpoch` runtime APIs, as the RFC proposes? Or should it be separate new functions?

rphmeier:
I would 'aye' this in its current state

tomaka:
I can't seem to be able to create a referendum ("bad origin" error), most likely because I'm not in the fellowship.

tomaka:
I wouldn't mind some help or guidance with this administrative process

tomaka:
Just wanted to repeat that I can't find any proper documentation or explanation about how to proceed, and this PR will be stuck for as long as nobody helps me with this bureaucracy.

tomaka:
/rfc process 0x8bfe18c1c9c892515872a551b9f2f71216a522e2bc1a81e26a2d03e167208aa4

tomaka:
/rfc process 0x56bfa1b49e69991617d8aed0c880101f5506e0595b06a6b97b9d5a6cc014b264

rzadp:
/rfc process 0x56bfa1b49e69991617d8aed0c880101f5506e0595b06a6b97b9d5a6cc014b264

tomaka:
/rfc process 0x56bfa1b49e69991617d8aed0c880101f5506e0595b06a6b97b9d5a6cc014b264

rzadp:
> The on-chain referendum has approved the RFC, however I was not able to merge the PR automatically.

The bot seems to be missing permissions to merge the PR, although I cannot find out what exactly is missing and what needs to be changed. There were no changes to the bot since the [last time it merged a PR](https://github.com/polkadot-fellows/RFCs/pull/7#issuecomment-1717730755).

rzadp:
/rfc process 0x56bfa1b49e69991617d8aed0c880101f5506e0595b06a6b97b9d5a6cc014b264

rzadp:
/rfc process 0x56bfa1b49e69991617d8aed0c880101f5506e0595b06a6b97b9d5a6cc014b264

bkchr:
/rfc process 0x56bfa1b49e69991617d8aed0c880101f5506e0595b06a6b97b9d5a6cc014b264

bkchr:
/rfc process 0x56bfa1b49e69991617d8aed0c880101f5506e0595b06a6b97b9d5a6cc014b264

rzadp:
/rfc process 0x56bfa1b49e69991617d8aed0c880101f5506e0595b06a6b97b9d5a6cc014b264

anaelleltd:
@tomaka, do you have an update on the status of this RFC?
Thanks.

tomaka:
See https://github.com/paritytech/polkadot-sdk/issues/1825

PR #10, opened 2023-07-20
Burn Revenue from Coretime Sales

jonasW3F:
With this RFC I want to start a discussion whether to burn the revenues from Coretime sales or to send them to Treasury. I give some arguments why burning might be favorable.


lolmcshizz:
Readers of this RFC may be interested in the ongoing Polkadot forum discussion started by Jonas wrt providing a fixed flow to the Treasury from inflation - https://forum.polkadot.network/t/adjusting-the-current-inflation-model-to-sustain-treasury-inflow/3301.

The combination of the 2 ensures sustainable funding of the Treasury, whilst allowing coretime sales to be burned and offset some inflation.

text/0010-burn-coretime-revenue.md
```diff
@@ -0,0 +1,31 @@
...
+
+It's in the interest of the Polkadot community to have a consistent and predictable Treasury income, because volatility in the inflow can be damaging, especially in situations when it is insufficient. This raises the question: Why should we consider adding more volatility to the Treasury by tying it to fluctuating Coretime sales? Notably, Coretime revenue often appears to be inversely correlated with times when the Treasury should increase spending. During periods of low Coretime utilization (indicated by lower prices), it's important for the Treasury to spend more on projects and endeavours to increase the demand for Coretime. This pattern underscores that Coretime sales, by their very nature, are an inconsistent and unpredictable source of funding for the Treasury. Given the importance of maintaining a steady and predictable inflow, it's unnecessary to rely on another volatile mechanism. Some might argue that we could have both: a steady inflow (from inflation) and some added bonus from Coretime sales, but  burning the revenue would offer further benefits as described below.
```
lolmcshizz, line 25: Would it be worth adding the proposed change in Treasury income from inflation to this same RFC so the entire change in tokenomics can be discussed in one place & ensures that reviewers have the complete picture?

text/0010-burn-coretime-revenue.md
```diff
@@ -0,0 +1,31 @@
...
+
+- **Collective Value Accrual:** Following the previous argument, burning the revenue also generates some externality, because it reduces the overall issuance of DOT and thereby increases the value of each remaining token. In contrast to the aforementioned argument, this benefits all token holders collectively and equally. Therefore, I'd consider this as the preferrable option, because burns lets all token holders participate at Polkadot's success as Coretime usage increases.
```
monsieurbulb, line 31: Can you share any models or analysis that outline expectations or scenarios for Coretime income to be burned as this would be useful as it regards treasury spending more generally.

jakubgregus:
Burning of revenues thus redistributing them to holders should come much later when Treasury and Polkadot will be fully sustainable (without reliance on inflation like now) with proper treasury management in place (so diversified with stablecoins for many months,6-24?, of Treasury spending). In general, distribution of revenue is usually not coming before project/company/org have not reach market leader position which requires substantial spending for infrastructure, research ,marketing, sales,etc... 

Good example from non-crypto world is Amazon which is notoriously reinvesting all revenue into growth even after 25+ years of being public co. and still was able to achieve significant monetary premium due to their dominant position rather than being great dividend stock.

In crypto, burning usually works more as meme (very short term before its activation), rather than fundamental value driver, because revenues of crypto protocols are not so strong that they would be able to offset sudden market sell offs. While in (rare) case of Ethereum (and its soon to be remarkable 3% of burned supply) it can also become easily double edged sword in the form of too aggressive increase of costs for their L2s and their users which can price them out from those L2s to other ecosystems, because of chasing monetary premium for asset rather than being best tech. But this will also change by introduction of proto and full Dank sharding which will decrease price of L2s call data by factor of 10 (with proto Dank sharding) and later factor of 100 (with full Dank sharding) coming 4-8 months first and who knows when full version. So Ethereum community will also improve costs for L2s despite chasing "ultra sound money" narrative initially. 

Tldr; burn is either insignificant due to low traction or significant due to big traction or due to too high value extraction from low traction. Then we are getting to more important questions/topics which have to be answered and resolved sooner: 

1. what will be the pricing of coretime (which needs to be competitive enough with other alternatives to Polkadot, if not best)
2. what is overall ecosystem strategy for building and becoming best ecosystem, not just technology wise, but in other categories which will drive much higher consumption of coretime than current ecosystem and that traction itself will increase monetary premium which will increase ecosystem security, value of treasury itself,etc... 

In any case I would focus on 1. & 2. above, accrue revenues to Treasury for now and activate burn when Treasury and ecosystem will be in better shape.

SultanDot:
To guarantee the Treasury what if

1. Halve of Apy Staking from 19% to 10% 
2. to be given to the Treasury 4.5% to be sold as Coretime for henceforth burned income
3. To be given to the Treasury 4.5% to develop value at cost and as a sound cash reserve without compromising value


Benefits for all of us to get a balance of values in the Ecosystem so that it can grow rapidly
And for Hodler tokens we are happy to hold value without creating costly hedges for Ecosystem users.

csaint02:
> Burning of revenues thus redistributing them to holders should come much later when Treasury and Polkadot will be fully sustainable (without reliance on inflation like now) with proper treasury management in place (so diversified with stablecoins for many months,6-24?, of Treasury spending). In general, distribution of revenue is usually not coming before project/company/org have not reach market leader position which requires substantial spending for infrastructure, research ,marketing, sales,etc...
> 
> Good example from non-crypto world is Amazon which is notoriously reinvesting all revenue into growth even after 25+ years of being public co. and still was able to achieve significant monetary premium due to their dominant position rather than being great dividend stock.
> 
> In crypto, burning usually works more as meme (very short term before its activation), rather than fundamental value driver, because revenues of crypto protocols are not so strong that they would be able to offset sudden market sell offs. While in (rare) case of Ethereum (and its soon to be remarkable 3% of burned supply) it can also become easily double edged sword in the form of too aggressive increase of costs for their L2s and their users which can price them out from those L2s to other ecosystems, because of chasing monetary premium for asset rather than being best tech. But this will also change by introduction of proto and full Dank sharding which will decrease price of L2s call data by factor of 10 (with proto Dank sharding) and later factor of 100 (with full Dank sharding) coming 4-8 months first and who knows when full version. So Ethereum community will also improve costs for L2s despite chasing "ultra sound money" narrative initially.
> 
> Tldr; burn is either insignificant due to low traction or significant due to big traction or due to too high value extraction from low traction. Then we are getting to more important questions/topics which have to be answered and resolved sooner:
> 
> 1. what will be the pricing of coretime (which needs to be competitive enough with other alternatives to Polkadot, if not best)
> 2. what is overall ecosystem strategy for building and becoming best ecosystem, not just technology wise, but in other categories which will drive much higher consumption of coretime than current ecosystem and that traction itself will increase monetary premium which will increase ecosystem security, value of treasury itself,etc...
> 
> In any case I would focus on 1. & 2. above, accrue revenues to Treasury for now and activate burn when Treasury and ecosystem will be in better shape.

I think this is exactly the right way of thinking about this issue considering we are largely in the growth phase of Polkadot. So a static income to the treasury from inflation, rather than basing income on revenue, is the wrong direction imho. If Polkadot is to have a sustainable business model (apart from 99% of networks out there), then we should gauge the value based on our efforts and revenue rather than inflation.

Lower inflation instead, and then maybe adjust after revenues from blockspace sales are clearer. Another note, Amazon hasn't paid dividends to shareholders since they started AWS - burns are a soft analogy as the end goal is to increase value to the token imv.

Vcryptor:
Every DOT that is being payed for Core Time has to be burnt. Otherwise we will CREATE LEVERAGE. Because any DOT that was bought on the market to pay for Core Time will be used by Treasury to fund something, then will be eventually sold on the market, bought by other team to pay for another Core Time to refill the Treasury and so on. The Loop.  

Treasury has enough funding atm. If we believe that DOT will reach its ATH which it should do because it's vital for shared security model to have a very high mcap then Treasury will get more that enough funding.

If we don't burn revenue form Core Time sals every single DOT will be used mullion times and therefor devalued.

amenlo9:
my suggestion:

- Coretimes pricing in USDT term
- Example: USDT100 per 4 weeks (about 20 DOTs now)
- Appchain consume the coretime (burn the DOT) and treasury received USDT100
- Future Opengov treasury request shall be pay by USDT inside treasure
- this prevent treasury requester dump the DOT in open market

my 2c

text/0010-burn-coretime-revenue.md
jonasW3F: done in [a0a8345](https://github.com/polkadot-fellows/RFCs/pull/10/commits/a0a83453a7a87b2c1185f8952cfc18439ac5bde8). Nonetheless, my primary focus remains on the discourse surrounding the burn mechanism.

jonasW3F:
> 

Hi Jakub,

distribution of revenue also happens when you divert the DOT to Treasury. I argued that it is better to distribute it to every token holder than distributing them to a small set of actors through Treasury (that might be able to recoup some of their costs by ramping up their treasury proposals). We should not forget that those who pay for Coretime generally have additional ways to benefit from their Coretime usage, generating revenue from users. Therefore, making sure that Coretime purchases are 100% costs seems reasonable to me.

On a sidenote: personally, I don't think that funding Treasury through inflation is a bad thing, because it taxes all token holders to fund projects that otherwise go unfunded (due to the freeriding problem). Since it (supposedly) benefits all token, it is okay to "tax" all token holders by using diluting the value of all DOTs. Therefore, it is unnecessary to tax a specific part of the network (like Coretime buyers). So, I am not against the idea to see inflation funding Treasury (or stakers) in the future.

The points you are mentioning in 1) and 2) are very relevant but are both outside the scope of this RFC and can also be treated independently from where to funnel the revenues. For discussions about pricing, I think [here](https://github.com/polkadot-fellows/RFCs/pull/1) is a better place. It seems to me that right now the Coretime supply (especially after it becomes fungible and divisible) is much larger than it's demand. That'll lead to low prices in the beginning, reducing (perceived) barriers of entry and thereby attracting many innovators trying out their code. So, just like you, I expect burns to be rather insignificant at the beginning. Once there are sufficient revenues from Coretime sales, it makes sense to counterbalance inflation with it, though.

jonasW3F:
> 

I disagree with the idea to make any core protocol mechanism dependent on a stablecoin, especially those not issued by a core protocol mechanism. A price shock to DOT will eventually affect demand and the price (in DOT) for coretime should adjust accordingly.

jonasW3F:
> Every DOT that is being payed for Core Time has to be burnt. Otherwise we will CREATE LEVERAGE. Because any DOT that was bought on the market to pay for Core Time will be used by Treasury to fund something, then will be eventually sold on the market, bought by other team to pay for another Core Time to refill the Treasury and so on. The Loop.
> 
> Treasury has enough funding atm. If we believe that DOT will reach its ATH which it should do because it's vital for shared security model to have a very high mcap then Treasury will get more than enough funding.
> 
> If we don't burn revenue form Core Time sales, every single DOT will be used million times and therefor devalued.

This goes into the direction of my "clear incentives" argument.

text/0010-burn-coretime-revenue.md
jonasW3F: > 

this depends on the pricing mechanism which is, to my knowledge, not yet finalized.

csaint02:
What we are describing here is fungibility of the token and using it as a currency; one of the utilities of DOT.

Whether the DOT is burned or not does not matter in this context - the entity is still paying for blockspace with DOT making it a cost. It's going into the Polkadot account aka treasury. What business sells something and then burns the profits?

I'm also not following the distribution part of things. You are making a strong assumption that teams will recoup costs of coretime with treasury proposals - but in the end if it is benefitting the entire ecosystem; isn't that a net positive? We as a community need to engage in governance to prevent bad actors from ramping up proposals without having clear contributions to the entire eco - rather than encoding a mechanism that prevents a what if imho. 

If we want to balance, then why not reduce inflation? I would even think burning a less than 50% percentage of the DOT from coretime sales is feasible, but not the entire lot

LucaPolimec:
If we try to compare it to a business then Polkadot should be compared to a central bank, which has expenses and profits in the same currency which it issues itself - hence it's the closest analogy and don’t think that is the best proxy to compare it to but offers some broader understanding comparisons I think.

Stakeholders to be considered in the Polkadot ecosystem:
- Stakers
- Validators
- Tokenholders
- Parachains
- Treasury

Coretime sales prices/demand will be dependent (mid-/long term) of the system being secure and providing the functionalities required for a certain business model which requires enough stakers and validators (for providing the security) as well as tokenholders and parachains (to enhance value on product and user activity for a higher market cap for increased security).

Burning the revenue vs sending it to the treasury has different outcomes towards the various stakeholders and their incentives:

Burning revenue of coretime sales:
- Frequency and consistency of coretime sales demand are uncertain variables and highly market cycle driven over the next years (in relation to available funding for new projects)
- Burning revenue of coretime sales incentivizes every DOT holder pro rata -> same incentives for the stakers, validators, tokenholders and the parachain projects which hold DOT (same benefit for active/inactive holders)

Revenue of coretime sales towards treasury:
- As coretime sales demand/revenue fluctuate with market cycles it operates countercyclical to treasury funding needs
    - Treasury would require higher inflows during bear markets and vice versa
- On-chain treasury should be value enhancing overall, hence decisions taken should increase overall ecosystem value in the long run
- As DOT is still in its growth phase, higher uncertainty with funding decisions are accompanied so some decisions provide less value than others which is fine but overall it has to yield a net positive for the ecosystem to remain existent (as we would be better off otherwise)
- Treasury since inception until today cannot state that funded proposals equally benefit all stakeholders which provide the value of coretime sales forward (active/inactive users -> decisions driven by active holders in favor of parachain projects)

By burning the coretime sales the protocol will not become deflationary but the supply reduction will benefit all stakeholders pro rata and inflation rate decreases over time (transition from growth phase based on demand to lower inflation). 

Personally I don’t think the goal should be to become deflationary at all for a utility token as it penalizes active users that drive adoption who use the token compared to passive holders. Nevertheless, the more attractive use cases follow with increased user numbers the inflation can be lowered over time and the coretime sales offer a great opportunity to implement the first step with the right incentive alignment of the involved parties for it.

monsieurbulb:
> @LucaPolimec If we try to compare it to a business then Polkadot should be compared to a central bank, which has expenses and profits in the same currency which it issues itself 

Following this central bank analogy, we can make another step related to conversations around 'government spending (public services) and income (taxation) and think about the whole balance sheet. 

What assets should the sovereign state retain / acquire / trade? 

For example, there is spending, there is burning, but the customers creating that demand (parachains) are where value accrues... and their success is what will lever up demand for coretime. 

There is a disconnect between coretime demand drivers and a glut of supply. 

Fix this by using treasury to bootstrap parachains whose tokens are backed by uncorrelated (to KSM/DOT) assets. 

Build stake in these assets as part of a diversified governance balance sheet. 

Example, British government + East India Company / British Petroleum.

jonasW3F:
> If we try to compare it to a business then Polkadot should be compared to a central bank, which has expenses and profits in the same currency which it issues itself - hence it's the closest analogy and don’t think that is the best proxy to compare it to but offers some broader understanding comparisons I think.
> 
> Stakeholders to be considered in the Polkadot ecosystem:
> 
> * Stakers
> * Validators
> * Tokenholders
> * Parachains
> * Treasury
> 
> Coretime sales prices/demand will be dependent (mid-/long term) of the system being secure and providing the functionalities required for a certain business model which requires enough stakers and validators (for providing the security) as well as tokenholders and parachains (to enhance value on product and user activity for a higher market cap for increased security).
> 
> Burning the revenue vs sending it to the treasury has different outcomes towards the various stakeholders and their incentives:
> 
> Burning revenue of coretime sales:
> 
> * Frequency and consistency of coretime sales demand are uncertain variables and highly market cycle driven over the next years (in relation to available funding for new projects)
> * Burning revenue of coretime sales incentivizes every DOT holder pro rata -> same incentives for the stakers, validators, tokenholders and the parachain projects which hold DOT (same benefit for active/inactive holders)
> 
> Revenue of coretime sales towards treasury:
> 
> * As coretime sales demand/revenue fluctuate with market cycles it operates countercyclical to treasury funding needs
>   
>   * Treasury would require higher inflows during bear markets and vice versa
> * On-chain treasury should be value enhancing overall, hence decisions taken should increase overall ecosystem value in the long run
> * As DOT is still in its growth phase, higher uncertainty with funding decisions are accompanied so some decisions provide less value than others which is fine but overall it has to yield a net positive for the ecosystem to remain existent (as we would be better off otherwise)
> * Treasury since inception until today cannot state that funded proposals equally benefit all stakeholders which provide the value of coretime sales forward (active/inactive users -> decisions driven by active holders in favor of parachain projects)
> 
> By burning the coretime sales the protocol will not become deflationary but the supply reduction will benefit all stakeholders pro rata and inflation rate decreases over time (transition from growth phase based on demand to lower inflation).
> 
> Personally I don’t think the goal should be to become deflationary at all for a utility token as it penalizes active users that drive adoption who use the token compared to passive holders. Nevertheless, the more attractive use cases follow with increased user numbers the inflation can be lowered over time and the coretime sales offer a great opportunity to implement the first step with the right incentive alignment of the involved parties for it.

From what I understand, you agree with my arguments presented. You make a good point about deflation. I also don't advocate to make Polkadot deflationary, since I believe a utility token does not benefit from people hoarding it and not willing to use it. That being said, counter-balancing inflation and having all token holder participate at the success of coretime sales is a good thing. Once we come to a point where coretime sales exceed the inflation (it's unclear if it ever will, as coretime also scales over time which should not lead to skyrocketing prices), inflation should be adjustable via governance (as proposed in my [other proposal](https://forum.polkadot.network/t/adjusting-the-current-inflation-model-to-sustain-treasury-inflow/3301)).

seunlanlege:
Why not just let it go to treasury? Treasury already burns unspent funds. I'd argue for long term funding for ecosystem development and R&D. This ecosystem is too young to be mimicking already successful ecosystems that have spent hundreds of millions on ecosystem funding.

lolmcshizz:
> I'd argue for long term funding for ecosystem development and R&D.

That would be accomplished through the separate ongoing discussion [here](https://forum.polkadot.network/t/adjusting-the-current-inflation-model-to-sustain-treasury-inflow/3301). So the suggested solution is:

- Send a fixed portion of inflation to the Treasury to ensure predictable and sustainable funding for ecosystem development & R&D
- **Do not** add additional volatile income stream to the Treasury from Coretime revenue, instead burning it

So the resulting impact of the combination is:
- Consistent Treasury funding
- *Potential* for reduction in total inflation/ burn narrative/ deflation narrative

text/0010-burn-coretime-revenue.md
```diff
@@ -0,0 +1,33 @@
...
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 19.07.2023                                                                   |
+| **Description** | Revenue from Coretime sales should be burned                                                                   |
+| **Authors**     |                   Jonas Gehrlein                                                                          |
+
+## Summary
+
+The Polkadot UC will generate revenue from the sale of available Coretime. The question then arises: how should we handle these revenues? Broadly, there are three potential paths – burning the revenue and thereby removing it from total issuance or divert it to the Treasury. This Request for Comment (RFC) presents arguments favoring burning as the preferred mechanism for handling revenues from Coretime sales.
```
KiChjang, line 11: I can only see 2 paths discussed here: burning or put into treasury. What's the 3rd option?

jonasW3F: you are right, I'll fix it. Originally, I also thought about the possibility to divert revenues to validators. But after making the point about "Clear Incentives", this struck me as an option not even worth including.

filippoweb3:
I would be in favor of initially burning all the coretime sales. This would be the only way to categorize those sales as “pure network revenue”, as if they go to the treasury, they would ultimately be sold (and slightly burned). The token needs a more considerable deflationary pressure, and coretime sales could provide that, although I do not expect it to be substantial at the beginning. So, this might be another motive not to send them to the treasury. At some point, the net inflation could fall far below 10% due to the burning mechanism, and at that point, there could be a vote to divert some sales to the treasury. OpenGov has proven very efficient in enacting emergency proposals (see ideal staking rate). Lastly, but probably more important to me, even if the sales won't be initially substantial, burning them will have a tremendous social impact, making the DOT token more interesting and changing the inflationary narrative.

AlistairStewart:
I'm strongly in favour of this. We expect that coretime revenue will be small initially and may ramp up later. Treasury spending needs more certainty, just like staking. Governance should set the parameters of the inflation model as regards to spending. Lower inflation is great as long as development and security are covered. So it is better to give the possibility of low inflation later than larger treasury spending later.

filippoweb3:
> Why not KIIS and reduce the staking inflation rate to 3%. 3% encourages consumption but is not punitive to those who buy and hold coretime to cover their short term consumption needs.
> 
> Coretime sales can either be given to treasury or burned. I prefer the former.... Devs need funding. The opengov protocol has checks and balances to define how the treasury is spent.

With Ref 166 now enacted the ideal staking rate to 59% the treasury already gets 2-2.5x more from inflation compared to before. I do not see currently a problem in treasury funding. Revenues from coretime sales will be likely variable and not substantial at the beginning, and the treasury needs a stable source of income from inflation. The token inflation could stay at 10% so that it continues to reward network participants and fill up the treasury. On the other side, coretime sales will be burned and create a more important deflationary pressure.

text/0010-burn-coretime-revenue.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Stakeholders
+
+Polkadot DOT token holders.
+
+## Explanation
+
+This RFC discusses potential benefits of burning the revenue accrued from Coretime sales instead of diverting them to Treasury. Here are the following arguments for it.
+
+It's in the interest of the Polkadot community to have a consistent and predictable Treasury income, because volatility in the inflow can be damaging, especially in situations when it is insufficient. As such, this RFC operates under the presumption of a steady and sustainable Treasury income flow, which is crucial for the Polkadot community's stability. The assurance of a predictable Treasury income, as outlined in a prior discussion [here](https://forum.polkadot.network/t/adjusting-the-current-inflation-model-to-sustain-treasury-inflow/3301), or through other equally effective measures, serves as a baseline assumption for this argument.
```
xlc, line 25: this basically is saying we cannot approve this RFC without the inflation model adjustment RFC is accepted

text/0010-burn-coretime-revenue.md
xlc: that means we cannot evaluate/estimate the impact of this RFC and unable to make objective decision

text/0010-burn-coretime-revenue.md
jonasW3F: Well, not necessarily. Even without the change proposed in the forum post, I doubt that coretime revenues can become a reliable funding source for the Treasury in the short-/midterm. So, funding needs to come from other sources. Either we further depend on the staking inefficiency or we explicitly fix it with the proposed mechanism.

With that in mind, my argument is that Treasury funding needs to be fixed in some other dimension and then it's better to burn the revenue based on all the other arguments presented. But, this point is particularly strong once we fix Treasury inflow with the mechanism described in the forum post.

nhattu9xpro:
Burn $dot

text/0010-burn-coretime-revenue.md
burdges: Anyway they're both linked.  Alistair and Jonas have a combined argument here that burning here and inflation adjustment together make sense. 

At a high level, this sort of thing is almost universal in crypto currency.  Bitcoin shall eventually become insecure by not doing this sort of thing https://economics.princeton.edu/working-papers/on-the-instability-of-bitcoin-without-the-block-reward/#  

Really, the onus is upon "fees pay" people to argue that fees paying does not fuck up anything else, otherwise we should do this sort of thing as a default.  Of course we might need to link both this and the treasury change, just to simplify the governance process, not sure.

BlackDemonZyT:
> Every DOT that is being payed for Core Time has to be burnt. Otherwise we will CREATE LEVERAGE. Because any DOT that was bought on the market to pay for Core Time will be used by Treasury to fund something, then will be eventually sold on the market, bought by other team to pay for another Core Time to refill the Treasury and so on. The Loop.
> 
> Treasury has enough funding atm. If we believe that DOT will reach its ATH which it should do because it's vital for shared security model to have a very high mcap then Treasury will get more than enough funding.
> 
> If we don't burn revenue form Core Time sales, every single DOT will be used million times and therefor devalued.

I TOTALLY agree with this point, all coretime sales needs to be burnt.

BlackDemonZyT:
> I would be in favor of initially burning all the coretime sales. This would be the only way to categorize those sales as “pure network revenue”, as if they go to the treasury, they would ultimately be sold (and slightly burned). The token needs a more considerable deflationary pressure, and coretime sales could provide that, although I do not expect it to be substantial at the beginning. So, this might be another motive not to send them to the treasury. At some point, the net inflation could fall far below 10% due to the burning mechanism, and at that point, there could be a vote to divert some sales to the treasury. OpenGov has proven very efficient in enacting emergency proposals (see ideal staking rate). Lastly, but probably more important to me, even if the sales won't be initially substantial, burning them will have a tremendous positive social impact, making the DOT token more interesting and changing the inflationary narrative (despite continuing to be an inflationary token).

Totally agree with this, all coretime sales needs to be burnt, if DOT increases in price, the treasury will have MORE THAN ENOUGH funds.

PolkaDom:
I suppose I wonder why we wouldn't just parameterize this? If we know the two options we're between are burning it or giving it to the treasury, just create a percentage distribution parameter and open it up to gov.

ggwpez:
/rfc process 0xebcd94361ae57b58b67e47d77cf9b7346bdd615df0530829ff4105f7aeb10d9c

PR #11, opened 2023-07-20
Add New Account Creation Mechanism on Asset Hubs

joepetrowski:
Asset sufficiency only serves two purposes: fee payment and account existence. Using the Asset Conversion pallet allows fee payment to handle fee payment in any asset and has already been implemented. We can also use Asset Conversion to create accounts. This change will make non-sufficient assets _almost_ as featureful as sufficient ones in a secure way.


gavofyork:
As I mentioned in the call yesterday I don't think it's sensible to force users to have to exchange their currency (e.g. USDT/USDC) into the single native token merely to store the asset on asset hub.

joepetrowski:
> As I mentioned in the call yesterday I don't think it's sensible to force users to have to exchange their currency (e.g. USDT/USDC) into the single native token merely to store the asset on asset hub.

Yeah I will change this to propose this as an option, but keep sufficiency.

joepetrowski:
> I am a bit torn about what I am to say because I hate making things more complex but this warrants discussion: I am not sure if we should change the default transfer function or expose this through a second transferMaybeCreate function (with perhaps a maxDeposit parameter that would make the tx fail if the deposit would become too costly). It will be impossible to know how much the fee in the token actually will be and this could vary greatly between the transaction being approved by the user and it being included in a block; a large LP could remove liquidity or a lot of other transactions could happen first. I would rather be able to explicitly allow for this than having only one transfer function.

Yes, I included the `transfer` logic as an example to express the logic in a way that was more clear, but not as implementation instructions. For sure it could be a new dispatchable that allows the sender to put a limit on how much of the asset is consumed to create the account.

text/RFC-0011-Add New Account Creation Mechanism on Asset Hubs.md
```diff
@@ -0,0 +1,167 @@
...
+- Wallet and UI/UX developers
+
+## Explanation
+
+By using the Asset Conversion protocol, the system can convert any asset to DOT as long as there is
+a path from that asset to DOT. As such, we can rely on the economic security provided by the
+existential deposit of DOT by simply converting some amount of the asset being transferred to the
+existential deposit.
+
+This conversion only need happen when the account does not yet exist. When the destination account
```
olanod, line 58:
```suggestion
This conversion only needs to happen when the account does not yet exist. When the destination account
```
As an illiterate non-native speaker I might be wrong 😅

text/RFC-0011-Add New Account Creation Mechanism on Asset Hubs.md
```diff
@@ -0,0 +1,167 @@
...
+        // The swap returns the amount of the asset consumed to acquire the ED.
+        let consumed = Swap::swap_tokens_for_exact_tokens(
+            from,                // sender
+            vec![asset, dot],    // path, where `dot` is Multilocation {parents: 1, interior: Here}
+            existential_deposit, // amount_out, we need the ED for the account
+            destination,         // send_to
+            ..
+        )?;
+        // We used some asset for the swap, so we have to subtract that from the amount.
+        let remaining_asset_amount = amount.saturating_sub(consumed);
```
olanod, line 111: Could it be additive instead of subtractive? I think if the sender agrees on `create_destination` it understands better the mechanics of the transfer and might be willing to cover the ED, the recipient on the other hand might be a new user how doesn't understand why he's getting 99.99 instead of the 100 he agreed to with the sender.

joepetrowski: Yeah, could be an option. That's getting more into the implementation detail. I think it gets to the core question of this RFC: Should AH directly add in a path to create accounts via Asset Conversion, or rely on people doing something like batching multiple calls together to handle the destination's existence? IMO, yes, with configuration options like "sender pays, receiver pays, fail if nonexistent".

text/0011-add-new-path-to-account-creation-on-asset-hubs.md
```diff
@@ -0,0 +1,167 @@
...
+This proposal would benefit the ergonomics of the system for end users by allowing all assets to
+create destination accounts when needed.
+
+### Compatibility
+
+This change would require changes to the Assets pallet to add the new account creation path.
+
+## Prior Art and References
+
+Discussions with:
```
xlc, line 156: Acala/Karura have the feature to allow assets to pay for tx fee with help of swap and ensure there is ED amount of the native token for sufficiency
https://github.com/AcalaNetwork/Acala/blob/master/modules/transaction-payment/src/lib.rs

text/0011-add-new-path-to-account-creation-on-asset-hubs.md
```diff
@@ -0,0 +1,167 @@
...
+
+- SR Labs auditors, in particular Jakob Lell and Louis Merlin
+- The monthly Asset Conversion ecosystem call, particular inspiration from Jakub Gregus
```
jak-pan, line 159: *Jakub Panik ✋🏻

We're doing something very similar in HydraDX but with some added complexity which you might not need. (Default "best" payment routes, not swapping directly just sending to treasury if we accept the currency, to save tx cost of swap) 

I have few small notes for now from the discussion - It would be great to have something like transfer + cover fees, because now we need to know upfront
1. sufficiency of asset on receiving chain - if insufficient handle this
2. the logic the chain has implemented for this sufficiency (this might become a non-issue later if there is some consensus on this)

Currently to be able to transfer i.e. USDC to asset hub we need to send dot first to pay the fee / ed. If you have the pool for USDC it's fine after this PR but if not, it is a problem and we are introducing branching logic to just sending assets x-chain. The transfer+pay could be possibly implemented by sending the fee payment asset together with the second asset (as we do right now) then possibly sent back to sender if not needed. Or possibly taken straight from the payer's account on the receiving chain. Just speculating here.

P.S. you probably don't want to create "receiver pays" since you could drain tokens from receiving account by sending random tokens to it, which is more griefing than chain attack, but still not great.

text/RFC-0011-Add New Account Creation Mechanism on Asset Hubs.md
jak-pan: Yeah, this is now handled on the frontend by "guesstimating" the destination fee and adding it to the amount "sent"

joepetrowski:
Going to close this as it's stale, and I think that it's largely solvable on the front end. Wallets can offer to batch a swap in front of a transfer if they see that the destination account doesn't exist.

acatangiu:
With `asset-conversion` being available on Asset Hub, and the DOT existential deposit being lowered to 0.01 DOT (below 1$ even if DOT price goes up to 99$), I propose we reopen this proposal and make it an automated mechanism on Asset Hub, removing all complexity from source chains or front ends.

On receipt on AssetHub of any asset amount that doesn't satisfy destination account ED, up to 0.01DOT is swapped from the transferred assets to satisfy the destination account ED, and not fail the transfer.

PR #12, opened 2023-07-24
Process for Adding New System Collectives

joepetrowski:
This RFC proposes a means for Polkadot stakeholders to ratify a new collective. The Fellowship should agree to add new collectives to the Collectives parachain runtime that pass the given process.


x676f64:
ChaosDAO would also like to explore this space.

text/RFC-0012-Process-for-Introducing-New-Collectives.md
```diff
@@ -0,0 +1,108 @@
...
+  can whitelist calls for referenda via the `WhitelistOrigin`.
+
+This information could all be in a single document or, for example, a GitHub repository.
+
+After publication, members should seek feedback from the community and Technical Fellowship, and
+make any revisions needed. When the collective believes the proposal is ready, they should bring a
+remark with the text `APPROVE_COLLECTIVE("{collective name}, {commitment}")` to a Root origin
+referendum. The proposer should provide instructions for generating `commitment`. The passing of
+this referendum would be unequivocal direction to the Fellowship that this collective should be
+part of the Polkadot runtime.
```
muharem, line 62: Who is expected to work on introducing an approved collective to the runtime/code base?
Should we clarify it in this RFC?

joepetrowski: Not exact but similar to: https://github.com/polkadot-fellows/RFCs/pull/2#discussion_r1254675263 

If the Fellowship is being paid by the network and they agree to a process to signal network desire for some code in a runtime that is within the Fellowship GitHub org, then someone in the Fellowship should take responsibility to make sure it gets done. The PR will need approval from high-ranking Fellows even if someone outside the Fellowship implements it.

olanod:
Looks *Ok* to me :) in general I think the approach of creating new instances of the different pallets that make up a collective like the fellowship is not very approachable nor scalable but I think this works as a start, at least the community will know there's a processes. To make the process easier/automatable for the fellowship I would be more strict about the format used for the submission and include some form of machine readable format that tools can use to auto generate the instantiation code and auto-submit a PR for the new collective. Markdown for the manifesto annotated with a "front matter" section has a good balance of being simple enough for the proposers(even if they don't use a tool) and still machine parsable.  

```md
---
collective: Great Collective
type: ranked
member_types: individuals
has_treasury: yes
ranks:
- name: simpletons  # used to generate origin and track
  track:  # overrides for the associated track default config
    decision_period: 5 days
- name: smarty-pants
  salary: 1 DOT
seeding:
- rank: 1
  account: 1universe1ruLer11111111111111111111111111111G91
---

# The Great Collective Manifesto
...
``` 

Later on as demand for collectives increases the fellowship might find worth it investing some dev hours creating a permission-less collectives system where new collectives are incubated continuously and every now and then one such collective might apply to become a "system collective" like the technical fellowship.

joepetrowski:
> Later on as demand for collectives increases the fellowship might find worth it investing some dev hours creating a permission-less collectives system where new collectives are incubated continuously

As a new feature/chain this could be valuable, but this is not at all the purpose of system collectives. System collectives _are_ meant to be permissioned; namely they need buy-in ("permission") from the network to be considered part of the Polkadot system, and as such need to justify their reason for existing and the powers they wish to acquire and get explicit approval from stakeholders.

The process is not meant to be _scalable_ (although adding/configuring a few pallets is hardly a huge lift), it is meant to ensure that people have an accepted process to demonstrate that buy-in and a reasonable expectation that this group (the Fellowship) will honor that signal.

ggwpez (approved):
LGTM, thanks!  
We can iterate on it once the first collectives come in and provide feedback.

liamaharon (approved):
LGTM. 

Regarding the suggestion for machine readable format: I think it's too early to know whether this would valuable or how it should be structured. I'm favour of proceeding without it for now, and adding it only if/when a clear need arises.

ggwpez:
/rfc process 0x4b54385ffd3ef633aa021d7c56d06de833071380f4ced0e747705dce1838f622

PR #13, opened 2023-07-24
Prepare `Core` runtime API for MBMs

ggwpez:
This RFC aims to ratify the proposed changes to the `Core` runtime API. 

Implementation is being conducted in https://github.com/paritytech/polkadot-sdk/pull/1781.  
Status: RFC has been slimmed down and the controversial change to `BlockBuilder` removed by finding an alternative.


text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
```diff
@@ -0,0 +1,120 @@
...
+The original motivation is Multi-Block-Migrations. They require the runtime to be able to tell the
+block builder that it should not attempt to include transactions in the current block. Currently,
+there is no communication possible between runtime and block builder that could achieve this.
+Further, it is necessary to execute logic right after inherent application but still before
+transaction inclusion.
+
+This RFC proposes a way of communication between the runtime and the block builder by changing the
+`initialize_block` function to return a `BlockExecutiveMode` enum. Additionally, an
+`after_inherents` function is introduced that runs after inherent but before transaction
+application.
```
tomaka, line 27: Random remark: why not make `after_inherents` return the `BlockExecutiveMode`, instead of `initialize_block`? Then it would just indicate to the client whether the block authorship should continue or stop now.

ggwpez: Thanks for the comment, originally it was implemented like this.  
But then came up the idea of optional inherents: https://github.com/paritytech/substrate/pull/14414#issuecomment-1606262686, and with that the need to already communicate this `Mode` before the inherents get applied. In general this allows for more freedom already before generating inherents. 

Does it sounds good to you?

tomaka: > then came up the idea of optional inherents

The comment you linked is Basti asking what optional inherents are, because as far as I understand there's no such thing as an optional inherent at the moment.

I'm not very opinionated, it just seems to me that the design is a tiny bit weirder than it could be.

ggwpez: Yes the runtime could alter the return value of `inherent_extrinsics` :thinking:   

There is still this one [comment from Gav](https://github.com/paritytech/substrate/pull/14414#issuecomment-1602222428) which mentions to already having two modes prior to inherent application.

kianenigma (commented):
To the best of my knowledge, the property of "Inherents must come first" is a FRAME concept, and in itself FRAME can rather easily be tweaked to support inherents that might come later in the block.

It seems like this RFC is baking this assumption into Substrate's client side abstractions as well. 

Can you elaborate further on the state of the above statement? Is it correct? Can substrate ever support inherents that will not come first?

tomaka:
The way the client authors a block is:

- Call `Core_initialize`.
- Call `BlockBuilder_inherent_extrinsics`, passing as parameter a struct containing for example the timestamp and some parachain head I'm not familiar with. The runtime returns a list of extrinsics that are not yet applied.
- Call `BlockBuilder_apply_extrinsic` once per extrinsic returned by `BlockBuilder_inherent_extrinsics`.
- (new with this RFC) Call `BlockBuilder_after_inherents`.
- Call `BlockBuilder_apply_extrinsic` once per extrinsic pulled from the transaction pool, until a certain time has passed.
- Call `BlockBuilder_finalize_block`.

The Substrate client does not support inherents that don't come first.

As I've mentioned above, the runtime is in control of the list of inherents that will be applied, as it decides of the list returned by `BlockBuilder_inherent_extrinsics`.

kianenigma:
> The Substrate client does not support inherents that don't come first.

You are mentioning the new procedure, and of course, in the new format the Substrate client only accepts inherents that come first. 

But this is a new assumption, as per this RFC, and I am asking why that is the case.

JoshOrndorff:
I also want to chime in about the idea of inherents always being first. I believe @tomaka's point is that the block builder has always put inherents at the beginning and this is not new. He is correct about that. I believe @kianenigma's point is that this was never specified as the the only way to do it, but rather just the way that the block builder happened to be implemented.

I share @kianenigma's concern that this removes flexibility, and I have a concrete use case. [Tuxedo](https://github.com/Off-Narrative-Labs/Tuxedo/) is a runtime framework for building utxo-based chains (it is vaguely analogous to frame).

Constraints:
1. In the UTXO model there is one very important invariant which is that every piece of state (utxo) is created explicitly as the output of some transaction.
2. It is also common for the difference in input and output values in the transactions to go to the block author as a reward.

In order for the block author's reward to be explicitly created by a transaction as required by #1, we need an inherent that creates the reward. But in order for it to reflect the correct tip amount which is only known after the other transactions are applied, the inherent must go at the end of the block.

My Request: **Please don't bake the assumption that inherents must be first into the Substrate protocol** level (I don't care if you do it in FRAME though).

ggwpez:
> My Request: **Please don't bake the assumption that inherents must be first into the Substrate protocol** level (I don't care if you do it in FRAME though).

Yes I see... The proposed design is strict with its assumptions. I will try the alternative approach with the transaction ordering that is mentioned in the "Future Directions" section.

kianenigma:
> This is not a new assumption. The block building procedure that the Substrate client follows always puts inherents first and has always done so.

The default substrate `block_builder` is also an _implementation_ and not the _specification_. 

As it stands now, but sides of a contract (`block_builder` and FRAME) adhere to this rule, but the contract in between them does not enforce this property. 

I don't necessarily want to disagree with this new API, but I want us to be clear about what we are doing: 

- Up until now, it would have been possible to make inherents not come first with some tweak to block builder and FRAME. 
- Going forward, this becomes increasingly harder, because it is now more baked into the contract between the client and runtime.

tomaka:
> The default substrate block_builder is also an implementation and not the specification.

The fact that inherents always come first is very much the specification. It is how it's designed. It's part of the protocol between the runtime and the client.

If you want inherents to not come first, you have to design a new feature. You have to modify the specification.

I can understand the argument that you want to leave the room later for the possibility to modify the specification so that inherents don't come first, and it is a valid argument against this RFC, but it is wrong to say that inherents do not necessarily come first now, because they do.

pepyakin:
I think the concept of inherents is limited. Right now we treat inherents as transactions, i.e. each inherent triggers a state transition, but I would argue that they are not transactions, or rather treating them as such is limiting. What they actually are trying to achieve is passing some data from the block builder into the runtime, and as we learned the hard way it's very useful to separate the data submission and data handling. The main motiviation to create the original issue (https://github.com/paritytech/substrate/issues/9210) was that some of that data is not available in `on_initialize`. As the solution to this we thought about adding `on_after_inherents`, which kind of solves some issues, but doesn't solve the others. There was an [idea](https://github.com/paritytech/substrate/issues/9210#issuecomment-875427736) to submit all the inherent data, we would not check/process the data and instead just stash that data somewhere, and finally in `on_after_inherents` we would finally unstash it and process.

But, the question is, why don't we treat inherents as what it is: just some data that block producer included in the block body? 

Here is a straw-man proposal. Imagine there is a special place in a block that is dedicated for inherent data. This data is represented as a (btree)map between the inherent ID and the inherent data. The runtime interface gets a function `get_inherent_data(inherent_id) → Option<Vec<u8>>`. It's behavior is defined as follows:

1. During normal block execution, this function should be able to read the inherents from the block directly. 
2. During candidate validation (in polkadot PVF), cumulus would read this data from the block similar to the normal block execution.
3. During block building, upon the first call this function would look up the `sp_inherents::InherentDataProvider` for the specified ID and get the corresponding data. If it's not the first call, then the data generated on the first call is returned.

Now, what does this construction give us?

Note that the generation of inherent data now becomes lazy if needed but it doesn't have to be that way and the inherent data could be fixed before the block creation and just returned as is in from the `InherentDataProvider`. Because the data is not fixed, this allow us to provide the inherent data at the end of the block. 

Another consequence, is that there is no such concept as a mandatory/optional inherents. If the inherent data returned is `None` but the runtime is not content with that, it can immediatelly signal that by e.g. panicking.

If `sp_inherents::InherentDataProvider` was changed in a way that it can inspect the storage of the system, then it would make it possible to support Tuxedo (as described by @JoshOrndorff [here](https://github.com/polkadot-fellows/RFCs/pull/13#issuecomment-1652127220)). Specifically, the inherent data provider would access the UTXO module that would provide the difference. Although, an alternative implementation is possible where tips are tallied up through events, this is just a demonstration of flexibility of this approach.

We could extend this proposal even more. The straw-man proposal above uses the static `InherentId` just because it is not too disimilar to what we have right now. However, it's easy to imagine how this mechanism could benefit from dynamic requests. To give you motivation, right now when we produce a parachain block, the collator supplies data from the relay-chain as inherent. Specifically, that data includes some storage proofs from the relay-chain required for channels to work. This is now implemented by tightly coupling the block production with the parachain-system pallet. I.e. block production knows what storage parachain-system will want to read and reads all the required data from the relay-chain along with the merkle proofs. But turns out people actually want to read the data from the relay-chain or potentially any other parachainchain (e.g. https://github.com/paritytech/cumulus/issues/320). So if we had two functions provided as inherents: 1. read storage from another chain 2. provide the multiproof of the read storage then we could support this use-case.

I must note however that the tight coupling between inherents checking and block building is desired for efficiency. In the above example, when we fetch the relay-chain storage proofs it won't be too good of an idea to wait until the runtime requests a key that we already know will be requested so we could as well request it in advance. However, note that this inherent mechanism doesn't preclude that: we could still initiate the requests for the storage items we know will be read in advance and then block in the `InherentDataProvider` only if the data didn't come in time.

Paging in @xlc @bkchr

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
```diff
@@ -0,0 +1,127 @@
...
+
+The original motivation is Multi-Block-Migrations. They require the runtime to be able to tell the
+block builder that it should not attempt to include transactions in the current block. Currently,
+there is no communication possible between runtime and block builder that could achieve this.
+Further, it is necessary to execute runtime logic right after inherent application but still before
+transaction inclusion.
+
+This RFC proposes a way of communication between the runtime and the block builder by changing the
+`initialize_block` function to return a `ExtrinsicInclusionMode` enum. Additionally, an
+`after_inherents` function is introduced that runs after inherent but before transaction
```
rphmeier, line 26: This sentence is redundant

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
```diff
@@ -0,0 +1,127 @@
...
+Backwards compatibility with the current runtime API SHOULD be implemented on the node side to not
+mandate a lockstep update.
+
+## Drawbacks
+
+[…] drawbacks relating to performance, ergonomics, user experience, security, or privacy: None
+
+The only drawback is that the block execution logic becomes more complicated. There is more room for
+error.  
+Downstream developers will also need to adapt their code to this breaking change.
```
rphmeier, line 62: Do we have a good definition of breaking changes? This seems to be opt-in and doesn't break node-side code unless the user's runtime is updated.

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
```diff
@@ -0,0 +1,127 @@
...
+- Polkadot Parachain Teams: They also have to adapt to the breaking changes but then eventually have
+  multi-block migrations available.
+
+## Explanation
+
+The only relevant change is on the node side in the block authoring logic. Any further preparation
+for MBMs can happen entirely on the runtime side with the provided primitives.
+
+All block authors MUST respect the `ExtrinsicInclusionMode` that is returned by `initialize_block`.
+They MUST always invoke `after_inherents` directly after inherent application. The runtime MAY
```
rphmeier, line 43: what parameters does `after_inherents` take? Is there any way to limit the amount of weight this can spend?

The use-case I'm thinking of is consensus algorithms where the amount of time which can be spent in authoring is variable, e.g. asynchronous backing scenarios where some blocks might be authored in a full 2 seconds and others only in 1 second.

Currently, block builders have the ability to gauge how heavy transactions are, and to stop authoring after submitting enough transactions.

Block builders should be able to limit the amount of time spent in `after_inherents`, although probably not arbitrarily.

Perhaps the `ExtrinsicInclusionMode` should return a minimum amount of weight that must be allotted to the rest of the block, so block builders can decide whether to simply give up authoring or continue.

rphmeier: To some extent I'm wondering if it makes sense to flip block-building inside out - in theory we could have a single `fn build_block(InherentData, some_total_weight_or_time_limit)` and supply new transactions through special host functions instead, so the runtime actually decides whether to pull transactions from the pool, etc.

bkchr: Weight is a runtime concept. How much weight you want to spent is also something that the runtime should decide and not the node. All information that you want to pass between initialize and after inherents, can be stored temporarily in the state and doesn't need to be passed to the function.

Regarding driving the block production from inside the runtime. I think there exist an issue or we at least discussed it at some point. The main problem being a panic in a transaction. We can not unwind the panic. We would may need to spawn a new was instance to ensure block production isn't affected by a panic.

rphmeier: > How much weight you want to spent is also something that the runtime should decide and not the node.

I disagree - there's no good way for the runtime to know how much time is available for authoring in the general case.

ggwpez:
Retracted for being too rigid with the `after_inherents` hook. 

I will factor out the `initialize_block` changes to another RFC since I think that is enough to get MBMs going. Downside is that we wont be able to use it for the `poll` hook.

ggwpez:
Reopening after further discussion at the Fellowship meeting today.  
Outcomes:
- This approach is a good intermediate solution so far since MBM are very much needed.
- In the future we want to switch to runtime-driven block building and get away from hard-coded logic. This would make the `BlockBuilder` trait obsolete anyway.
- The flaw here remains the inflexibility but it is being traded off for earlier MBMs.
- Note: The `BlockBuilder` is not required to be used, a block author can author the block however he wants as long as the block is valid. No direct use to use `BlockBuilder`.

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
```diff
@@ -0,0 +1,126 @@
...
+  OnlyInherents,
+}
+```
+
+A block author MUST respect the `ExtrinsicInclusionMode` that is returned by `initialize_block`. The runtime MAY reject blocks that violate this requirement. 
+
+It is RECOMMENDED that block authors keep transactions in their transaction pool (if applicable)
+for as long as `initialize_block` returns `OnlyInherents`. The assumption is that these transactions become valid once the runtime finishes the MBM.  
+Backwards compatibility with the current runtime API SHOULD be implemented by block authors to not mandate a lockstep update of the authoring software.  
+This could be achieved by checking the runtime API version and assuming that `initialize_block` does not have a return value when the version is lower than 7.
```
bkchr, line 54: Not sure we need this paragraph.

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
```diff
@@ -0,0 +1,126 @@
...
+```rust
+enum ExtrinsicInclusionMode {
+  /// All extrinsics are allowed in this block.
+  AllExtrinsics,
+  /// Only inherents are allowed in this block.
+  OnlyInherents,
+}
+```
+
+A block author MUST respect the `ExtrinsicInclusionMode` that is returned by `initialize_block`. The runtime MAY reject blocks that violate this requirement.
```
bkchr, line 49:
```suggestion
A block author SHOULD respect the `ExtrinsicInclusionMode` that is returned by `initialize_block`. The runtime will reject any non-inherent transaction, meaning the block author would only waste computation time. 
```

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
```diff
@@ -0,0 +1,126 @@
...
+
+### `BlockBuilder::last_inherent`
+
+A block author MUST always invoke `last_inherent` directly after applying all runtime-provided inherents. The runtime MAY reject blocks that violate this requirement.
+
+### Combined
+
+Coming back to the three main features and how they can be implemented with these runtime APIs changes:
+
+**1. Multi-Block-Migrations**: The runtime is being put into lock-down mode for the duration of the migration process by returning `OnlyInherents` from `initialize_block`. This ensures that no user provided transaction can interfere with the migration process. It is absolutely necessary to ensure this, since otherwise a transaction could call into un-migrated storage and violate storage invariants. The entry-point for the MBM logic is `last_inherent`. This is a good spot, because any data that is touched in inherents, is not MBM-migratable anyway. It could also be done before all other inherents or at the end of the block in `finalize_block`, but there is no downside from doing it in `last_inherent` and the other two features are in favour of this.
```
bkchr, line 64: > This is a good spot, because any data that is touched in inherents, is not MBM-migratable anyway

This is not entirely true. Things like the parachain consensus system could still be migrated by a MBM. This would require that the `provide_inherent` call returns an empty "inherent", aka doesn't attach any data to it.

bkchr (approved):
Looks good to me

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
pgherveou, line 64: > The runtime is being put into lock-down mode for the duration of the migration process

So a pallet with a slow multi-block migration, will block all other pallets until it's fully migrated. I guess that is the only way we can safely put all transactions from the pool on pause until the mbm and lock-down mode is over.

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
tomaka: IMO "MUST" is better.
The runtime isn't meant to try to detect bugs in the host, because it can't anyway. It can only assume that the host behaves as expected. In that specific situation, the runtime is able to detect bugs in the host, but I don't see why we should create an exception here just because it's possible.
I'd even suggest that the runtime panics if the host ignores the `ExtrinsicInclusionMode`.

bkchr: > I'd even suggest that the runtime panics if the host ignores the `ExtrinsicInclusionMode`.

Yeah 100%. This is what "will reject" means.

ggwpez: I was using `MAY` since the inherent Extrinsic could have been a storage No-OP, in which case the runtime would have no way of detecting it.  
But it is probably easier to use `MUST` here.

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
ggwpez: I dont quite get what you mean. What i wanted to express is, that the logic of an inherent could try to decode storage that should currently not be accessed since an MBM is ongoing.

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
ggwpez: Okay, less text = less bugs 😄

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
ggwpez: `after_inherents` was now gladly removed.

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
ggwpez: Sentence was removed.

ggwpez:
The RFC has been slimmed down and the controversial change to `BlockBuilder` removed by finding an alternative.

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
```diff
@@ -0,0 +1,113 @@
...
+```rust
+enum ExtrinsicInclusionMode {
+  /// All extrinsics are allowed in this block.
+  AllExtrinsics,
+  /// Only inherents are allowed in this block.
+  OnlyInherents,
+}
+```
+
+A block author MUST respect the `ExtrinsicInclusionMode` that is returned by `initialize_block`. The runtime MUST reject blocks that have extrinsics in them while `OnlyInherents` was returned.
```
kianenigma, line 50:
```suggestion
A block author MUST respect the `ExtrinsicInclusionMode` that is returned by `initialize_block`. The runtime MUST reject blocks that have non-inherent extrinsics in them while `OnlyInherents` was returned.
```

nit, or else we have got the definition wrong here: https://paritytech.github.io/polkadot-sdk/master/sp_runtime/generic/struct.UncheckedExtrinsic.html

Although these namings probably need revision based on https://github.com/paritytech/polkadot-sdk/pull/2280 as well

kianenigma (approved):

text/0013-prepare-blockbuilder-and-core-runtime-apis-for-mbms.md
ggwpez: Oh thanks, yea this was def wrong before.

ggwpez:
/rfc process 0xbb5a58e7aef7153e78d37d7248c7114739e0da19040356045d1b56de86813cc2

PR #14, opened 2023-07-25
Improve locking mechanism for parachains

xlc:
If this is accepted, I plan to implement this to close https://github.com/paritytech/polkadot/issues/7539

Draft PR for Polkadot https://github.com/paritytech/polkadot/pull/7560

### DIFF
--- a/text/0014-improve-locking-mechanism-for-parachains.md
+++ b/text/0014-improve-locking-mechanism-for-parachains.md
@@ -8,11 +8,15 @@
 
 ## Summary
 
-The manager of a parachain has permission to manage the parachain when the parachain is unlocked. Parachains are by default locked when onboarded to a slot. This requires the parachain wasm/genesis must be valid, otherwise a root track governance action on relaychain is required to update the parachain.
+This RFC proposes a set of changes to the parachain lock mechanism. The goal is to allow a parachain manager to self-service the parachain without root track governance action.
 
-This RFC proposes a mechanism to allow a parachain manager to self-service the parachain without root track governance action.
+This is achieved by remove existing lock conditions and only lock a parachain when:
+- A parachain manager explicitly lock the parachain
+- OR a parachain block is produced successfully
 
 ## Motivation
+
+The manager of a parachain has permission to manage the parachain when the parachain is unlocked. Parachains are by default locked when onboarded to a slot. This requires the parachain wasm/genesis must be valid, otherwise a root track governance action on relaychain is required to update the parachain.
 
 The current reliance on root track governance actions for managing parachains can be time-consuming and burdensome. This RFC aims to address this technical difficulty by allowing parachain managers to take self-service actions, rather than relying on general public voting.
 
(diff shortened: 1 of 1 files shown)


text/0014-improve-locking-mechanism-for-parachains.md
```diff
@@ -0,0 +1,107 @@
...
+A parachain is automatically locked once it is onboarded on a slot[^4].
+
+Only the relaychain Root origin or the parachain itself can unlock the lock[^5].
+
+This creates an issue that if the parachain is unable to produce block, the parachain manager is unable to do anything and have to rely on relaychain Root origin to manage the parachain.
+
+This RFC proposes to automatically ignore the lock status of the parachain with either of the conditions:
+
+1. The wasm and genesis state are both empty.
+2. The parachain is considered unable to perform block using the following hueuristic:
```
shawntabrizi, line 48:
```suggestion
2. The parachain is considered unable to create a block using the following hueuristic:
```

shawntabrizi:
This makes sense to me. Basically, the idea is:

- lots of chains have shown to have problems at genesis
- we want to make it as easy as possible for chains to fix themselves at the genesis
- we want chains to be locked by default, with the explicit choice to unlock themselves

So the solution is: detect when a chain is broken at genesis, allow the chain to fix itself, lock it as soon as it is shown to be "alive".

This obviously will not work for every possible scenario which parachain teams may encounter. For example a chain which set the wrong sudo key or multi-sig and stuff like that, which can produce blocks, but then does not have access to their governance.

The question here is then:

Shouldn't we just lock a parachain on the production of block 1? I don't see an advantage to a chain getting locked, then getting unlocked, then getting locked again.

Chains with some malicious intention could always fake being broken at genesis to then unlock their chain, so might as well assume every chain is unlocked until it produces it first block.

This should overall simplify the logic of this locking system, since there is a single definite point where we add the lock, and no need to detect for a pause or anything else to trigger an unlock.

xlc:
> Shouldn't we just lock a parachain on the production of block 1? I don't see an advantage to a chain getting locked, then getting unlocked, then getting locked again.

The chain could be locked before it is onboarded. For example, create a crowdloan will apply the lock. The para manager can also voluntarily lock the parachain. My proposal will allow rescue parachain for such cases.

shawntabrizi:
We lock a parachain during a crowdloan to prevent the code from significantly changing from what the users thought they were contributing to.

However, it seems like:

- historically, lots of chains have uploaded a minimal runtime which would get later upgraded to something totally different anyway
- teams could simply "not make a block" for a long time to bypass the lock anyway
- crowdloans are going to be deprecated in favor of the new core assignment systems

As for para managers voluntarily locking their parachain, seems like they might only want to do it after the first block anyway.

xlc:
I agree crowdloan lock introduces more trouble than assurances.

One more question about this approach: How to handle existing parachains? With my suggestion, parachains that never produced block will automatically be unlocked. With the new approach, they won't, and should we have a migration for them?

arrudagates:
I think it's fair to do a migration and unlock every ParaId that never
produced block #1. Would there be any drawbacks to this?

xlc:
I can't think much reasons why we shouldn't do that. But I also want to see a few more supports before making the adjustments.

shawntabrizi:
@xlc i think unlocking them seems fine. If we can contact the teams, perhaps we can simply ask them what they would like us to do. Wouldn't be too crazy to make a hand-coded migration for something like this.

xlc:
It is only the parachains with lease but never produced blocks are qualified, so I will image it won't be an easy task to contact them.
On the other hand, unlock parachain wouldn't cause any issues for them. They can always relock it. So this could be fine as long as we communicated enough to all the teams and the communities about this change.

text/0014-improve-locking-mechanism-for-parachains.md
```diff
@@ -0,0 +1,105 @@
...
+| --------------- | ---------------------------------------- |
+| **Start Date**  | July 25, 2023                            |
+| **Description** | Improve locking mechanism for parachains |
+| **Authors**     | Bryan Chen                               |
+
+## Summary
+
+The manager of a parachain have the permission to manage the parachain when the parachain is unlocked. Parachain is by default locked when onboarded on a slot. This requires the parachain wasm/genesis must be valid otherwise a root track governance action on relaychain is required to update the parachain.
+
+This RFC proposes a mechanism to allow parachain manager self-service the parachain without root track governance action.
```
rphmeier, line 13:
```suggestion
This RFC proposes a mechanism to allow a parachain manager to self-service the parachain without root track governance action.
```

text/0014-improve-locking-mechanism-for-parachains.md
```diff
@@ -0,0 +1,105 @@
...
+The key scenarios this RFC seeks to improve are:
+
+1. Rescue a parachain with invalid wasm/genesis.
+
+While we have various resources and templates to build a new parachain, it is still not a trivial task. It is very easy to make a mistake and resulting an invalid wasm/genesis. With lack of tools to help detect those issues[^1], it is very likely that the issues are only discovered after the parachain is onboarded on a slot. In this case, the parachain is locked and the parachain team has to go through a lengthy governance process to rescue the parachain.
+
+2. Perform lease renewal for an existing parachain.
+
+One way to perform lease renewal for a parachain is by perform a least swap with another parachain with longer lease. This requires the other parachain must be operational and able to perform XCM transact call into relaychain to dispatch the swap call. Combined with the overhead of setting up a new parachain, this is an time consuming and expensive process. Ideally, the parachain manager should be able to perform the lease swap call without having a running parachain[^2].
+
```
rphmeier, line 28: Could you add some hard requirements in MUST/SHOULD terms?

text/0014-improve-locking-mechanism-for-parachains.md
```diff
@@ -0,0 +1,105 @@
...
+## Stakeholders
+
+- Parachain teams
+
+## Explanation
+
+A parachain can either be locked or unlocked[^3]. With parachain locked, the parachain manager does not have any privileges. With parachain unlocked, the parachain manager can perform various actions such as updating the wasm/genesis, perform lease swap, etc.
+
+A parachain is automatically locked once it is onboarded on a slot[^4].
```
rphmeier, line 37: Would you mind expanding this somewhat beyond parachain slots, i.e. a general description of locking/unlocking functionality. And then a section on how this is used in crowdloans. That would be more future-proof.

text/0014-improve-locking-mechanism-for-parachains.md
```diff
@@ -0,0 +1,105 @@
...
+
+One way to perform lease renewal for a parachain is by perform a least swap with another parachain with longer lease. This requires the other parachain must be operational and able to perform XCM transact call into relaychain to dispatch the swap call. Combined with the overhead of setting up a new parachain, this is an time consuming and expensive process. Ideally, the parachain manager should be able to perform the lease swap call without having a running parachain[^2].
+
+## Stakeholders
+
+- Parachain teams
+
+## Explanation
+
+A parachain can either be locked or unlocked[^3]. With parachain locked, the parachain manager does not have any privileges. With parachain unlocked, the parachain manager can perform various actions such as updating the wasm/genesis, perform lease swap, etc.
```
rphmeier, line 35: Be specific about all the actions that the parachain manager can perform

text/0014-improve-locking-mechanism-for-parachains.md
```diff
@@ -0,0 +1,105 @@
...
+
+2. Perform lease renewal for an existing parachain.
+
+One way to perform lease renewal for a parachain is by perform a least swap with another parachain with longer lease. This requires the other parachain must be operational and able to perform XCM transact call into relaychain to dispatch the swap call. Combined with the overhead of setting up a new parachain, this is an time consuming and expensive process. Ideally, the parachain manager should be able to perform the lease swap call without having a running parachain[^2].
+
+## Stakeholders
+
+- Parachain teams
+
+## Explanation
```
rphmeier, line 33: Describe which areas of the runtime will be altered - pallets, interfaces, XCM calls, origins?

text/0014-improve-locking-mechanism-for-parachains.md
```diff
@@ -0,0 +1,105 @@
...
+
+|                 |                                          |
+| --------------- | ---------------------------------------- |
+| **Start Date**  | July 25, 2023                            |
+| **Description** | Improve locking mechanism for parachains |
+| **Authors**     | Bryan Chen                               |
+
+## Summary
+
+The manager of a parachain have the permission to manage the parachain when the parachain is unlocked. Parachain is by default locked when onboarded on a slot. This requires the parachain wasm/genesis must be valid otherwise a root track governance action on relaychain is required to update the parachain.
```
rphmeier, line 11:
```suggestion
The manager of a parachain has permission to manage the parachain when the parachain is unlocked. Parachains are by default locked when onboarded to a slot. This requires the parachain wasm/genesis must be valid, otherwise a root track governance action on relaychain is required to update the parachain.
```

rphmeier: This summary is 99% about the problem, not about the solution. It should be more about the solution.

text/0014-improve-locking-mechanism-for-parachains.md
```diff
@@ -0,0 +1,105 @@
...
+
+A parachain can either be locked or unlocked[^3]. With parachain locked, the parachain manager does not have any privileges. With parachain unlocked, the parachain manager can perform various actions such as updating the wasm/genesis, perform lease swap, etc.
+
+A parachain is automatically locked once it is onboarded on a slot[^4].
+
+Only the relaychain Root origin or the parachain itself can unlock the lock[^5].
+
+This creates an issue that if the parachain is unable to produce block, the parachain manager is unable to do anything and have to rely on relaychain Root origin to manage the parachain.
+
+This RFC proposes to automatically ignore the lock status of the parachain with either of the conditions:
```
rphmeier, line 43: Needs more details - should a new `Call` be added to bypass the lock, which checks these conditions?

text/0014-improve-locking-mechanism-for-parachains.md
```diff
@@ -0,0 +1,105 @@
...
+- Allow parachain to renew lease without actually run another parachain: https://github.com/paritytech/polkadot/issues/6685
+- Always treat parachain that never produced block for a significant amount of time as unlocked: https://github.com/paritytech/polkadot/issues/7539
+
+## Unresolved Questions
+
+How long should we wait before considering a parachain is unable to produce block? It should not be too long otherwise will defects the purpose of this RFC. It should not be too short otherwise will introduce centralization risk. It is proposed to be 7 days.
+
+## Future Directions and Related Material
+
+Everything this RFC touches should be gone in the future once coretime related changes are implemented and therefore future considerations are not applicable.
```
rphmeier, line 99: I do not believe this is true - while we will blur the distinction between parachains with and without slots, the `paras`, `paras_registrar`, and other mechanisms will remain, albeit with some alterations.

xlc: We need a better generalized long term solution which will repalce this one.

xlc:
@shawntabrizi @arrudagates @rphmeier I have updated the RFC to use @shawntabrizi 's mechanism. i.e. only lock parachain when a new block is introduced. Please take another look.

text/0014-improve-locking-mechanism-for-parachains.md
xlc: I am not exactly sure what to put for those without making this too much implementation details. This doesn't alter public interfaces and which pallet to modify shouldn't matter.

text/0014-improve-locking-mechanism-for-parachains.md
xlc: added

text/0014-improve-locking-mechanism-for-parachains.md
xlc: added a status quo section

text/0014-improve-locking-mechanism-for-parachains.md
xlc: in status quo section

arrudagates:
Looks good to me! It is possible to also solve the issue of parachains being unable to swap leases if they halt while being locked, but with Polkadot 2.0 being worked on, that won't be necessary.

gavofyork:
With Agile Coretime and the removal of crowdloans, there's no need for locking.

xlc:
> With Agile Coretime and the removal of crowdloans, there's no need for locking.

That's why this is a short term solution and #16 should result in a long term solution compatible with Agile Coretime.

eskimor (approved):

eskimor:
> With Agile Coretime and the removal of crowdloans, there's no need for locking.

How so? Having a single account that can do whatever to a parachain does not sound desirable for an established parachain with significant funds on it.

rphmeier:
> How so? Having a single account that can do whatever to a parachain does not sound desirable for an established parachain with significant funds on it.

It's definitely part of the security model for a parachain.
One thing a parachain might do is set the account to a burn address, effectively removing the manager altogether.
Another thing a parachain might do is set the account to be a sovereign account of a smart contract running on some other chain, which is used in recovery proceedings.
Or a parachain might set the account to be a normal account on the relay chain, which some individual has a private key for.

Users should simply be aware of this for their trust model.

gavofyork:
> > With Agile Coretime and the removal of crowdloans, there's no need for locking.
> 
> How so? Having a single account that can do whatever to a parachain does not sound desirable for an established parachain with significant funds on it.

What @rphmeier said, though the possibility I was especially considering was setting it to the Sovereign Account of some semi-trusted collective such as the Ecosystem Technical Fellowship or the Polkadot Alliance.

PR #18, opened 2023-08-10
no space in remark body

xlc:


bkchr (approved):

Tbaut (approved):

shawntabrizi (approved):

README.md
```diff
@@ -44,9 +44,9 @@ To open an RFC, follow these steps:
...
 
-The Fellowship may also decide to reject an RFC by issuing a remark with the text `RFC_REJECT(xxxx, h)`. This is a formality to provide clarity on when PRs (or their analogue on non-GitHub platforms) may be closed. PRs may be closed by their author, as well. PRs may be closed when sufficiently stale, as well - after a period of 1 year without acceptance.
+The Fellowship may also decide to reject an RFC by issuing a remark with the text `RFC_REJECT(xxxx,h)`. This is a formality to provide clarity on when PRs (or their analogue on non-GitHub platforms) may be closed. PRs may be closed by their author, as well. PRs may be closed when sufficiently stale, as well - after a period of 1 year without acceptance.
```
rphmeier, line 49: Seems better to clarify that the amount of whitespace is irrelevant

ggwpez: Having a fixed format makes makes it easier for automation. Otherwise when whitespaces are irrelevant it would be possible to submit anything that matches this regex `\w*RFC_REJECT\w*(\w*xxxx\w*,\w*h\w*)\w*`.  
Unless we specify that a singular whitespace char between the `,` and the hash is irrelevant.

xlc: Yeah by not allow whitespace can simplify all the related tooling. We should use a tool to generate the remark body anyway so there isn't much reason to accept whitespace.

PR #20, opened 2023-08-10
Treasurer Track Confirmation Period Duration Modification

Leemo94:


rrtti:
This RFC proposes a change to the duration of the confirmation period for the treasurer track from 3 hours to at least 48 hours.

Yung-Beef:
> * Treasurer track should have much more similar dynamics to Root, with a 7 day confirmation period and a 28 day decision period and linear/near-linear approval/support curves.

Root currently has a 1 day confirmation period, are you implying increasing that 7 days as well?

Leemo94:
@gavofyork currently ``prepare_period``, ``confirm_period`` and ``min_enactment_period`` are the same for tracks on Kusama as they are on Polkadot. For the ``decision_period`` there is a difference whereby a lot of tracks have a 28 day decision period on Polkadot, and a 14 day decision period on Kusama.

With regards to your comment above, would you also propose to apply those changed parameters to Kusama? 

For example, would a 7 day confirmation period for the Treasurer track on Kusama be too long when the decision period is 14 days? 

I really don't know the answer to this one, I just recall that previously there was a desire from the Kusama community/token holders to have OpenGov be a little bit faster than it's original implementation which [caused a few decision periods to be reduced](https://github.com/paritytech/polkadot/pull/6372#discussion_r1038948770) from 28 days to 14 days. I wonder if a similar sentiment will arise if the confirmation period was too long / conservative?

Definitely agree with you in regards to being a _bit_ more conservative on Polkadot.

Thanks for clarifying RE: conviction locks, a few people in the community were wondering if increasing the locks on preexisting convicted funds was something that may happen.

gavofyork:
> @gavofyork Root currently has a 1 day confirmation period, are you implying increasing that 7 days as well?

I guess I am :)

gavofyork:
> With regards to your comment above, would you also propose to apply those changed parameters to Kusama?

Only talking about Polkadot here - basically just stick with the 28 day periods.

> Definitely agree with you in regards to being a _bit_ more conservative on Polkadot.

I think Polkadot would be fine at the current periods; let's not forget that it is the curves which are important more so than the period, which is really just a timeout.

> Thanks for clarifying RE: conviction locks, a few people in the community were wondering if increasing the locks on preexisting convicted funds was something that may happen.

👍

PR #22, opened 2023-08-22
Adopt Encointer Runtime

brenzi:
Encointer is a system chain on Kusama since Jan 2022 and has been developed and maintained by the Encointer association. This RFC proposes to treat Encointer like any other system chain and include it in the fellowship repo with [this PR](https://github.com/polkadot-fellows/runtimes/pull/17)


text/0022-adopt-encointer-runtime.md
```diff
@@ -0,0 +1,55 @@
+# RFC-0022: Adopt Encointer Runtime
+
+|                 |                                                       |
+| --------------- |-------------------------------------------------------|
+| **Start Date**  | Aug 22nd 2023                                         |
+| **Description** | merge Encointer runtime into fellowship runtimes repo |
+| **Authors**     | @brenzi                                               |
```
ggwpez, line 7: I think a legal name would be good, otherwise maybe the legal entity of Encointer also works.

ggwpez (commented):
Thanks for the MR!

I left some formal comments, only one thing about explaining your maintenance plan a bit would be good.

text/0022-adopt-encointer-runtime.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Motivation
+
+Encointer does not seek to be in control of its runtime repository. As a decentralized system, the fellowship has a more suitable structure to maintain a system chain runtime repo than the Encointer association does.
+
+Also, Encointer aims to update its runtime in batches with other system chains in order to have consistency for interoperability across system chains. 
+
+## Stakeholders
+
+* Fellowship: streamlined reviews of system chain runtime upgrades
```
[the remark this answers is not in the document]
brenzi: No, the responsability for the review is already with the fellowship if it whitelists Encointer runtime upgrades as its a system chain and teleporter already

Moving the runtime into the fellowship repo really just simplifies the review process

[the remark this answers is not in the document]
brenzi: suggested a different sentence

text/0022-adopt-encointer-runtime.md
```diff
@@ -0,0 +1,55 @@
...
+* Kusama Network: all system chains in one place
+* Encointer Association: Further decentralization of the Encointer Network necessities like devops.
+* Encointer devs: May be eligible for fellowship membership because of their contributions to the encointer runtime - being a system chain. 
+
+## Explanation
+
+There are no details to elaborate on. It's just about the location of our runtime code. 
+[Our PR](https://github.com/polkadot-fellows/runtimes/pull/17) has all details.
+
```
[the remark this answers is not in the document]
brenzi: added more details. please let us know if that makes sense now

brenzi:
what's the process here? how and when will this matter be decided?

ggwpez (approved):
>what's the process here? how and when will this matter be decided?

I will advertise it in the channel to get some more reviews.  

Would say in two weeks the latest the on-chain voting should be started.

text/0022-adopt-encointer-runtime.md
```diff
@@ -0,0 +1,61 @@
+# RFC-0022: Adopt Encointer Runtime
+
+|                 |                                                                           |
+| --------------- |---------------------------------------------------------------------------|
+| **Start Date**  | Aug 22nd 2023                                                             |
+| **Description** | Permanently move the Encointer runtime into the Fellowship runtimes repo. |
+| **Authors**     | @brenzi for Encointer Association, 8000 Zurich, switzerland               |
```
ggwpez, line 7:
```suggestion
| **Authors**     | @brenzi for Encointer Association, 8000 Zurich, Switzerland               |
```

gilescope:
I'm happy to sponsor this RFC and maintain it until such time as encointer devs join the fellowship.

text/0022-adopt-encointer-runtime.md
```diff
@@ -0,0 +1,61 @@
...
+
+[Our PR](https://github.com/polkadot-fellows/runtimes/pull/17) has all details about our runtime and how we would move it into the fellowship repo.
+
+Noteworthy: All Encointer-specific pallets will still be located in encointer's repo for the time being: https://github.com/encointer/pallets 
+
+It will still be the duty of the Encointer team to keep its runtime up to date and provide adequate test fixtures. So far, Encointer pallets have proven to be very stable and the runtime only needed to be updated every few months. 
+However, frequent dependency bumps with Polkadot releases would be beneficial for interoperability and could be streamlined with other system chains collectively by the fellowship. This would allow to upgrade all system chains jointly (including Encointer) regularly with a batch referendum 
+
+Further notes:
+* Encointer currently releases none of its crates on crates.io
```
bkchr, line 36: This would need to change. We only want to have crates on crates.io if possible.

text/0022-adopt-encointer-runtime.md
```diff
@@ -0,0 +1,61 @@
...
+* Encointer devs: Being able to work directly in the Fellowship runtimes repo to streamline and synergize with other developers. 
+
+## Explanation
+
+[Our PR](https://github.com/polkadot-fellows/runtimes/pull/17) has all details about our runtime and how we would move it into the fellowship repo.
+
+Noteworthy: All Encointer-specific pallets will still be located in encointer's repo for the time being: https://github.com/encointer/pallets 
+
+It will still be the duty of the Encointer team to keep its runtime up to date and provide adequate test fixtures. So far, Encointer pallets have proven to be very stable and the runtime only needed to be updated every few months. 
+However, frequent dependency bumps with Polkadot releases would be beneficial for interoperability and could be streamlined with other system chains collectively by the fellowship. This would allow to upgrade all system chains jointly (including Encointer) regularly with a batch referendum
```
bkchr, line 33: I think the fellowship can not commit on doing these dependency bumps etc. With the switch to using `crates.io` releases, this should also not really be required. Encointer could continue to use older releases of the crates until someone updates it. This doesn't mean that someone updating the other system chains, not also updates Encointer, but there should not be any real requirement in doing this.

bkchr (commented):
Generally I'm okay with this. I don't think that Encointer will make the job more complicated for the fellowship. The fellowship runtimes repo only uses crates from crates.io. This means that actually every runtime can update in its own speed and doesn't block the others.

text/0022-adopt-encointer-runtime.md
brenzi: OK. we can do that. we didn't succeed in upstreaming all our no_std patches to our dependencies, but then we just need to publish all our forks to crates.io

text/0022-adopt-encointer-runtime.md
brenzi: I'll rephrase

brenzi:
assuming the comments have been adressed, can we move forward with this? I'll test the bot and my new assumed privilege right away

bkchr (approved):

olanod:
I can not vote but anyway I would *aye* it if it's what you guys really want(are all RFCs gonna be voted by ranks 3+ only?), Encointer has been a role model for us to follow that showed us it's *possible*(with a lot of pain) to to build a common good with resources from the ecosystem.

My comment during last fellows meetup was just an observation, you guys asked to be a common good when it was not yet well defined what it meant to be one, now we shifted towards the concept of system chains instead of common good ones and Encointer got dragged into this new definition during the transition, but I think **Encointer is not a system chain**, it's a common good that deserves to be funded by the treasury but it's not an extension of the core protocol that would fit with other fellowship maintained runtimes.

I'm curious if you guys have considered giving up your system chain status and "be free", as a very experimental project you might want to iterate often, change and break things, I feel that being dependent on Kusama's governance to do runtime upgrades it's a huge drag that damages the project's ability to adapt to the needs of your users.  
While developing Virto(Kreivo now on Kusama) I always had the goal to *do as Encointer*, don't have a token(make it hard for the L1 to fall into the control of the same few) and aim to do something nice, something for the common good(but remain sustainable). However if we want to stay kreiv-ative giving up the control of our runtime was a no-go, so I just waited patiently(some years now?) for parathreads to be a thing while doing other work on the client side. Now the wait is over, on-demand blockspace is not there yet but Kusama slots being "free" meant our time has come, we still build a common good chain(whether it's an official definition or not), we can iterate quickly and if the ecosystem feels it's worth supporting the ideas and innovations that can come out this project then we can keep the chain alive. Wouldn't this kind of model fit your project better?

brenzi:
@olanod Thank you for your thoughtful comment.

I have previously commented on [my take](https://github.com/polkadot-fellows/runtimes/pull/17#issuecomment-1711422542). Like you, I don't think the "system chain" label is a good fit for Encointer. But then, it's mostly a marketing label and I have no strong feelings here.

Concerning our freedom to iterate often and change/break things, I don't believe that joining this repo changes much. Lacking our own root origin, we have to wait for OpenGov approval anyway and it's the Fellowship which can accelerate by whitelisting, so we need to wait anyway. 

From a purely technical point of view, we would have been much better off as a solochain the last two years: Less downtime, faster bugfix deployment, faster blocktime. But at our current state, "solo" would have meant de facto "centralized" and that's not what we're in for. Decentralization is an integral part of our field-experiment. 

Grabbing a cheap parachain slot on Kusama could be a viable alternative if we establish personhood-based democratic governance for our chain. However, in that case we might chose to go one step further and elect our own validators just as well and go straight to a very interesting experiment of delegated personhood consensus. This, however, would require quite substantial groundwork in advance.

So, back to where we are now: We submitted this RFC to get a clear statement from Fellowship on the currently perceived status of Encointer by the technocratic arm of OpenGov: Does the Fellowship think that Encointer should be more tightly integrated or not? We can accept both outcomes, but we request clarity on the matter.

brenzi:
/rfc process 0x3c07696acaba98022d0960a0c14da01d96bd747eead9380b3aac5c1f5c38b046

Issue #21, opened 2023-08-17
Bot for automatic merging apprroved RFCs/closing rejected RFCs


bkchr:
After RFCs are approved on chain, they can be merged directly. This should be done by a bot and should not require any manual intervention. The same applies for rejects.

The bot should monitor the fellowship referenda for referendums that want to enact a `remark` with `RFC_APPROVE` or `RFC_REJECT`. The exact process on how a referendum is executed, is documented in the `README`. After the referendum was approved, the bot should do the appropriate action(merge for `RFC_APPROVE` and close for `RFC_REJECT`).

xlc:
To keep things simple, we don’t necessarily need to constantly monitor the chain. We could for example have a github action trigger that takes a block number as input, and manually trigger an action based on the event in that block. In this way we don’t have external server dependency.

rzadp:
Related (or duplicate?): https://github.com/paritytech/opstooling/issues/283

xlc:
It is just like how most of the onchain contracts works. The monitor is offchain and trigger is a permisionless onchain action that verifies some conditions and enforce it.

In this case, the monitor is manual, or could be a server, doesn't matter. The trigger is a permisionless Github Action bot. It could be a comment to the PR for example that anyone can do it. It is the bot/GH action verifies if the RP is indeed approved and then perform the merge. The whole point is following the onchain decision, not depending on some specific admins on this repo.

tomaka:
My two cents is that a manual bot is a better idea.

If the bot is automatic, that means we need a server, which raises the question of who maintains and monitors the server, and we all know that this will be half-assed and that the bot will likely break down all the time.

bkchr:
> It could be a comment to the PR for example that anyone can do it.

Like the bot @rzadp for creating the proposal. I like the idea!

xlc:
I already have a working PoC.

Demo: https://github.com/xlc/RFCs/pull/14
GH action code: https://github.com/xlc/fellowship-process-bot

xlc:
It is a PoC and therefore don’t cover every edge cases. But it is easy to support in-line proposal. 

The pjs apps UI doesn’t offer a way to create inline proposal other than manual construction.

rzadp:
I have ported my thing to [work as a GitHub Action](https://github.com/paritytech/rfc-propose).

Here it is in action: https://github.com/paritytech/rfc-propose/pull/5

xlc:
[...]

PR #24, opened 2023-08-31
Add a GitHub Action proposing the creation of RFC referenda

rzadp:
- The action aims to help with the creation of the RFC on-chain referenda.
- The action will respond to `/rfc-propose` comments on RFC PRs.
- The output of the action will [look like that](https://github.com/paritytech-stg/RFCs/pull/5#issuecomment-1700852905).
- Related to https://github.com/polkadot-fellows/RFCs/issues/21

### DIFF
--- a/.github/workflows/rfc-propose.yml
+++ b/.github/workflows/rfc-propose.yml
@@ -1,0 +1,18 @@
+name: RFC propose
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  pull-requests: write
+
+jobs:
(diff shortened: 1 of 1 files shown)


rzadp:
> So what needs to be done? Just merge it and then it should work? Or do I need to register the workflow somewhere or something like this?

Should work after merging, unless actions outside of `polkadot-fellows` are disabled in repository `settings/actions`.
(I don't have access there so can't see)

bkchr:
Seems to work! Nice work @rzadp :)

PR #26, opened 2023-09-03
Sassafras Consensus

davxy:
Draft

### DIFF
--- a/text/0000-sassafras-consensus.md
+++ b/text/0000-sassafras-consensus.md
@@ -1,0 +1,785 @@
+# RFC-xx: Sassafras Consensus Protocol
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | September 03, 2023                                                                          |
+| **Description** | Sassafras consensus protocol description and structures                                     | 
+| **Authors**     | Davide Galassi                                                                              |
+
+
+## Abstract
+
+Sassafras is a novel consensus protocol designed to address the recurring
+fork-related challenges encountered in other lottery-based protocols, such as
+Babe.
+
+Sassafras aims to establish a unique association between each epoch's slots
+and the validators, ensuring that there will be one and only one validator
+per slot.
+
+The protocol ensures the anonymity of the validator associated to a slot until
+the slot is not claimed at block production time.
+
+
+## 1. Motivation
+
+
+Sassafras Protocol has been extensively documented in a comprehensive
+[research paper](https://eprint.iacr.org/2023/031.pdf). However, this RFC serves
+the purpose of conveying essential implementation details that are crucial for
+interoperability and clarifying aspects left open for implementation discretion.
+
+### 1.1. Relevance to Implementors
+
+This RFC focuses on providing implementors with the necessary insights into the
+protocol's operation. It takes precedence over the research paper in cases where
+discrepancies arise between the two documents.
+
+(TODO: remove this)
+Example: In this RFC, ticket claims occur in the epoch immediately following
+issuance, whereas the original protocol description specifies a two-epoch gap.
+The approach outlined in this RFC should be followed.
+
+### 1.2. Supporting Sassafras for Polkadot
+
+In addition to fostering interoperability, another objective of this RFC is to
+facilitate the implementation of Sassafras within the Polkadot ecosystem. While
+the specifics of deployment mechanics are beyond the scope of this document, it
+paves the way for integrating Sassafras into the Polkadot network.
+
+
+## 2. Stakeholders
+
+### 2.1 Developers of Relay-Chains and Para-Chains
+
+Developers responsible for creating relay-chains and para-chains within the
+Polkadot ecosystem who intend to leverage the benefits offered by the Sassafras
+Protocol.
+
+### 2.2 Developers of Polkadot Relay-Chain
+
+Developers contributing to the Polkadot relay-chain, which plays a pivotal role
+in facilitating the interoperability and functionality of the Sassafras Protocol
+within the broader Polkadot network.
+
+## 3. Notation and Convention
+
+This section outlines the notation and conventions used throughout the document
+to ensure clarity and consistency.
+
+### 3.1. Data Structure Definitions
+
+Data structures are primarily defined using [ASN.1](https://en.wikipedia.org/wiki/ASN.1),
+with a few exceptions:
+- Integer types are not explicitly defined in ASN.1 and `Un` types should be
+  interpreted as n-bit unsigned integers.
+
+To ensure that serialized data structures can be used interchangeably, it's
+important to consider the arrangement of fields in their definitions.
+
+In cases where no specific instructions are given, structures should be
+serialized using [SCALE](https://github.com/paritytech/parity-scale-codec).
+
+### 3.2. Pseudo-Code
+
+Through this document it is advantageous to make use of code snippets as part
+of the comprehensive description. These code snippets shall adhere to the
+subsequent conventions:
+
+- Code snippets are presented in a Rust-like pseudo-code format.
+
+- The function `SCALE(x: T)` returns the `SCALE` encoding of the variable `x`
+  with type `T`.
+
+- The function `BYTES(x: T)` returns an `OCTET STRING` which represents the raw
+  byte array representation of the object `x` with type `T`.
+  - if `T` is `VisibleString` (aka *ascii* string): it returns the sequence of
+    bytes of its *ascii* representation.
+  - if `T` is `Un`: it returns the little-endian encoding of the integer `Un`
+    as `n` bytes.
+
+- The function `Un(x: OCTET STRING)` returns a `Un` using the little-endian
+  encoding of the integer `x`
+
+- The function `TRANSCRIPT(x)` represents the construction of a transcript
+  object as defined by [`ark-transcript`](https://docs.rs/ark-transcript/0.0.1/ark_transcript/)
+  library.
+
+### 3.3. Incremental Introduction of Types and Functions
+
+Types and helper functions will be introduced incrementally as they become
+relevant within the document's context.
+
+We find this approach more agile, especially given that the set of types used is
+not extensive or overly complex.
+
+This incremental presentation enhances readability and comprehension.
+
+
+## 4. Protocol Introduction
+
+The Sassafras Protocol employs a binding mechanism between validators and slots
+through the use of a **ticket**.
+
+The protocol is organized into five discrete and asynchronous phases:
+
+### 4.1. Submission of Candidate Tickets
+
+Validators generate and submit their candidate tickets along with validity
+proofs to the blockchain. The validity proof ensures that the tickets are
+legitimate while maintaining the anonymity of the authorship.
+
+### 4.2. Validation of Candidate Tickets
+
+Submitted candidate tickets undergo a validation process to verify their
+authenticity, integrity and compliance with other protocol-specific rules.
+
+### 4.3. Tickets and Slots Binding
+
+After collecting all candidate tickets, a deterministic method is employed to
+associate some of these tickets with specific slots.
+
+### 4.4. Claim of Ticket Ownership
+
+Validators assert ownership of tickets during the block production phase. This
+step establishes a secure binding between validators and their respective slots.
+
+### 4.5. Validation of Ticket Ownership
+
+During block verification, the claims of ticket ownership are rigorously
+validated to uphold the protocol's integrity.
+
+
+## 5. Bandernatch VRFs Cryptographic Primitives
+
+This chapter provides a high-level overview of the Bandersnatch VRF primitive as
+it relates to the Sassafras protocol.
+
+It's important to note that this section is not intended to serve as an
+exhaustive exploration of the mathematically intensive foundations of the
+cryptographic primitive. Instead, its primary purpose is to offer a concise and
+comprehensible interpretation of the primitive within the context of this RFC.
+
+For a more detailed and mathematical understanding of Ring-VRFs, we recommend
+referring to the Ring-VRF research [paper](https://eprint.iacr.org/2023/002.pdf).
+
+### 5.1. VRF Input
+
+The VRF Input, denoted as `VrfInput`, is constructed using a domain and some
+arbitrary data using the `vrf_input(domain: OCTET STRING, buf: OCTET STRING)`
+function. This function is left opaque here and can be read in the actual
+reference implementation.
+
+The VRF Input, denoted as `VrfInput`, is constructed by combining a domain identifier
+with arbitrary data using the `vrf_input` function:
+
+```rust
+    fn vrf_input(domain: OCTET_STRING, buf: OCTET_STRING) -> VrfInput;
+```
+
+While the specific implementation details of this function are intentionally
+omitted here, you can find the complete implementation in the
+[`bandersnatch_vrfs`](https://github.com/w3f/ring-vrf/blob/18614458ca4cb335c88d4e710c13906a76f51e43/bandersnatch_vrfs/src/lib.rs#L57) 
+reference implementation.
+
+Helper function to construct a `VrfInput` from a sequence of `data` items:
+
+```rust
+    fn vrf_input_from_items(domain: OCTET_STRING, data: SEQUENCE_OF OCTET_STRING) -> VrfInput {
+        buf = OCTET_STRING(SIZE(0));
+        for item in data {
+            buf.append(item);
+            buf.append(length(item) as U8);
+        }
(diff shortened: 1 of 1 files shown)


text/0026-sassafras-consensus.md
```diff
@@ -0,0 +1,991 @@
...
+left undefined as are not relevant. Their definitions can be found in the
+`bandersnatch_vrfs` reference implementation.
+
+
+## 6. Sassafras Protocol
+
+### 6.1 Epoch's First Block
+
+The first block produced for an epoch `N` is required to include the descriptor
+for the next epoch `N+1`.
```
tomaka, line 366: What about block 0? Does it belong to an epoch? Does it contain a descriptor?

tomaka (commented):
I haven't reviewed everything carefully, but from a light client perspective I need a way to verify the authenticity of block headers without verifying block bodies, and for warp syncing purposes I need a way to retrieve the current state of Sassafras (i.e. anything necessary to verify the block headers of the children of a block) through runtime calls or storage reads.

davxy:
> I haven't reviewed everything carefully, but from a light client perspective I need a way to verify the authenticity of block headers without verifying block bodies, and for warp syncing purposes I need a way to retrieve the current state of Sassafras (i.e. anything necessary to verify the block headers of the children of a block) through runtime calls or storage reads.

Block verification is performed on the header. You need to fetch some data from chain state to check if the claimed slot is associated to a ticket

tomaka:
> Block verification is performed on the header. You need to fetch some data from chain state to check if the claimed slot is associated to a ticket

The RFC should give some details about that, with the names of the runtime functions, layout of the data returned (which must be stable), etc.

burdges:
> I need a way to verify the authenticity of block headers without verifying block bodies

@eskimor We need a Merkle proof into the chain state for the sassafras ticket and validator's public key.  You guys did a bunch of this in async backing, so can one of you guys who @davxy how that worked?

burdges:
Actually @tomaka do you have an example of using a Merkle proof for a validator key?  We must do this for babe already I guess, no?

tomaka:
The way it's done in smoldot for Babe is by calling three runtime functions that return data structures containing all the information needed: the list of current and next epoch validators, current epoch number, slots per epoch, etc

It could in theory also simple be done by reading directly from the storage. The fact that it's a runtime function call is just for pragmatic purposes. It's to make it possible to change the format and location of the values in storage without breaking what the function returns. Also runtime functions have a version number, which again is useful for backwards compatibility, while plain storage items don't.

So the Merkle proof is actually a "call proof". The proof generator calls the function(s) and builds a proof containing all the trie nodes that are accessed during the call.

text/0026-sassafras-consensus.md
```diff
@@ -0,0 +1,988 @@
...
+            buf.append(length(item) as U8);
+        }
+        return vrf_input(domain, buf);
+    }
+```
+
+### 5.2. VRF Output
+
+Given a `VrfInput` object, the corresponding `VrfOutput`, also referred to as
+`VrfPreOutput`, is computed using a Bandersnatch secret key.
```
burdges, line 198: I only call this `VrfPreOutput` since I wanted to deprecate the `VrfOutput` being used here.  Also `vrf_output` is now a different function.

Alone `preoutput` is pretty damn random, but it's algebraic so it carries "interesting" relationships between different uses, especially related keys.

text/0026-sassafras-consensus.md
```diff
@@ -0,0 +1,988 @@
...
+This section defines the data that is to be signed using the VRF primitive:
+
+```rust
+    VrfSignatureData ::= SEQUENCE {
+        transcript: Transcript,
+        vrf_input: SEQUENCE_OF VrfInput 
+    }
+```
+
+- `transcript`: represents a `ark-transcript` object.
```
burdges, line 244: append "a hash of the signed data which should not influence the vrf outout"

text/0026-sassafras-consensus.md
burdges: Also, this is more copying this information in state than including "in the block", but whatever.

text/0026-sassafras-consensus.md
```diff
@@ -0,0 +1,988 @@
...
+Each epoch has an associated randomness value defined by the
+`NextEpochDescriptor` `Randomness` element.
+
+The first block of epoch `N` contains the randomness associated to the next epoch `N+1`.
+
+Randomness for epoch `N+1` is computed using the VRF output values which are associated
+to each block of epoch `N-1`.
+
+Assuming block `Bi` is produced during epoch `N-1`. Then the randomness for epoch `N+1`
+is incrementally generated as `RandomnessAccumulator = Blake2(Bi.VrfOut)'`.
```
burdges, line 405: `Bi.RandomnessAccumulator = Blake2(Bi.parent.RandomnessAccumulator, Bi.claim.vrf_output())` or something like that.

text/0026-sassafras-consensus.md
```diff
@@ -0,0 +1,988 @@
...
+```rust
+    ProtocolConfiguration ::= SEQUENCE {
+        attempts_number: U32,
+        redundancy_factor: U32
+    }
+```
+
+- `attempts_number`: number of tickets that can be submitted by each next-epoch authority.
+- `redundancy_factor`: factor that enhances the likelihood of a candidate ticket
+  being deemed valid according to some protocol specific checks.
```
burdges, line 425: I think this section should describe the consequences of the parameters. 
```
- `attempts_number` influences the anonymity of block producers.  All tickets have a public attempt number less than `attempts_number`.  If two tickets share an attempt number then adversaries know they must be by different block producers, which reduces anonymity late in the epoch.  We do not mind `max_attempts < epoch_length` though because this loss already becomes small when `attempts_number = 64` or `128`.  And larger requires more computation.
- `redundancy_factor` controls the expected number of extra tickets produced beyond `epoch_length`.  
```
I forget exactly what `redundancy_factor` really does right now, but we want it to be meaningful, not just 1 or 0.  An option goes like:

We need a probability `p` of a particular attempt yielding a valid ticket, so that roughly `n p - redundancy_factor * sigma > epoch_length` where `n = num_validators * attempts_number` and `sigma = n p (1-p)` is the standard deviation under the normal approximation.  Approximating `sigma` by `sigma' = n p' (1 - p')` with `p' = epoch_length / n` then `p = (epoch_length + redundancy_factor * sigma) / n`.  Aka there being aura-like blocks in the middle of the epoch is almost a `redundancy_factor` sigma event.

text/0026-sassafras-consensus.md
```diff
@@ -0,0 +1,1079 @@
...
+    next_epoch_randomness = H(curr_accumulator ++ curr_epoch_randomness ++ next_epoch_index)
+
+    With:
+    - curr_accumulator depending on prev_accumulator
+    - next_epoch_index depending on curr_epoch_index 
+
+    So the only extra thing curr_epoch_randomness brings to the table is `...`.
+    But if we recursively unroll this we end up `...` being the genesis randomness
+    which is always 0.
+</TODO>
```
davxy, line 441: @burdges I need your opinion about this snip I've marked with `TODO`.

In short here I ask (in the context of computing the randomness of next epoch) what is the point of adding current epoch randomness when this is a function of the accumulator (which we already use in the hash).

Well at rigor, current epoch randomness is a function of the accumulator one epoch ago, but current accumulator is a function of it as well.

I hope my point 🫨

text/0026-sassafras-consensus.md
```diff
@@ -0,0 +1,1088 @@
...
+### 12.1. Interactions with the Runtime
+
+- **Outbound Interface**. Interfaces exposed by the host which are required by the runtime.
+  These are commonly dubbed *Host Functions*.
+
+- **Unrecorded Inboud Interfaces**. Interfaces exposed by the runtime which are required by the host.
+  These are commonly dubbed *Runtime APIs*.
+
+- **Transactional Inboud Interfaces**. Interfaces exposed by the runtime which alter the state.
+  These are commonly dubbed *Extrinsics* and *Inherents*.
```
davxy, line 1063: @tomaka given that this RFC already contains quite a lot of information, I thought that may be worth dedicating an RFC to all the topics which concerns the actual integration with Polkadot.

This RFC is more intended to be a bridge between the research and the engineering. And propose the actual structures and procedures to implement the protocol.

I want to propose soon an RFC containing all the interactions which are expected to happen between the runtime, the host and the external world. That is runtime api, host functions, storage entries, extrinsics and inherents.

IMO is worth to keep these topics separated as:
- there is a separation between the protocol description and the actual integration with a system (i.e. Polkadot)
- maybe is better to don't bloat too much this RFC

text/0026-sassafras-consensus.md
```diff
@@ -0,0 +1,1079 @@
...
+            BYTES(slot)
+        ]
+    );
+
+    randomness_vrf_output = claim.signature.outputs[0];
+
+    randomness = vrf_bytes(32, randomness_vrf_input, randomness_vrf_output);
+
+    accumulator = BLAKE2(32, CONCAT(accumulator, randomness));
+```
```
burdges, line 983: `accumulator` just accumulates forever here, right?

We should not reset `accumulator = ""` anyware except maybe in genesis (and maybe the genesis hash or whatever is a better choice)

text/0026-sassafras-consensus.md
burdges: It's fine if `accumulator` just accumulates forever.  We need `curr_epoch_randomness` to be copied from `accumulator` far enough in the past that we have consensus on its current value even if grandpa dies.  So:

Every block does `accumulator_after_block = H(accumulator_before_block, randomness)` like you describe below.

An epoch end does
```
curr_epoch_randomness = next_epoch_randomness;
next_epoch_randomness = H(accumulator_before_block,next_epoch_index); 
```
Including `next_epoch_index` in this hash means we can leave it out of the `vrf_input` hashes.

Does the after vs before matter here?  We obtain finality by waiting for a whole epoch, which becomes a sloppy notion, so they're not overly critical.  If otoh someone can rewind the chain via disputes or whatever, then corner cases matter, so maybe it easier to just be precises here that `next_epoch_randomness` gets get from the `accumulator` coming out of the final block of the previous epoch.  It's likely easier to code this way.

text/0026-sassafras-consensus.md
davxy: Never resets. On genesis, I'll put there genesis hash 👍

text/0026-sassafras-consensus.md
```diff
@@ -0,0 +1,1079 @@
...
+    index_bytes = BLAKE2(4, SCALE( (epoch_randomness, slot) ));
+    index = U32(index_bytes) mod authorities_number;
+```
+
+<TODO>
+What about using `epoch_randomness_accumulator` instead of `epoch_randomness`?
+The accumulator is updated using the randomness which ships with every block, thus
+we know who is the author of block N only after block N-1 has been imported.
+Is a bit more Dos resistant.
+</TODO>
```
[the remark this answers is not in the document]
burdges: We'd want authorities to be elected at least one epoch before the randomness that determines their order, and for that randomness to be determined at the end of `curr_epoch - 2`.

We should clarify how we name randomness, but I guess randomness accumulated in epoch n is referred to as the randomness of epoch n+1 maybe?  We never use it in that epoch, bu that's fine.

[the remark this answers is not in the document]
davxy: Yes, the authorities are elected one epoch before the randomness that determines their order.

---

Yes, the randomness accumulated for epoch n is used as the input for H to compute the randomness for epoch n+2.

As you can see from the RFC, the current formula is something like:
 
Randomness(n+2) = H( Accumulator(n), target_epoch_index = n+2, Randomness(n+1) )

The question was: can we remove Randomness(n+1) from the hash? I.e. using:

Randomness(n+2) = H( Accumulator(n), target_epoch_index = n+2 )

The question stems from the fact that Randomness(n+1) is somehow derived from an older value of the accumulator (which never resets) so I don't see the point of including it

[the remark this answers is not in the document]
burdges: We can & should remove that randomness(n+1) yes, if only to make the code easier to read

burdges:
Around the randomness-ticket cycle here, we have six stages which occur in sequence:  
0. authority set determined defined in previous epoch,
1. accumulate block randomnesses for full epoch to define future randomness r,
2. wait for probabilistic finality on r,
3. make & publish tickets using r,
4. wait for probabilistic finality on ticket set that uses r, 
7. claim tickets in full epoch.

We always want 1 & 5 to be full epochs of course, but exactly how 2-4 align with epochs remains a question.  If each of 2-4 requires a full epoch like in praos, then we've many epochs between becoming an authority and doing the work, which causes annoyances elsewhere in the protocol.

We instead optimized this by placing 2-4 into one single epoch, so the whole cycle does not last longer than in praos, but this means how long each of 2-4 take becomes a configuration parameter, along with the full epoch length, so how do we choose these?

A full epoch should be long enough so that we believe one honest node's block randomness gets accumulated, so that a randomness adversary loses control over the randomness.  They can still bias randomness, but this bias becomes relatively small.

We need 4 to be long enough to have probabilistic finality upon the set of ticket, meaning so adversaries cannot fork out blocks that posted tickets during 3.  We need 3 to be long enough that all valid tickets get posted to the chain, which depends upon the chain's censorship resistance.  

We need 2 to be long enough to have probabilistic finality on the accumulation r of randomness in the previous epoch.  If 2 winds up too short, then we do not fail per se but now different sassafras tickets make sense for each surviving chain fork.  In other words, prover and verifier time wind up multiplied by the number of forks, which becomes considerable CPU time.  Although bad, I think 2 could run into 3 slightly, which gives us some slack for 2.

A priori, I'd suggest 4 be roughly half the full epoch, with 2 and 3 being a quarter epoch each.  If 3 needs more time, then we could go as small as 2 being 1/6th of an epoch, and 3 being 1/3rd of an epoch.  If we need more than 1/3rd of an epoch for 3 then we should discuss optimizations I think.

We sort the ticket set during 4 too, but when in 4 matters.  If too early, then we'll sort separately on each fork.  If too late, then we might end the epoch without the sorted ticket list.  Aka we have 
4a. wait for ticket set
4b. sort ticket set
4c. wait for sorted ticket set

A priori I'd suggest we sort right in the middle, so all told our initial config should be:  1st quarter does 2.  2nd quarter does 3.  3rd quarter minus epsilon does 4a.  around 3rd quarter mark do 4b.  4th quarter does 4c.

text/0026-sassafras-consensus.md
```diff
@@ -0,0 +1,921 @@
...
+#### 6.5.3. Ticket Body
+
+Every ticket candidate has an associated body, defined as:
+
+```rust
+    TicketBody ::= Sequence {
+        attempt_index: Unsigned8,
+        opaque: OctetString,
+    }
+```
```
burdges, line 513: Is opaque where the erased key and revealed key live?  It's fine if erased key gets ignored in the spec, but it'd be messy to abstract over whether it's there or not.

In a claim, the erased signature is on the claim itself, making it part of the defintiion of a claim.  It'll become how nodes contact block producers eventually too, which even backs it into the networking.

We made the erased key signature is an `Option` in the claim, because doing so let the chain produce blocks in slots claimed by nodes who restarted, but we could alter core node beahvior significantly based upon whether blocks have erased key signatures.  I've not pushed this angle so far, but it's the original motivation, handy if adversaries become more advanced, and it's quite useful in forensics anyways.

davxy: So the idea here is to simplify the specs and align some stuff to JAM Safrole.

As far we understood these two keys are not **strictly** required for the operations of the protocol. I know, are nice tools offering some extra properties, but not **strictly** required for Sassafras to work.

In the end, we rely on the VRF output to check if the slot can be claimed by the block producer and the ephemeral keys were some "bonus".

**If** there exist some potential security issue using the VRF output only then we have a bigger problem as JAM only uses it and doesn't rely on the ephemeral keys.

All in all I strive to keep this spec simple and not over-engineered for problems we don't actually have.

The "opaque" octet-string you see here is just some data without any particular meaning to the protocol and can be used by the ticket producer to commit stuff that may be useful for other contexts (don't know... as an example  may be used to post some auxiliary "channel" that can be used by future tickets submitters to forward tickets to the block producer.

davxy:
> Voting for this referenda is **ongoing**.
> 
> Vote for it [here](https://collectives.polkassembly.io/referenda/131)

Referenda moved here: https://collectives.polkassembly.io/referenda/133

davxy:
/rfc process 0x705dbafd4a25aa0a99ffa785dfd26c59a4ec11a7a13db9abbc59c68a0c98db10

PR #32, opened 2023-10-02
Proposal for Minimal Relay

joepetrowski:
This RFC proposes a direction and prioritization for migrating core functionality off the Relay Chain. The focus is on Identity, Staking, and Governance. The RFC includes a brief discussion on the challenges associated with each one and the most probable migration/implementation path.


text/0032-minimal-relay.md
```diff
@@ -0,0 +1,190 @@
...
+Identity will be one of the simpler pallets to migrate into a system chain, as its logic is largely
+self-contained and it does not "share" balances with other subsystems. As in, any DOT is held in
+reserve as a storage deposit and cannot be simultaneously used the way locked DOT can be locked for
+multiple purposes.
+
+Therefore, migration can take place as follows:
+
+1. A runtime upgrade will block all calls to the pallet, preventing updates to identity info.
+1. The frozen state will form the genesis of a new system parachain.
+1. A Relay Chain runtime upgrade will allow migrating the deposit to the parachain. The parachain
```
xlc, line 102: how does one get refunds of the reserves?

text/0032-minimal-relay.md
```diff
@@ -0,0 +1,190 @@
...
+
+A pub-sub protocol may also lend itself to these types of interactions.
+
+### Deployment
+
+Actual migrations should happen based on some prioritization. This RFC proposes to migrate Identity,
+Staking, and Governance as the systems to work on first. A brief discussion on the factors involved
+in each one:
+
+#### Identity
```
xlc, line 91: to confirm, we will have a system parachain just for identity?

text/0032-minimal-relay.md
```diff
@@ -0,0 +1,190 @@
...
+for situations like [parachains
+halting](https://forum.polkadot.network/t/stalled-parachains-on-kusama-post-mortem/3998).
+
+The only complication arises from the fact that both Asset Hub and the Staking parachain will have
+DOT balances; therefore, the Governance chain will need to be able to credit users' voting power
+based on balances from both locations. This is not expected to be difficult to handle.
+
+## Drawbacks
+
+None at present.
```
xlc, line 149: Increased complexity as existing dApps/wallets have to connect to multiple chains to maintain existing functionality.
Less compute power for staking due to pov limitation.

text/0032-minimal-relay.md
```diff
@@ -0,0 +1,190 @@
...
+Describe the impact of the proposal on the exposed functionality of Polkadot.
+
+### Performance
+
+This is an optimization. The removal of public/user transactions on the Relay Chain ensures that its
+primary resources are allocated to system performance.
+
+### Ergonomics
+
+This proposal alters very little for coretime users (e.g. parachain developers). Application
```
xlc, line 167: Parachains that interact with identity/staking/governance will need to coordinate with the migration to avoid outage.

text/0032-minimal-relay.md
```diff
@@ -0,0 +1,190 @@
...
+Application developers will need to interact with multiple chains in the network.
+
+## Prior Art and References
+
+- [Transactionless Relay-chain](https://github.com/paritytech/polkadot/issues/323)
+- [Moving Staking off the Relay Chain](https://github.com/paritytech/polkadot-sdk/issues/491)
+
+## Unresolved Questions
+
+There remain some implementation questions, like how to use balances for both Staking and
```
xlc, line 183: I don't think we need identity as its own chain. Then we have the question of how to determine if a feature should lives on its own parachain or coexisting with another parachain. e.g. we could have the identity pallet also lives on the governance parachain.

xlc: The current migration paln results downtime. i.e. we have to freeze identity state, setup new genesis, run the new chain, and then it is up again. This is likely going to take weeks. Can we do better to reduce the downtime?

text/0032-minimal-relay.md
xlc: have you considered to merge identity for Kusama and Polkadot? like we have a single fellowship for both networks. I see no reason to have two set of identities on each networks.

text/0032-minimal-relay.md
```diff
@@ -0,0 +1,190 @@
...
+- Parachains that interact with affected logic on the Relay Chain;
+- Core protocol and XCM format developers;
+- Tooling, block explorer, and UI developers.
+
+## Explanation
+
+The following pallets and subsystems are good candidates to migrate from the Relay Chain:
+
+- Identity
+- Preimage
```
shawntabrizi, line 42: It seems hard to migrate the preimages from the relay chain since these are almost by definition storage items which are heavy and large. These need to stay on the relay chain so that there isnt some kind of PoV limit which bottlenecks access to preimages.

shawntabrizi: Many of the medium term scaling solutions to me actually involve allowing others to post data onto the relay chain directly, and having mechanisms which allow the parachain validation to access those preimages on the relay chian.

https://github.com/paritytech/polkadot-sdk/issues/811

burdges: It's only mentioned here, not elsewhere, so not sure what this means, but..

In principle wasm or risc-v preimage could live in availability, where they get reposted/retained periodically.  We want all artifacts to be prebuilt on all the validators, so the validators would've downloaded all the preimages in the past, built them, and retained the build artifacts.  Any new validators fetch them from availability.  I seriously the savings here warrant the development time, but who knows maybe this becomes easy.

Is more possible?  In principle artifacts could be prebuilt only by approval checkers, which proves the artifact to be buildable. We'd face several problems here:
- Artifact could build today but no longer build tomorrow, due to some parameter change, host upgrade, etc.
- Artifacts needed build times within our timeouts previously. It's possible our time overruns scheme makes this less problematic, not completely sure.  
- We've fairly slow wasm build times, but maybe risc-v improves this.
 
We'd always introduce unpredictability into resource usage this way, meaning this harms performance somewhat.  At the same time it's unclear what this really gains us.

text/0032-minimal-relay.md
```diff
@@ -0,0 +1,190 @@
...
+- Balances
+- Staking
+	- Staking
+	- Election Provider
+	- Bags List
+	- NIS
+	- Nomination Pools
+	- Fast Unstake
```
burdges, line 50: We could "fork" the balances chain into an elections chain each era, which then brings heavier blocks.

We'll need to simplify the slashing system first I think, but even then lashing maybe presents challenges.

xlc:
I think it is best to do the simple things first: migrate identity and governance, and leave the tricky parts in a future RFC. Otherwise we won't be able to accept this RPC until all the hard questions are unanswered and it will unnecessary blocking the straightforward actions.

text/0032-minimal-relay.md
```diff
@@ -0,0 +1,190 @@
...
+
+The Relay Chain, in many cases, will still need to interact with these subsystems, especially
+Staking and Governance. These subsystems will require making some APIs available either via
+dispatchable calls accessible to XCM `Transact` or possibly XCM `Instruction`s in future versions.
+
+For example, Staking provides a pallet-API to register points (e.g. for block production) and
+offences (e.g. equivocation). With Staking in a system chain, that chain would need to allow the
+Relay Chain to update validator points periodically so that it can correctly calculate rewards.
+
+A pub-sub protocol may also lend itself to these types of interactions.
```
burdges, line 83: In cases, these XCM messages could easily be as expensive as, or more expensive than, the underlying operations, no?  It's plausible points could be moved almost completely off the relay chain too though.

It's tricky to do validator disabling properly via messages.

text/0032-minimal-relay.md
burdges: We'd ideally do XCMP or similar first because those messages being kept off the relay chain maybe more important than moving some tooling off.

text/0032-minimal-relay.md
shawntabrizi: Isn't availability bandwidth the same limit which leads to a maximum 5 MB PoV?

Or what would be the expectations on nodes to have and retain data here?

The common use cases for on-relay-chain pre-images for the PoV to me are:

- large runtime upgrades
- commonly used smart contracts

Especially in the case of immutable smart contracts, seems very sensible to have them stored on on the relay chain directly.

text/0032-minimal-relay.md
joepetrowski: This is a reasonable idea. The parachain is for Identity to get it off the Relay Chain. Of course with more advanced Coretime scheduling it's probably reasonable to have slower scheduling.

The reason to keep them separate for now is that Kusama is a nice dress rehearsal for Polkadot. But I have nothing against a second migration that deprecates the Kusama chain and does all identities through Polkadot's system.

text/0032-minimal-relay.md
joepetrowski: Fair enough, CoreJam also makes use of a Storage pallet that looks similar. I will remove this from the list of candidates.

text/0032-minimal-relay.md
joepetrowski: The current plan is that Staking have a Balances pallet.

joepetrowski:
> I think it is best to do the simple things first: migrate identity and governance, and leave the tricky parts in a future RFC. Otherwise we won't be able to accept this RPC until all the hard questions are unanswered and it will unnecessary blocking the straightforward actions.

The goal of this RFC is not to answer all the "hard questions" and implementation details. It is to state a direction and objective: Identity, Staking, and Governance off the Relay Chain. Of course, for each one there will be different "hard questions" and implementation challenges. Part of the process will be figuring them out, but it's better for parachain and UI developers to know now that the architecture of this state/logic that they interact with is being rebuilt.

text/0032-minimal-relay.md
joepetrowski: Sure, it's possible. Of course we need to figure out of VMP or pub-sub is better in each case.

xlc:
I guess we can signal the intention to move those features to system parachain. But without more feasibility study, we cannot say for sure we are going to move staking to a parachain. There is a good chance that it cannot happen without some major refactorings and I will say it is too early to make such decision at this stage.

text/0032-minimal-relay.md
joepetrowski: This is quite simple and I've already written a quick implementation.

1. Add a function to `reap_identity` that will:
    1. Unreserve the deposit;
    2. Teleport enough (about 1/100th) to the parachain;
    3. Call a `poke_deposit` function that will reserve the deposit on the parachain.

text/0032-minimal-relay.md
joepetrowski: I can add it, but this has always been part of building on Polkadot. Applications span multiple chains. That's one of the primary motivations for having shared security. And these dApps/wallets should be instantiating light clients anyway, not connecting to RPC providers.

text/0032-minimal-relay.md
joepetrowski: As you've said, a lot of the use of Identity is for reading. While there is downtime, it is only to updating identities and providing judgements. The service still exists for rendering identity info. We can try to coordinate upgrades/merging PRs, but IMO the complexity of a zero-downtime migration outweighs the benefits.

joepetrowski:
> There is a good chance that it cannot happen without some major refactorings and I will say it is too early to make such decision at this stage.

Well this is Polkadot eating its own dog food. Yes it might (probably will) take some big refactorings. But we should not say, "it's hard to do X on a parachain but we're special so let's just put it on the Relay", and then expect others to do complicated things in parachains. If it requires refactorings, let's do them.

text/0032-minimal-relay.md
joepetrowski: > I don't think we need identity as its own chain. Then we have the question of how to determine if a feature should lives on its own parachain or coexisting with another parachain.

Also, deposits are much lower on parachains. Current identity with 21 DOT is very expensive for a lot of people. Having it on a parachain makes it much more accessible.

text/0032-minimal-relay.md
burdges: I suppose the staking chain only discusses validators with the relay chain, never nominators, so maybe points could be done on the relay chain.

text/0032-minimal-relay.md
```diff
@@ -0,0 +1,196 @@
...
+provides these common guarantees by executing the state transitions on a strict subset (a backing
+group) of the Relay Chain's validator set.
+
+However, state transitions on the Relay Chain need to be executed by _all_ validators. If any of
+those state transitions can occur on parachains, then the resources of the complement of a single
+backing group could be used to offer more cores. As in, they could be offering more coretime (a.k.a.
+blockspace) to the network.
+
+By minimising state transition logic on the Relay Chain by migrating it into "system chains" -- a
+set of parachains that, with the Relay Chain, make up the Polkadot protocol -- the Polkadot
```
rphmeier, line 28: We should avoid mixing the general desire to take functionality off of the relay chain with an implementation which gives each function its own chain. We should consider offloading each of these components onto a single chain (staking + balances + identity) for better synchronous composability between them. 

Or to broaden the scope, consider other combinations of system chains which are already deployed.

The underlying goals are to maximize secure blockspace and effectively use blockspace. Anything which requires less than a single full core will have to either use coretime less frequently (poor system UX) or will use the core constantly but inefficiently (resource misallocation).

text/0032-minimal-relay.md
```diff
@@ -0,0 +1,196 @@
...
+## Motivation
+
+Polkadot's scaling approach allows many distinct state machines (known generally as parachains) to
+operate with common guarantees about the validity and security of their state transitions. Polkadot
+provides these common guarantees by executing the state transitions on a strict subset (a backing
+group) of the Relay Chain's validator set.
+
+However, state transitions on the Relay Chain need to be executed by _all_ validators. If any of
+those state transitions can occur on parachains, then the resources of the complement of a single
+backing group could be used to offer more cores. As in, they could be offering more coretime (a.k.a.
```
rphmeier, line 24: I have seen this argument crop up but I will comment to lay out the counterarguments:
  1. This presumes that the resources spent on including user transactions in blocks _can_ necessarily be spent on adding more cores.
  2. If (1) holds, this presumes that work done on cores will necessarily be more valuable than work done directly on the relay chain

re: (1) Current configurations specify resource requirements for cores (in data / compute) which are discrete and this implies that there is a high likelihood of some leftover resources which are sufficient to execute user transaction but insufficient to add another core. Furthermore, the actual data burden on the network can be estimated by the sizes of PoVs placed onto cores in recent blocks.
re: (2) Synchronous composability is often more valuable than asynchronous

text/0032-minimal-relay.md
burdges: Ideally, we place balances + staking + governance together on one parachain.  We'd still handle era points on the relay chain perhaps, but not dots.  We should figure out if this is possible though, aka how governance controls the relay chain, and how era points and slashing work.

Accumulate being quite flexible and permissionless is one of the two new ideas in corejam.  Accumulate becomes much safer if accumulate cannot perform logic based upon dot balances, but only based upon authorization conditions set by the parachains themselves. 

I'd envision nominations using a chain-forked pattern whatever we do, i.e. the staking chain spins off an nomination chain, to which validators add their opinions.  This requires running parachain blocks longer.

The other of the two new ideas in corejam is to have blocks run a long time by saving their memory state.  Although nice, we think this cannot really help with state migrations or npos, due to how memory accesses work.  We can likely use other techniques to do longer running npos computations on parachains though, like simply allowing them to run longer.

text/0032-minimal-relay.md
burdges: At some level, I do think cores could usefully consume otherwise spent on user transactions.  We do not necessarily have a straight line there, but it's still all bandwidth, cpu time, and disk space.  The question is more:  How much resources really?

I think the synchronous vs asynchronous dichotomy winds up being largely illusory here.  We do know some computations parallelize well and some badly, ala npos, but that's not really your dichotomy.  All of web2 is built upon sharded data models aka what you call asynchronous.  It's true that sharded data models require more thought, with this RFC being one example, but they're so dominant in web2 that they're probably not blocking adoption much.

gavofyork:
> I think it is best to do the simple things first: migrate identity and governance, and leave the tricky parts in a future RFC. Otherwise we won't be able to accept this RPC until all the hard questions are unanswered and it will unnecessary blocking the straightforward actions.

The point of the RFC is to get everyone on the same page regarding the future of the Relay-chain. This is nothing new and I gave several presentations mentioning all of this in 2019.

gavofyork:
I really cannot see any reasonable argument to intentionally put work on an obvious bottleneck point of the network which can be moved into a non-bottleneck point.

Synchronous composability specifically on the Relay-chain, whose entire point under CoreJam is to join async compute results and accumulate state machine transitions, does not sound sensible to me at all.

I've said it before and I'll say it again: The Relay-chain should do one thing and do it well. That means being the head of a secure decentralised multicore computer. Sticking an EVM smart contract in there and/or continuing to allow end-user-application-level transactions and general interaction goes directly against this mantra.

text/0032-minimal-relay.md
gavofyork: I really cannot see any reasonable argument to intentionally put work on an obvious bottleneck point of the network which can be moved into a non-bottleneck point.

Synchronous composability specifically on the Relay-chain, whose entire point under CoreJam is to join async compute results and accumulate state machine transitions, does not sound sensible to me at all.

I've said it before and I'll say it again: The Relay-chain should do one thing and do it well. That means being the head of a secure decentralised multicore computer. Sticking an EVM smart contract in there and/or continuing to allow end-user-application-level transactions and general interaction goes directly against this mantra.

text/0032-minimal-relay.md
rphmeier: Anything that requires non-uniform work package execution times/weights seems impractical. It should be possible to extend computations over multiple blocks, by paging in the relevant data needed for each step of the algorithm at any point. 

Yes, spinning off child chains/tasks for heavy computation such as elections or governance vote-tallying makes a lot of sense, especially since parachain execution is already stateless. The System product benefits a lot from having as many user-exposed functionalities as possible living in the same chain, especially when there is no clear impetus to scale horizontally.

@joepetrowski

joepetrowski: Once two features are in the same chain, users and applications get very accustomed to their being in the same chain, which makes it difficult to split them into multiple chains. I understand that with today's usage, many of these subsystems could all go into a single system parachain. However, as usage and requirements increase (e.g. several thousand validators), it will hit a bottleneck and need to be split.

IMO it makes more sense to keep these subsystems in separate chains but take advantage of core scheduling to handle resource allocation. We could use one core to process staking, governance, and identity in round-robin fashion (perhaps prioritizing staking at certain times like session changes). You mention degraded system UX but we're still talking 18 second instead of 6 second block times. When a single core no longer meets the execution needs of a subsystem, it's much easier and more agile to allocate a dedicated core to that subsystem than it is to launch a new chain and migrate all the data/logic to it (plus getting parachains to re-address their XCM programs, applications to target new chains, etc.).

RE spinning off child chains for heavy computations, yes this is reasonable but I think it's an optimization. The first priority is to get this work off the Relay Chain. Then we can optimize with things like child chains.

text/0032-minimal-relay.md
burdges: We can mark this thread resolved I think.  We know the total data flow between the relay chain and the staking+governance chain should not be too terribly large.  We'll upload PVFs to the relay chain I think, so maybe they'll need some pre-authorization from the governance chain, just one of a number of problems to solve, but not too difficult overall.

sourabhniyogi:
> > I think it is best to do the simple things first: migrate identity and governance, and leave the tricky parts in a future RFC. Otherwise we won't be able to accept this RPC until all the hard questions are unanswered and it will unnecessary blocking the straightforward actions.
> 
> The point of the RFC is to get everyone on the same page regarding the future of the Relay-chain. This is nothing new and I gave several presentations mentioning all of this in 2019.

Here is some [data](https://docs.google.com/spreadsheets/d/1Hwt_z2_WrfrswbDpho4LxghctSsnTRbl0jhHhWPhKQk/edit#gid=1210313544
) from Q3 2023 on users and weights that might help you prioritize what should get moved off the relay chain and in what order:
```
SELECT  call_section, count(distinct 	
signer_pub_key) activeAddresses, round(sum(weight)/1000000000000, 1) weight, count(*) numCalls, count(distinct extrinsic_id) numExtrinsic FROM `bigquery-public-data.crypto_polkadot.calls0` WHERE TIMESTAMP_TRUNC(block_time, DAY) >= TIMESTAMP("2023-07-01") and TIMESTAMP_TRUNC(block_time, DAY) <= TIMESTAMP("2023-09-30")  group by call_section having call_section not in ("utility", "proxy") order by weight desc, activeAddresses desc LIMIT 100
```

https://docs.google.com/spreadsheets/d/1Hwt_z2_WrfrswbDpho4LxghctSsnTRbl0jhHhWPhKQk/edit#gid=1210313544

The above says very clearly: staking and balances!   Everything else (identity, governance) looks like premature scaling and spinning up chains for the hell of it, to disturbing levels.

Zeroing in on those 2, there are many ways you could proceed:
(A) Staking chain
(B) Balances chain (AssetHub)
(C) 1 staking chain, 1 balances chain
(D) 2 chains both doing staking + balances
(E) 4 chains: 2 doing staking and 2 doing balances

To improve XCM's async patterns, I would pick (A) or (B).  This is sorely needed, and I would try to make [this guy](https://forum.polkadot.network/t/developer-experience-must-be-our-1-priority/3957/20) happy.  Given a forced choice, pick (A) over (B) -- it has fewer and more sophisticated stakeholders
To focus on substantive innovation on CoreJam/Coreplay's **async** innovations ALONE I would go straight to (C)
To focus on substantive innovation on CoreJam/Coreplay's **sync+async** innovations I would go straight to (D)
(E) is a hidden IQ test.  Think about why.

I don't think Polkadot engineering has gone through the "heterogenous app chains => homogenous shards" conceptual change required to even choose (D).   This an important prereq.  

I suspect from many years of engineers here being trained on "heterogenous app chains" and specifically against "homogenous shards", a lot of you will reflexively think (C)=>(E) is a great idea, but actually (E) is a test of whether you have any regard for locality considerations, reducing unnecessary messaging, and to put it bluntly, just want to spin up new chains (governance! identity!), like, for the hell of it.  Its seems pretty clear the 1.0 heuristic of 

**Every use case should get its own chain!**

has to be discarded in favor of something more intelligent in 2.0, recognizing that

**1.0 chains as a primitive were not efficient! RIP Chains!**

and 

**in 2.0, we do more resource allocation enabled by CRJA: Actors, Work Classes, Slabs, ... in ALL use cases**  

The kind of solution hinted at [here](https://github.com/polkadot-fellows/RFCs/blob/gav-coreplay/text/coreplay.md#scaling-a-common-use-case) should be extended to the _visibly tiny_ (in terms of weight) use cases of governance and identity -- which, would NOT result in a whole governance and identity chains but instead address locality in sync/async composability much more deliberately in a new Actor model (Coreplay), or in Work Classes (CoreJam) and end up using resources efficiently.

joepetrowski:
> The above says very clearly: staking and balances! Everything else (identity, governance) looks like premature scaling and spinning up chains for the hell of it, to disturbing levels.

Thanks for collecting the info. However, it only seems to look at user-dispatch transactions and not count things like on-initialize? Governance for example occasionally does some heavier things like vote tallying. So all the weight of a block doesn't necessarily come from user-interaction.

It may look like premature scaling, but like I said [here](https://github.com/polkadot-fellows/RFCs/pull/32#discussion_r1347949939), core allocation is much more agile than state/logic allocation. These subsystems can be in separate chains but still share a minimal number of cores. When they surge, core allocation can happen dynamically. And like Jeff said, we can even spin off child chains for ephemeral heavy tasks.

The RFC also mentions very little about "spinning up chains for the hell of it", and is focused on _getting functionality **off** the Relay Chain_. In fact, rather than a "new" governance chain, I'd expect to just migrate governance into Collectives. This also comes back to your data about current weight consumption and future plans: We expect many more collectives and sub-treasuries.Therefore, we want the agility of being able to move to a dedicated core if need be, but can share a core with other system chains when appropriate. Just like a staking system that can support thousands of validators, better to functionally isolate these components and do proper resource allocation with Agile Coretime than to limit the ability of one subsystem. 

And to get to balances, well we need staking, governance, and identity off the Relay Chain first. Again, balances to Asset Hub, governance to Collectives. No chains "for the hell of it".

> in 2.0, we do more resource allocation enabled by CRJA: Actors, Work Classes, Slabs, ... in ALL use cases
>
> The kind of solution hinted at [here](https://github.com/polkadot-fellows/RFCs/blob/gav-coreplay/text/coreplay.md#scaling-a-common-use-case) should be extended to the visibly tiny (in terms of weight) use cases of governance and identity -- which would NOT result in a whole governance and identity chains -- but instead address locality in sync/async composability much more deliberately in a new Actor model (Coreplay), or in Work Classes (CoreJam) and end up using resources efficiently. This is the future.

Yup, and to make CoreJam happen, we need to take these functions off the Relay Chain.

gavofyork:
Furthermore, the Relay-chain's runtime is the closest Polkadot gets to a single point of failure. We can't avoid upgrading it in general, but the simpler and more minimal we can keep it the better, especially regarding pallet code which necessarily executes with system-level privileges.

gavofyork:
Agglomerating the pallets of galances, governance, identity and staking into one "blob" chain separate from Assets Hub, Bridges Hub and Collectives seems to needlessly presume that merely because they happened to begin life on the same chain (owing simply to the fact their functions were needed prior to the existence of the parachains logic) they should remain so.

Migrating production pallets between chains is a difficult and time-consuming job at the best of times. Placing them all on one chain presupposes that not a single one of them nor any yet-to-be-devised functional-companion code which does actually need to be within the same environment will ever grow beyond the needs of a single chain. This appears unfounded to me. Unless there is a clear need for pallets to live together, I would err on the side of future-proofing.

sourabhniyogi:
> However, it only seems to look at user-dispatch transactions and not count things like on-initialize?

You're right -- I added in an [extrinsics sheet](https://docs.google.com/spreadsheets/d/1Hwt_z2_WrfrswbDpho4LxghctSsnTRbl0jhHhWPhKQk/edit#gid=2135602009) to get at the on-initialize where beloved `paraInherent` totally dominates over everything else [even timestamp does]. This might inform you about just how little you might be gaining through this, using weight as a quick guide (I understand its not enough).  Intuition as a total outsider is to future proof not by sticking to 1.0 heterogenous system chains but move to 2.0 homogenous shards in CRJA.  

Specifically, instead of moving the pallets of balances, governance, identity and staking into one "blob" chain + AssetHub, just jump to straight to 2 => 4 => 8 => ... (or 2 => 3 => 4) homogenous shards coordinating as many use cases as you have to, and do it all at once with CRJA.  This is "2.0 FTW".  A good future proof solution would address mitosis-like or linear growth and addresses how new shards get added and doesn't have engineers doing scheduled migrations as new shards come online.    

I'm not a superengineer doing substrate heroic migrations here, but if were, I'd sure like to know step 2 will work in detail (how homogenous shards ala CRJA/CP on 2 chains) and step 3 at a high level (how more shards come on line in a future proof way) before doing step 1 (move {X, Y, ...} off the relay chain).  Doing step 1 (moving OFF the relaychain) and having ANOTHER migration later looks quite short sighted!  If your end goal is 1 chain do to X and 1 chain to do Y, even non-substrate engineers can see its a temporary solution with an IOU baked into the number "1".

rphmeier:
> Placing them all on one chain presupposes that not a single one of them nor any yet-to-be-devised functional-companion code which does actually need to be within the same environment will ever grow beyond the needs of a single chain
> I would err on the side of future-proofing

Starting from the practical consideration that cross-chain UX is quite poor at the moment, I would prefer to err on the side of ergonomics, not presuppose that all of these components together will require substantially more resources than a single core can provide. The ergonomics are poor.

This is both empirically true at the moment and there is little indication that it would change in the near future

xlc:
There are overheads of a new chain in pretty much all the directions: dev, ops, UX, performance, governance, etc. 
We need to take considerations of all those overheads in short term and long term when deciding a new system parachain.

text/0032-minimal-relay.md
arrudagates: What if we use the current collectives chain as a hub for everything that's social/offchain related?

arrudagates:
I believe a good compromise can be made here, we can meet in the middle of ergonomics and future proofing by bundling functionality that fits within a narrow category, staking + balances, identity + collectives, for example.

text/0032-minimal-relay.md
rphmeier: > You mention degraded system UX but we're still talking 18 second instead of 6 second block times

With the amount of traffic that an identity chain currently handles, it'd take weeks to make a full block. Occupying a full core even 1/3 of the time is inefficient, UX challenges for users coordinating across many chains notwithstanding.

I see the case for being inefficient with coretime in the face of a coretime surplus, though I expect that the pressures under load would point back to consolidation and the practical experience for end-users of navigating many chains tips it in favor of consolidation now for me.

Full support of moving these pallets off of the relay chain.

sourabhniyogi:
> I believe a good compromise can be made here, we can meet in the middle of ergonomics and future proofing by bundling functionality that fits within a narrow category, staking + balances, identity + collectives, for example.

Here is a [neat vs scruffy](https://en.wikipedia.org/wiki/Neats_and_scruffies) way to compromise:
1. Kusama be [Team Scruffy](https://twitter.com/rphmeier/status/1707098462347514328/photo/1), roughly pursuing UX/ergonomic + experimentation first (supporting hybrid chains, smart contracts on the relay chain) and not pursuing #32 (or any migrations) unless absolutely necessary for CoreJam/CorePlay
2. Polkadot be "Team Neat", roughly pursuing scalability-at-all-costs (contra any ergonomic), pursuing #32 at all costs

The belief of Team Neat is that Kusama will die a horrible death at the hands of long-term scalability.
The belief of Team Scruffy is that Polkadot will die a horrible death at the hands of long-term ergonomic.

Competition between the two efforts will be done healthily, however, like within 2 branches of the same car company appealing to different market segments (Tesla S vs Tesla Model 3) who want to see the car company (Tesla) succeed and recognize their customers (developers, users) actually want different kinds of things.

Then, parachain BCC customers and CorePlay/CoreJam ICC customers of DOT+KSM CoreTime will report back on the amazing scalability of Polkadot (with more cores, more efficiency, higher security, blob chains splitting into ever more blob chains, etc.) and the amazing usability of Kusama (with fewer or maybe even NO system chains).  

Wdyt?

joepetrowski:
I will stay out of the "smart contracts / EVM on Relay Chain" discussion, at least in this thread. But I'm generally OK with the "Polkadot goes neat" approach. However, I want to highlight that the differences in this thread are rather small now and there is generally a lot of agreement. I think it all comes down to Identity:

- :heavy_check_mark: Most agree that Staking, Governance, Identity, and Balances should move **off** the Relay Chain (details on where they go aside).
- :heavy_check_mark: Staking is heavily used and we have plans to grow the validator set well beyond 1,000. It probably deserves its own chain. Here is where I disagree with @sourabhniyogi 's Scruffy Kusama approach: We'll want to make sure the logic continues to work properly in a parachain with very high validator counts. Having a Kusama Staking chain seems really valuable to Polkadot.
- :heavy_check_mark: Governance will just go to Collectives. Chain already exists. Seems uncontroversial?
- :heavy_check_mark: Balances move to Asset Hub. Again, already exists.
- :thinking: Identity, own chain or with governance? Also, with @xlc 's suggestion, we could just remove Identity from Kusama and use the data from Polkadot's system.

So we are talking about at least one new parachain (Staking), maybe two (Identity). My opinion stands that it'd be better to functionally isolate it and just schedule it rarely. Yes it's an ergonomic cost, but having 3 system paras instead of 4 doesn't obviate the need for better tooling and UX around multi-chain applications. And if these 4 can still be run on the same core most of the time (even with Identity say once per hour), then it's still using only one core for system operations.

sourabhniyogi:
> I will stay out of the "smart contracts / EVM on Relay Chain" discussion, at least in this thread. But I'm generally OK with the "Polkadot goes neat" approach. However, I want to highlight that the differences in this thread are rather small now and there is generally a lot of agreement. I think it all comes down to Identity:

For "Polkadot (but not Kusama) goes neat", could we use some counting of 7s 6s 5s 4s and 3s in the ["Process"](https://github.com/polkadot-fellows/RFCs#process_) to decide a bunch of Y/N questions according to some algorithm that uses rank?  At the end of the day, how are the results of this supposed to relate to governance's tool of referendums?   Shouldn't a proposal be put in from to the community?  

I reread the [Manifesto](https://github.com/polkadot-fellows/manifesto/blob/main/manifesto.pdf) and without a process, the general agreement looks like "1.6 The Papal Model".  Its a good time to test out the program.

rphmeier:
> Staking is heavily used and we have plans to grow the validator set well beyond 1,000. It probably deserves its own chain

The only heavy things here are elections and rewards. If there is no decoupling of staking intentions / nominations from elections and rewards, I agree it needs its own chain. There are definitely merits to having staking intentions / nominations / reward destinations localized with other functionality.

Governance going to collectives and balances going to assethub makes sense given that these chains are already live and it adds no additional burden.

Identity being added to governance/collectives would make sense here.

However, none of these details are actually reflected in the RFC text - RFCs require sufficient detail for implementation.

text/0032-minimal-relay.md
```diff
@@ -0,0 +1,196 @@
...
+- Core protocol and XCM format developers;
+- Tooling, block explorer, and UI developers.
+
+## Explanation
+
+The following pallets and subsystems are good candidates to migrate from the Relay Chain:
+
+- Identity
+- Balances
+- Staking
```
Ank4n, line 43: Small note that the Staking and EPM pallet would need significant refactor and may be almost a complete rewrite in case of EPM before being ready for Staking Chain. As it stands currently, they are not very PoV friendly.

joepetrowski:
> The only heavy things here are elections and rewards. If there is no decoupling of staking intentions / nominations from elections and rewards, I agree it needs its own chain. There are definitely merits to having staking intentions / nominations / reward destinations localized with other functionality.

Sure, there are tradeoffs to everything. For systems that consume high amounts of blockspace, it's better to be agile with Coretime scheduling so that critical functionality (especially elections and slashing) can execute when needed. Of course there are merits to having things localized, every engineering decision has tradeoffs. There are also merits to it having its own chain.

> However, none of these details are actually reflected in the RFC text - RFCs require sufficient detail for implementation.

I will add more detail as a result of the discussion. But like Gav said [here](https://github.com/polkadot-fellows/RFCs/pull/32#issuecomment-1748988214), this is to get everyone on the same page regarding the _Relay Chain_. I have a rather concrete and partially implemented plan for Identity migration (at least to its own chain), it will take some mods if this goes to the (already-running) Governance chain. @muharem is planning an RFC for the plan to move governance off the Relay Chain. And @gpestana / @Ank4n have done quite a lot of work on Staking plans, which I'm sure they can convert to an RFC once ready.

> could we use some counting of 7s 6s 5s 4s and 3s in the ["Process"](https://github.com/polkadot-fellows/RFCs#process) to decide a bunch of Y/N questions on what should go into various blob chains according to some algorithm that uses rank?

Isn't that the approval referendum of an RFC / whitelisted runtime?

> Shouldn't a proposal be put in front of the community for something as important as UX/ergonomics?

Yes, a runtime upgrade / new parachain proposal. And I do believe UX/ergonomics are important, when arguing for a separate chain for Identity I am not arguing _against_ UX. My point is that the marginal cost of another chain is small. Whether we have 5 chains or 6 (AssetHub, BridgeHub, Coretime, Staking, Governance, (Identity)), we still need tools that make multi-chain application development ergonomic. Coretime scheduling should manage appropriate resource allocation because that's what it's designed to do.

xlc:
> My point is that the marginal cost of another chain is small.
I don't believe that's true at this stage. We want to make the cost to be small but unless someone make a proposal explain on how we can make it small in future, I won't accept it as an argument.

gavofyork:
> > My point is that the marginal cost of another chain is small.
> 
> I don't believe that's true at this stage. We want to make the cost to be small but unless someone make a proposal explain on how we can make it small in future, I won't accept it as an argument.

Care to expand on why you believe exactly having 4 system chains rather than 3 has such a large additional cost?

xlc:
None of us can just state something without provide explanations so here are some areas to check for the cost of operate a new chain:

- Initial development time. That including copy template, ensure deps are aligned with other chains, update the name and configurations, generate genesis, testing to ensure it actually works. Maybe someone can check the time used to setup the collective parachain?
- Maintenance time. This is relatively small as it is mostly duplicating the changes.
- Integration tests & e2e tests & manual tests. Not terribly bad but does require more work. Also keep in mind the number of XCM tests can be `O(n^2)` as we need to ensure every system parachains can interact with each other via XCM. At least for the one with balances pallet.
- Audit. 
- Infra. RPC nodes, Collator nodes, monitoring etc. We can't share the nodes (yet) and they are not cheap to run.
- Governance runtime upgrade overhead. The overhead can be reduced with batch call but still, we don't yet have the ability to run some script and just tell be the upgrade runtime proposal is good.
- dApps overhead. One more chain to connect. More set of RPC nodes to deal with OR one more chain spec in light client. The chain spec files are big because they include wasm and use hex format for it. Add 1mb payload to a web app is always bad UX.
- Devs overhead. We simply don't have a good library to deal with multiple APIs. This overhead can be considered none once we have it but we don't have it.

joepetrowski:
> Initial development time. That including copy template, ensure deps are aligned with other chains, update the name and configurations, generate genesis, testing to ensure it actually works. Maybe someone can check the time used to setup the collective parachain?

I already have a branch largely configured, so close to zero. Anyway, the initial configuration takes a few hours. For testing, we have infrastructure ready to get this started on Rococo and Westend.

> Maintenance time. This is relatively small as it is mostly duplicating the changes.

Yup, very little.

> Integration tests & e2e tests & manual tests. Not terribly bad but does require more work. Also keep in mind the number of XCM tests can be `O(n^2)` as we need to ensure every system parachains can interact with each other via XCM. At least for the one with balances pallet.

Testing is not a resource-constrained environment. Testing n^2 cases is not a big deal, still takes less time than running benchmarks. And they are all automated.

> Audit

Very little, it's a pretty standard system parachain configuration.

> Infra. RPC nodes, Collator nodes, monitoring etc. We can't share the nodes (yet) and they are not cheap to run.

Collator nodes are not expensive. I think the Infra bounty for them is something like 100-200 USD / month / collator.

> Governance runtime upgrade overhead. The overhead can be reduced with batch call but still, we don't yet have the ability to run some script and just tell me the upgrade runtime proposal is good.

Yes we do. This is easy. Few lines of code in the script. Almost zero overhead.

> dApps overhead. One more chain to connect. More set of RPC nodes to deal with OR one more chain spec in light client. The chain spec files are big because they include wasm and use hex format for it. Add 1mb payload to a web app is always bad UX.

The goal of Polkadot (and the reason for shared security) was precisely that applications could span multiple chains. Apps need a solution for this and it has nothing to do with system chains.

> Devs overhead. We simply don't have a good library to deal with multiple APIs. This overhead can be considered none once we have it but we don't have it.

I've mentioned this several times on the forum and I think there's a W3F Grants RFP for it (@Noc2 ?). Like the point above, yes it's needed but it's needed in general, for all 50+ paras on Polkadot, and really has not much to do with a single system chain. Hopefully this is more incentive to make it.

text/0032-minimal-relay.md
```diff
@@ -0,0 +1,196 @@
...
+Note: The Auctions and Crowdloan pallets will be replaced by Coretime, its system chain and
+interface described in RFC-1 and RFC-5, respectively.
+
+### Migrations
+
+Some subsystems are simpler to move than others. For example, migrating Identity can be done by
+simply preventing state changes in the Relay Chain, using the Identity-related state as the genesis
+for a new chain, and launching that new chain with the genesis and logic (pallet) needed.
+
+Other subsystems cannot experience any downtime like this because they are essential to the
```
kianenigma, line 64: Co-existing will be quite tricky for staking, and we would likely have to do a one-off migration. The migration is possible to happen within an era, but even if not, having one extended "super era" is not too bad. 

With a one-off migration, all the staking data is moved to the system chain. Some historical staking data will be kept on the RC for retroactive rewards and slashes, and will be removed after a maximum of 84 eras.

gavofyork: Is this due to mutual exclusivity of freezing on the Relay-chain and the Staking chain?

Ank4n: We discussed an alternative migration strategy to staking chain by running two parallel staking system for a while.
Roughly the way it would work is:

Staking system in the Relay chain works independently of Staking Chain. Relay Chain has a configuration that decides how many validators are provided from staking chain, the rest elected locally in the Relay Chain. We can then slowly scale up this number to 100% Staking Chain.

Validators who want to validate via Staking Chain, would need to unbond, teleport assets to Staking Chain, and set their intent to validate again (we could potentially improve this by allowing teleport of staked assets without unlocking). In the beginning it may be easy to get into the active validator set via Staking chain until we reach some equilibrium eventually getting to similar economic security as Relay Chain. Once we go to 100% from Staking Chain, at some point we can kick all validators/nominators who did not migrate.

The main opposition to this approach were

- Total stake for active validators (economic security) may drop while in transition phase.
- Validators have to move and re-acquire the approval score they gained on the Relay Chain.
- It might be harder to move all nominators voluntarily to Staking Chain.

We (weakly) reached to the consensus that one-off migration might be simpler. We will need to migrate the staking
ledgers and the staked balance to the Staking Chain and do this in one era (which can be an extended era).
The era history can stay in Relay Chain allowing validators to claim historical rewards. Once 84 eras (history depth)
have passed, we purge the historical data and clean up staking code in relay chain.

Keen to hear if you have some thoughts/feedbacks on both these approaches (this probably deserves its own RFC but we do have an [open issue](https://github.com/paritytech/polkadot-sdk/issues/491) where we have some more notes from our discussions so far).

Tbaut: This migration will require a lot of upfront communication. The less users and validators have to do, the better. I believe that a one-off migration would be much easier to handle on the wallets side, and communication side. Wallets and partners (think CEXs etc) should be prepared, and when the block hits, they'll be able to switch to displaying/using the staking chain. With a gradual move however, I can see a period when users don't know if they are on the RC or the SC, UIs would have to show and handle both, which would likely be very confusing.

gavofyork: I wholeheartedly agree with a one-off migration approach, in both this case and as a general pattern for moving systems between chains or performing a large complex upgrade.

burdges: We should obviously not require that users change wallet software to make new accounts, transfer, stake, etc.  That's a big no no.

We need domain separation material in signatures of course, but the exact semantics of that domain separation material can evolve if required.  If the new chain is *the* staking chain, then it can have a "virtual" genesis hash of the original relay chain or whatever.  We donno how to have multiple staking chains anyways.

I guess governance could be more nuanced. 

As for migration, we could've the staking chain "fork" off from the relay chain at block height x, which freezes transactions on the relay chain, then the staking chain runs a migration pallet, and finally the staking chain unfreezes its own transactions.  Anyways a chain "fork" operation should not be expensive and should be useful elsewhere.

We'll want a bunch of O(1) drops here, so those should be implemented first.  We might clean up the state a bit first too, like fixing the slashing spans system, and droping that old state.  As a backup, we proved a limited recovery sudo which kills the staking chain and unfreezes the relay chain accounts.

Ank4n: > As for migration, we could've the staking chain "fork" off from the relay chain at block height x, which freezes transactions on the relay chain, then the staking chain runs a migration pallet, and finally the staking chain unfreezes its own transactions. Anyways a chain "fork" operation should not be expensive and should be useful elsewhere.

What we really need to move is staking pallet state and part of user balance that is staked. Unstaked balances should eventually end up in AssetHub once we get rid of balances from relay chain (and should be part of separate migration I think). We could do a fork and then get rid of things that we don't need in Staking Chain and translate what we need (I guess thats what your migration pallet would do). Relay chain can only freeze staking related transactions and kill it once staking chain is operational. Does that make sense?

@kianenigma had the idea to prepare data that we need to migrate, calculate the state root and securely send it to staking chain (root track or xcm), and then have a bot that fills the storage on staking chain through permissionless calls. Wdyt? May be forking off is simpler?

joepetrowski:
https://collectives.polkassembly.io/member-referenda/28

sourabhniyogi:
@joepetrowski @gavofyork  I see that you added in a new section on Kusama, to practice migration.    I was hoping Kusama could be a place for CoreJam + Coreplay experimentation with fewer system chains, a place for Team Scruffy, but its more like Team Neat needs to abort Team Scruffy at birth!    I think your coercing #32 into Kusama will deal a blow to Polkadot/Kusama ecosystem, or at least damn Kusama to be nothing more than a "testnet" -- its too big of a bet on the ergonomics / UX all coming together too quickly.  

How can we have separate #32 referendums for Polkadot (for scalability) and for Kusama (for usability)?    I'd like to have a situation where Kusama can live to be MORE than a testnet, where developers have a nice "simple" place to work with the new 2.0 CoreJam + Coreplay programming model, including Availability patterns.

If having Kusama be more than a testnet is impossible (because Rococo + Chopsticks  "practice" migration is insufficient), I would like to see a THIRD production network, say the "CoreTime network", a production network (relay chain) dedicated to improving 2.0 CoreJam + Coreplay usability for new devs with lower complexity than its Big Brother Polkadot.   

Can you chart a course?

joepetrowski:
The section on Kusama was added insofar as to state how it can be useful in its capacity as canary net in achieving the goals set out in this RFC. Adding a vision and roadmap for Kusama development is way out of scope for this RFC.

When CoreJam comes, and it's probably approximately a year to have it in a "Kusama ready" state, I'm sure Kusama will again play a leading role on the frontier.

shawntabrizi:
I will just speak my thoughts here, hoping to come from a constructive place.

I spiritually agree with this proposal, and I am inclined just to vote AYE.

At the same time, I certainly know that a single document with < 300 lines can capture the underlying complexity, challenges, and decisions that will need to be made once this is actually underway.

So I kind of want to understand what we are voting for here. If the vote is to establish direction, then I am a strong AYE.

However, I guess I would also expect that different parts of the actual migration process will also manifest as RFCs to be approved and so on. Perhaps one system chain at a time, where even chains like Staking may take multiple RFCs and upgrades to actually get to a final envisioned state.

Is that an accurate picture of what this vote is about, or is there some leniency we are supposed to give on how exactly this is all accomplished?

burdges:
I'd say the vote says "kusama should eventually do this unless we hit real obstructions".  If we later discover it sucks on kusama despite our best efforts, then yeah we might do something slightly different, although more likely we tweak functionality.

The vote should _not_ say "we're going to push this through quickly".  In particular, we want whatever pattern of migrations happens here, like parachain forking, to be useful for customer parachains when they want to buy multiple slots, without using elastic scaling.  We should probably not do this until we can explain what we're doing to customer parachain teams.

joepetrowski:
@shawntabrizi yes this is to establish direction. Regarding each subsystem:

- I already have a very concrete plan for how to do Identity (it's the simplest anyway). I can already start a new RFC for that if that's the format Fellows want.
- @muharem is working on an RFC specifically for governance.
- @gpestana, @Ank4n, and @georgepisaltu are working on Staking and will have an RFC with concrete plans as they evolve.

I don't think it's entirely clear yet what should be an RFC and what not, though. Of course, changes to the core protocol should be in RFC format. A step-by-step migration plan for Identity? I'm not sure an RFC is more valuable than a detailed GitHub tracking issue.

sourabhniyogi:
> The section on Kusama was added insofar as to state how it can be useful in its capacity as canary net in achieving the goals set out in this RFC. Adding a vision and roadmap for Kusama development is way out of scope for this RFC.
> 
> When CoreJam comes, and it's probably approximately a year to have it in a "Kusama ready" state, I'm sure Kusama will again play a leading role on the frontier.

@joepetrowski  Got it!  I can see the level of depth of thought given to the ergonomic/UX considerations so voiced by @rphmeier and concerns raised by @xlc, my own RFC #33.   I closed #33 in favor of [this stand in of a roadmap](https://github.com/colorfulnotion/optimism/blob/develop/CORETIME.md) -- there is no path for Kusama worth developing.  Thank you for setting up the foundations so clearly.

xlc:
I share similar feeling with @shawntabrizi and that's why I haven't voted.

I can see this established a general direction for relaychain and system parachains work, which is indeed useful to have. However, there are not enough technical details to follow so we need additional RFCs to discuss the exact actions. And we are likely to not follow this RFC in case some unexpected obstacle were discovered in future.

Quoting myself from a comment from other proposal

https://github.com/polkadot-fellows/RFCs/pull/35#pullrequestreview-1670004354

> I see this is more of a feature request instead of RFC. There are no enough technical details for reviewers to evaluate the full impact of this proposal and for developers to implement if accepted.

shawntabrizi:
Perhaps we should have different kinds of RFCs like: `[Direction]`, `[Feature Request]`, `[Implementation]`.

In any case, since this is a "direction" based RFC, and I agree with the direction, I will vote AYE.

joepetrowski:
/rfc process 0xa29d08c3030bf527de0fadef3f658ea1e1d3200c4a7530038d693cbe97cab7f3

PR #33, opened 2023-10-03
CoreJam System Chain on EVM L2s with OP Stack

sourabhniyogi:
This proposal adapts the CoreJam architecture to EVM L2s, specifically utilizing OP Stack + Solidity instead of Polkadot/Substrate.  Almost all CoreJam concepts are retained, but CoreJam's interfaces are replaced with Solidity/EVM Contracts + OP Stack's Golang.


rphmeier:
With respect to (1) I believe this is out of scope of adding EVM contracts to Polkadot per-se (though is something I'd very much like to see). As long as Polkadot's DA is exposed, L2 stacks can be built anywhere on top of Polkadot.

As you mentioned in the RFC, I did indeed write a post advocating for every chain to have smart contracts. I think my arguments on synchronous composability still hold. I'd tentatively support having some very limited amount of Polkadot relay-chain blockspace allocated to smart contracts, though in the event that on-chain processing becomes the bottleneck for the number of cores, I'd advocate for the smart contract functionality to be outmoded by the number of cores.

gavofyork:
CoreJam already provides for the possibility of permissionless logic to be introduced for determination on the Relay-chain itself.

CoreJam, as it already stands, enables both (1) and (2) without needing to reassign "blockspace" in this way.

sourabhniyogi:
> CoreJam already provides for the possibility of permissionless logic to be introduced for determination on the Relay-chain itself.
> 
> CoreJam, as it already stands, enables both (1) and (2) without needing to reassign "blockspace" in this way.

I am interested in porting OP Stack to use Polkadot DA and need stronger design advice on how to do that in 2023 through mid 2024 as it will take 2-4 quarters to develop CoreJam.    

Given your clear resolve in your comment to #32  "Sticking an EVM smart contract in there [the Relay Chain] and/or continuing to allow end-user-application-level transactions and general interaction goes directly against this mantra [from 2019]" being _directly_ at odds with @rphmeier's hybrid chain / experimentation advice (which I'm trying to follow here =) ),  I suggest the following pragmatic strategy:
1. we use Kusama ALONE to develop non-Substrate EVM L2 using Kusama DA over 2-4 quarters, while CoreJam+Coreplay develops (surely, also on Kusama).     _Kusama was for experimentation, right?_  
2. you allow the COMMUNITY, in 2-4 quarters to decide whether it would be appropriate to extend (1) to Polkadot or not.  I would expect the Fellows, armed with a working and fully developed CoreJam architecture, to advise rather than unilaterally rule on this process.   _Is this not the function of why OpenGov was created?  With "Fellows"  not deciding the fate of everything without governance?_

sourabhniyogi:
@gavofyork @rphmeier I put this [Optimism on Kusama 2.0](https://github.com/colorfulnotion/optimism/blob/develop/README.md) proposal together, in a Team Neat Polkadot vs Team Scruffy Kusama way:

https://github.com/colorfulnotion/optimism/blob/develop/README.md

While I wish there were mature OpenGov and Fellowship protocols to decide these important issues, I think its ok if you founder types decide it.

sourabhniyogi:
Kusama has no path other than to be a canary network.

sourabhniyogi:
@gavofyork [@rphmeier] Rather than chart a course for Kusama 2.0 with CoreJam+EVM, I rewrote this to be a  a totally different approach to CoreJam.   In last month's call, you asked "is Polkadot just parachains, is CoreJam some other thing" and I think CoreJam's map-reduce is too just promising to be just on Polkadot/Substrate, done in Rust only, developed with Polkadot 1.0, and hope you can lead people out of "Polkadot is better than EVM/Smart Contracts/Solidity" to see things differently.   I think a new generation of students deserve it.

Lots of basic questions I had, having gone through the exercise:
* how EVM users should submit WorkItem going into work packages/services
* how to bring Instantaneous CoreTime Credits back, why you removed it in favor of Bulk CoreTime only
* how to map dynamic backing groups into a set of OP Stack nodes
* how to measure cores 
* how to measure DA throughput 
I can mostly go divergent from here on out and come up with my own answers to a lot of this, but it would be cool if CoreJam here is CoreJam there with your best advice on how that could happen.

Issue #19, opened 2023-08-10
Light clients and downloading block bodies


tomaka:
After having submitted a transaction, a light client generally wants to know when and where the transaction is included in the finalized chain. This is both for UI purposes and to know when to stop gossiping that transaction.

In order to know when a transaction is included, the light client needs to download the body of each block of the best chain and compare the transactions of the body with the one that it has submitted.

While this is not a problem right now, if block bodies become larger in the future, light clients might have to use a lot of bandwidth just to download block bodies. After all, block bodies is one of the scaling factors of the chain. The busier the chain, the bigger the block bodies.

While a light client could ask a full node to send back a proof that a transaction is in a certain block, the full node could simply lie and pretend that the transaction isn't there. There is no way to prove that a transaction is not in a certain block body without sending the entire block body.

I don't have a solution to that problem, which is why I'm just opening an issue.

tomaka:
One way to solve the problem could be to switch `extrinsics_root` to using V1 of the trie: https://github.com/paritytech/substrate/blob/ece92d6b30ee7f9fbda7a84bee76f8ff6d2c92a1/frame/system/src/lib.rs#L150

If this is done, we could then add a networking request for light clients to ask full nodes for a Merkle proof of the extrinsics trie that only contains the hashes of the extrinsics.
This proof would be around `35 * k` to `40 * k` bytes, where `k` is the number of extrinsics. This seems like a reasonable size to me.

Switching the `extrinsics_root` to V1 would be a breaking change, but we could reuse the `state_version` field for the migration (and rename it to `system_version` for example) by setting it to `2`.

Another possibility could be to modify the extrinsics trie so that the trie keys are something like `concat(transaction_hash, index)`. This would make it possible for a full node to generate a very small Merkle proof that an extrinsic is or is not in the block body.

seunlanlege:
> `concat(transaction_hash, index)`

Light clients don't know the transaction index, i think just using transaction hash as the key is fine.

tomaka:
The light client doesn't need to know the transaction index.
When a light client wants to know whether a transaction of a certain hash is included, the full node would send to it a Merkle proof of all the nodes whose key starts with that hash.

josepot:
I’d like to propose a simpler and IMHO more pragmatic alternative to what was suggested above.

Instead of modifying the networking protocol and changing the `extrinsics_root` trie version, we could introduce a new runtime storage map: `System.Extrinsics`.

[...]

PR #42, opened 2023-10-25
Introduce System version to RuntimeVersion object

vedhavyas:
This RFC proposes adding `system_version` and remove `state_version` from `RuntimeVersion` object. This makes it possible to use `StateVersion::V1` for both Storage and Extrinsic root derivations

### DIFF
--- a/text/0042-extrinsics-state-version.md
+++ b/text/0042-extrinsics-state-version.md
@@ -17,6 +17,9 @@ we would like to propose adding `system_version` that can be used to derive both
 ## Motivation
 
 Since the extrinsic state version is always `StateVersion::V0`, deriving extrinsic root requires full extrinsic data.
+This would be problematic when we need to verify the extrinsics root if the extrinsic sizes are bigger. This problem is
+further explored in https://github.com/polkadot-fellows/RFCs/issues/19
+
 For `Subspace` project, we have an enshrined rollups called `Domain` with optimistic verification and Fraud proofs are
 used to detect malicious behavior.
 One of the `Fraud proof` variant is to derive `Domain` block extrinsic root on `Subspace`'s consensus chain.
--- a/text/0042-extrinsics-state-version.md
+++ b/text/0042-extrinsics-state-version.md
@@ -1,16 +1,18 @@
-# RFC-0042: Add Extrinsic State version to `RuntimeVersion`
+# RFC-0042: Add System version that replaces StateVersion on RuntimeVersion
 
-|                 |                                               |
-|-----------------|-----------------------------------------------|
-| **Start Date**  | 25th October 2023                             |
-| **Description** | Add extrinsic state version to RuntimeVersion |
-| **Authors**     | Vedhavyas Singareddi                          |
+|                 |                                             |
+|-----------------|---------------------------------------------|
+| **Start Date**  | 25th October 2023                           |
+| **Description** | Add System Version and remove State Version |
+| **Authors**     | Vedhavyas Singareddi                        |
 
 ## Summary
 
-At the moment, extrinsics root is derived using `StateVersion::V0` without the possibility to use different version
-unlike storage state version.
-It would be useful for projects like `Subspace` to make extrinsics state version be part of the `RuntimeVersion`.
+At the moment, we have `system_version` field on `RuntimeVersion` that derives which state version is used for the
+Storage.
+We have a use case where we want extrinsics root is derived using `StateVersion::V1`. Without defining a new field
+under `RuntimeVersion`,
+we would like to propose adding `system_version` that can be used to derive both storage and extrinsic state version.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


text/0042-extrinsics-state-version.md
```diff
@@ -0,0 +1,102 @@
...
+pub const VERSION: RuntimeVersion = RuntimeVersion {
+		spec_name: create_runtime_str!("rococo"),
+		impl_name: create_runtime_str!("parity-rococo-v2.0"),
+		authoring_version: 0,
+		spec_version: 10020,
+		impl_version: 0,
+		apis: RUNTIME_API_VERSIONS,
+		transaction_version: 22,
+		state_version: 1,
+		extrinsic_state_version: 0,
```
bkchr, line 56: Maybe we should not add this here and instead only bump the `CoreApi` version to `5`. This should signal that the `state_version` is also used for the extrinsic root calculation.

vedhavyas: Since `state_version` is using V1, when runtime is updated, all the projects will move from V0 to V1 for extrinsic derivation. I'm not sure if that is okay from existing projects but I don't strong opinion on this actually. I was mainly aiming to not break as much as possible

bkchr: Yeah, I also don't know what is the best way forward...

Let's wait for some other people to comment. The RFC in general looks good to me.

tomaka: IMO `state_version` should be renamed to something else (`system_version` or whatnot), and a value of `2` means that the extrinsics trie also uses trie V1.
In other words, this should be just a version number that is increased whenever we realize that there's a better way to do things, and not a way to pimp your blockchain the way you want.

bkchr: Hmm. Yeah we probably could do this. :shrug:

vedhavyas: One clarifying question here is if we use `system_version` to `2`
Would that mean both storage state_version and extrinsic state_version uses always `1` ?
From what I see on the polakdot runtime here - https://github.com/polkadot-fellows/runtimes/blob/main/relay/polkadot/src/lib.rs#L139
Its till uses `state_version` 0 and so it would still use `system_version` as 0 then, which is fine.

But if other existing chains wants to use extrinsic state_version `1` while keeping the storage state_version `0`, this would not be possible with this just one `system_version` and they might accidentally break their consensus.

I still think adding this extrinsic state_version would be ideal and solves the problem above.
Thoughts @bkchr ?

vedhavyas:
Would be really helpful if we can get some more feedback on this one from other team members :)

text/0042-extrinsics-state-version.md
bkchr: Kusama will be upgraded by to `state_version: 1` with the next runtime upgrade that should be applied on monday. Then Polkadot will also follow this. Generally all chains should upgrade to this state version, especially parachains as it can improve the storage proof size, aka giving more space when building a POV. Given this, I even think that the idea of Pierre with `system_version` is a good one.

vedhavyas:
@bkchr @tomaka The RFC is updated as per our discussion above. Please let me know your feedback

text/0042-extrinsics-state-version.md
```diff
@@ -0,0 +1,104 @@
...
+		transaction_version: 22,
+		system_version: 1,
+	};
+```
+
+## Drawbacks
+
+Since the `CoreApi` version would need to be bumped, there would require a Runtime upgrade for existing projects
+with `system_version` set to `1` so that it wont brick their chain. New projects can choose between `1` or `2`
+depending on their use-case.
```
tomaka, line 65: I don't see why `CoreApi` would need to be bumped. This whole paragraph seems obsolete.

vedhavyas: Should be corrected now

tomaka:
The `Motivation` section is not super clear. Another motivation is https://github.com/polkadot-fellows/RFCs/issues/19.
But I would vote yes.

vedhavyas:
> The Motivation section is not super clear.

This is the only use case we have right now. Is the wording confusing or does not meet standards?

>  Another motivation is https://github.com/polkadot-fellows/RFCs/issues/19.

I can tag this in the motivation section if it helps

tomaka:
I just mean that someone who isn't deeply familiar with the trie format and V0 and V1 will not understand the "Motivation" section. But that's not a big deal to me.

vedhavyas:
@tomaka what would be the next steps here ?
I can submit a change set to polkadot-sdk as well. Please let me know

nazar-pc:
Any blockers for moving forward with this?

ggwpez:
> Any blockers for moving forward with this?

"Just" needs some approvals 😄

tomaka:
Looks good to me.

One possibility could be to use `3` instead of `2`, so that the field can be interpreted as a bitfield, but then we'll be stuck if for example there's a new trie version. Overall whether to use a bitfield or plain numerical values would require anticipating how we're going to use this field in the future, which isn't really possible, and I don't think it matters much.

tomaka:
If there's no other comment, I can submit your RFC for voting.

vedhavyas:
No other changes from my end. 
Thank you!!

tomaka:
Proposed: https://collectives.polkassembly.io/referenda/70

bkchr (approved):

tomaka:
/rfc process 0x10ac786bc859fbaad56ca436bd822a01e0f120d37f85680621700323d35756d9

tomaka:
/rfc process 0x2cd018900d9c915ce8104cc6e37141b663cb46d13aadbff9dbfa038211228789

vedhavyas:
@tomaka Can you explain what is the process next? I would be happy to produce a PR for this RFC

bkchr:
@vedhavyas feel free to open a pr and link to the RFC in the pr description. Thank you!

anaelleltd:
@vedhavyas, do you have an update on the status of this RFC?
Thanks.

bkchr:
Was implemented [here[(https://github.com/paritytech/polkadot-sdk/pull/4257).

PR #43, opened 2023-10-30
Introduce `storage_proof_size` Host Function for Improved Parachain Block Utilization

skunert:
https://github.com/paritytech/polkadot-sdk/pull/1462 has been open for a while and is now well-reviewed. Opening an RFC since it introduces a new host function that is relevant for parachains and light-clients.

### DIFF
--- a/text/0000-storage-proof-size-hostfunction.md
+++ b/text/0000-storage-proof-size-hostfunction.md
@@ -1,0 +1,48 @@
+# RFC-0000: Introduce `storage_proof_size` Host Function for Improved Parachain Block Utilization
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 30 October 2023                                                                    |
+| **Description** | Host function to provide the storage proof size to runtimes.                                                                    |
+| **Authors**     | Sebastian Kunert                                                                                            |
+
+## Summary
+
+This RFC proposes a new host function for parachains, `storage_proof_size`.  It shall provide the size of the currently recorded storage proof to the runtime. Runtime authors can use the proof size to improve block utilization by retroactively reclaiming unused storage weight.
+
+## Motivation
+The number of extrinsics that are included in a parachain block is limited by two constraints: execution time and proof size. FRAME weights cover both concepts, and block-builders use them to decide how many extrinsics to include in a block. However, these weights are calculated ahead of time by benchmarking on a machine with reference hardware. The execution-time properties of the state-trie and its storage items are unknown at benchmarking time. Therefore, we make some assumptions about the state-trie:
+- **Trie Depth:** We assume a trie depth to account for intermediary nodes.
+- **Storage Item Size:** We make a pessimistic assumption based on the `MaxEncodedLen` trait.
+
+These pessimistic assumptions lead to an overestimation of storage weight, negatively impacting block utilization on parachains.
+
+In addition, the current model does not account for multiple accesses to the same storage items. While these repetitive accesses will not increase storage-proof size, the runtime-side weight monitoring will account for them multiple times. Since the proof size is completely opaque to the runtime, we can not implement retroactive storage weight correction.
+
+A solution must provide a way for the runtime to track the exact storage-proof size consumed on a per-extrinsic basis.
+
+## Stakeholders
+- **Parachain Teams:** They should include this host function in their runtime and node.
+- **Light-client Implementors:** They should include this host function in their runtime and node.
+
+## Explanation
+This RFC proposes a new host function that exposes the storage-proof size to the runtime. As a result, runtimes can implement storage weight reclaiming mechanisms that improve block utilization.
+
+This RFC proposes the following host function signature:
+```rust
+fn storage_proof_size() -> u64;
+```
+The host function MUST return an unsigned 64-bit integer value representing the current proof size. In block-execution and block-import contexts, this function MUST return the current size of the proof. To achieve this, parachain node implementors need to enable proof recording for block imports. In contexts without proof recording, this function MUST return 0. 
+
+## Performance, Ergonomics, and Compatibility
+### Performance
+Parachain nodes need to enable proof recording during block import to correctly implement the proposed host function. Benchmarking conducted with balance transfers has shown a performance reduction of around 0.6% when proof recording is enabled. 
+
+### Ergonomics
+The host function proposed in this RFC allows parachain runtime developers to keep track of the proof size. Typical usage patterns would be to keep track of the overall proof size or the difference between subsequent calls to the host function.
+
+### Compatibility
+Parachain teams will need to include this host function to upgrade.
+
+## Prior Art and References
+- Pull Request including proposed host function: [PoV Reclaim (Clawback) Node Side](https://github.com/paritytech/polkadot-sdk/pull/1462).
+- Issue with discussion: [[FRAME core] Clawback PoV Weights For Dispatchables](https://github.com/paritytech/polkadot-sdk/issues/209#top)
--- a/text/0000-storage-proof-size-hostfunction.md
+++ b/text/0000-storage-proof-size-hostfunction.md
@@ -1,4 +1,4 @@
-# RFC-0000: Introduce `storage_proof_size` Host Function for Improved Parachain Block Utilization
+# RFC-0043: Introduce `storage_proof_size` Host Function for Improved Parachain Block Utilization
 |                 |                                                                                             |
 | --------------- | ------------------------------------------------------------------------------------------- |
 | **Start Date**  | 30 October 2023                                                                    |
--- a/text/0043-storage-proof-size-hostfunction.md
+++ b/text/0043-storage-proof-size-hostfunction.md
@@ -31,7 +31,7 @@ This RFC proposes the following host function signature:
 ```rust
 fn storage_proof_size() -> u64;
 ```
-The host function MUST return an unsigned 64-bit integer value representing the current proof size. In block-execution and block-import contexts, this function MUST return the current size of the proof. To achieve this, parachain node implementors need to enable proof recording for block imports. In contexts without proof recording, this function MUST return 0. 
+The host function MUST return an unsigned 64-bit integer value representing the current proof size. In block-execution and block-import contexts, this function MUST return the current size of the proof. To achieve this, parachain node implementors need to enable proof recording for block imports. In other contexts, this function MUST return 0.
 
 ## Performance, Ergonomics, and Compatibility
 ### Performance
--- a/text/0043-storage-proof-size-hostfunction.md
+++ b/text/0043-storage-proof-size-hostfunction.md
@@ -21,17 +21,17 @@ In addition, the current model does not account for multiple accesses to the sam
 A solution must provide a way for the runtime to track the exact storage-proof size consumed on a per-extrinsic basis.
 
 ## Stakeholders
-- **Parachain Teams:** They must include this host function in their runtime and node.
-- **Light-client Implementors:** They should include this host function in their runtime and node.
+- **Parachain Teams:** They MUST include this host function in their runtime and node.
+- **Light-client Implementors:** They SHOULD include this host function in their runtime and node.
 
 ## Explanation
 This RFC proposes a new host function that exposes the storage-proof size to the runtime. As a result, runtimes can implement storage weight reclaiming mechanisms that improve block utilization.
 
 This RFC proposes the following host function signature:
 ```rust
-fn ext_storage_proof_size_version_1() -> Option<u64>;
+fn ext_storage_proof_size_version_1() -> u64;
 ```
-The host function MUST return an `Option` of an unsigned 64-bit integer value representing the current proof size. In block-execution and block-import contexts, this function MUST return a `Some` variant including the current size of the proof. To achieve this, parachain node implementors need to enable proof recording for block imports. In other contexts, this function MUST return `None`.
+The host function MUST return an unsigned 64-bit integer value representing the current proof size. In block-execution and block-import contexts, this function MUST return the current size of the proof. To achieve this, parachain node implementors need to enable proof recording for block imports. In other contexts, this function MUST return 18446744073709551615 (u64::MAX), which represents disabled proof recording.
 
 ## Performance, Ergonomics, and Compatibility
 ### Performance
--- a/text/0043-storage-proof-size-hostfunction.md
+++ b/text/0043-storage-proof-size-hostfunction.md
@@ -29,7 +29,7 @@ This RFC proposes a new host function that exposes the storage-proof size to the
 
 This RFC proposes the following host function signature:
 ```rust
-fn storage_proof_size() -> u64;
+fn ext_storage_proof_size_version_1() -> u64;
 ```
 The host function MUST return an unsigned 64-bit integer value representing the current proof size. In block-execution and block-import contexts, this function MUST return the current size of the proof. To achieve this, parachain node implementors need to enable proof recording for block imports. In other contexts, this function MUST return 0.
 
--- a/text/0043-storage-proof-size-hostfunction.md
+++ b/text/0043-storage-proof-size-hostfunction.md
@@ -21,7 +21,7 @@ In addition, the current model does not account for multiple accesses to the sam
 A solution must provide a way for the runtime to track the exact storage-proof size consumed on a per-extrinsic basis.
 
 ## Stakeholders
-- **Parachain Teams:** They should include this host function in their runtime and node.
+- **Parachain Teams:** They must include this host function in their runtime and node.
 - **Light-client Implementors:** They should include this host function in their runtime and node.
 
 ## Explanation
@@ -29,9 +29,9 @@ This RFC proposes a new host function that exposes the storage-proof size to the
 
 This RFC proposes the following host function signature:
 ```rust
-fn ext_storage_proof_size_version_1() -> u64;
+fn ext_storage_proof_size_version_1() -> Option<u64>;
 ```
-The host function MUST return an unsigned 64-bit integer value representing the current proof size. In block-execution and block-import contexts, this function MUST return the current size of the proof. To achieve this, parachain node implementors need to enable proof recording for block imports. In other contexts, this function MUST return 0.
+The host function MUST return an `Option` of an unsigned 64-bit integer value representing the current proof size. In block-execution and block-import contexts, this function MUST return a `Some` variant including the current size of the proof. To achieve this, parachain node implementors need to enable proof recording for block imports. In other contexts, this function MUST return `None`.
 
 ## Performance, Ergonomics, and Compatibility
 ### Performance


text/0000-storage-proof-size-hostfunction.md
```diff
@@ -0,0 +1,48 @@
...
+- **Light-client Implementors:** They should include this host function in their runtime and node.
+
+## Explanation
+This RFC proposes a new host function that exposes the storage-proof size to the runtime. As a result, runtimes can implement storage weight reclaiming mechanisms that improve block utilization.
+
+This RFC proposes the following host function signature:
+```rust
+fn storage_proof_size() -> u64;
+```
+The host function MUST return an unsigned 64-bit integer value representing the current proof size. In block-execution and block-import contexts, this function MUST return the current size of the proof. To achieve this, parachain node implementors need to enable proof recording for block imports. In contexts without proof recording, this function MUST return 0.
```
tomaka, line 34: > In contexts without proof recording, this function MUST return 0. 

This would introduce non-determinism, unless you specify in which situations proof recording is enabled and which situations it is disabled.

tomaka:
Another of my worries is: how do you handle situations where the runtime does this?

```rust
for key in all_possible_keys_that_exist() {
    drop(storage_get(&key));
}
``` 

This would lead to a very very large proof that the host needs to keep in memory (I don't think you can realistically implement proof recording using the database). This means that you need to introduce a client-side limit to the maximum size of a proof.

text/0000-storage-proof-size-hostfunction.md
burdges: It's deterministic for polkadot verification though.

Yeah, parachains could screw up light client protocols, or even their own block building, by doing this, but these cases seem like "parachain authors fucked up".  Anyways, they could already do this by using `static mut`s, which they must do sometimes given some design choices in Frame.

skunert: > IMO the only way that makes sense to me would be to always unconditionally enable proof recording

Hmm yes I see, the downside of this is that it comes with a performance impact.

Wdyt about adjusting the wording to the opposite, i.e. the host function must always return 0 except for block building and import, regardless of proof recording setting?

tomaka:
I also would like to raise the concern that the algorithm used to generate a proof must now be precisely spec'ced (at least on paper).

I'm not completely sure of what I'm saying because it's been a long time since I've looked into this, but if I remember correctly Substrate sometimes generates sub-optimal proofs (i.e. some proofs include slightly too much information). Adding this host function means that other host implementers must reproduce the exact same flaws as Substrate, but also that adding these optimizations would break backwards-compatibility.

text/0000-storage-proof-size-hostfunction.md
tomaka: > Wdyt about adjusting the wording to the opposite, i.e. the host function must always return 0 except for block building and import, regardless of proof recording setting?

I guess that works.

However if I remember correctly `BlockBuilder_apply_extrinsic` is used when generating a block but also outside of block building to know whether a transaction will succeed (which is an information that validating a transaction alone doesn't provide), so the implementation of this function would need to make sure to not accidentally say that transactions are invalid if that function returns 0.

It's not a blocker, but it's an extra footgun that wouldn't be there if proof record was always enabled.

burdges:
> I also would like to raise the concern that the algorithm used to generate a proof must now be precisely spec'ced (at least on paper).

Parachains are necessarily free to chose their own storage proof algorithms. 

> if I remember correctly Substrate sometimes generates sub-optimal proofs

Yeah, there is not necessarily a well defined storage proof size until the collator produces the PoV block, like adjusting the relay parent could change the size even if the block does not otherwise change.  Any given PoV block does however have some well defined storage proof size, so this makes sense to the collator and polkadot, but not before.

tomaka:
> Parachains are necessarily free to chose their own storage proof algorithms.

Is the new host function only accessible to parachain runtimes when executed by parachain hosts? Or must the relay chain host also implement it? If it's the former, then this change doesn't even need an RFC, as it would be an implementation detail of parachains. If it's the latter, then there's a contradiction here.

bkchr:
> Is the new host function only accessible to parachain runtimes when executed by parachain hosts? Or must the relay chain host also implement it?

Only parachain hosts need to implement it. The RC isn't involved into this. However, light clients (like smoldot) will may need for re-executing blocks or similar.

text/0000-storage-proof-size-hostfunction.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Motivation
+The number of extrinsics that are included in a parachain block is limited by two constraints: execution time and proof size. FRAME weights cover both concepts, and block-builders use them to decide how many extrinsics to include in a block. However, these weights are calculated ahead of time by benchmarking on a machine with reference hardware. The execution-time properties of the state-trie and its storage items are unknown at benchmarking time. Therefore, we make some assumptions about the state-trie:
+- **Trie Depth:** We assume a trie depth to account for intermediary nodes.
+- **Storage Item Size:** We make a pessimistic assumption based on the `MaxEncodedLen` trait.
+
+These pessimistic assumptions lead to an overestimation of storage weight, negatively impacting block utilization on parachains.
+
+In addition, the current model does not account for multiple accesses to the same storage items. While these repetitive accesses will not increase storage-proof size, the runtime-side weight monitoring will account for them multiple times. Since the proof size is completely opaque to the runtime, we can not implement retroactive storage weight correction.
```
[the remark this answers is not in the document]
bkchr: > You're accepting the fact that the runtime can't determine the proof size because it's too complicated

The current way it is implemented, it is impossible to get the proof size in a different way. The runtime would need to have raw access to the trie to compute this data. The runtime is currently nowhere close to have access to these information.

skunert:
> However, light clients (like smoldot) will may need for re-executing blocks or similar.

This was my main motivator for opening this RFC. Even though this will not be included on the relay chain, light clients will need to have the host function.

burdges:
If you run PoV blocks then you'd want this host function.  Could we maybe disable this host function when not running PoV blocks though?

If smoldot runs a PoV block then yes sure it's expose this, but then it knows the size.  If it runs a lone tx without full context, then it has nothing to expose here.  In practice, this means `validate_block` can call this host function, but the parachain authors fucked up if it gets invoked elsewhere under `execute_block`.

skunert:
> Amusingly collators would not be validating one another's weight reclamation, but depending upon the relay chain for this. lol

What do you mean by this? If a collator would build a block reclaiming too much weight, the block would fail to import on other collators. And fail `validate_block` too of course.

skunert:
> > Parachains are necessarily free to chose their own storage proof algorithms.
> 
> Is the new host function only accessible to parachain runtimes when executed by parachain hosts? Or must the relay chain host also implement it? If it's the former, then this change doesn't even need an RFC, as it would be an implementation detail of parachains. If it's the latter, then there's a contradiction here.

So yeah, maybe an RFC was not necessary after all. If nothing changes here until next week, I will close here and merge the PR.

tomaka:
I said that this doesn't need an RFC because parachains are in principle free to do whatever they want here.
However there's an argument to make that systems parachains should also be precisely specced.

bkchr:
> I said that this doesn't need an RFC because parachains are in principle free to do whatever they want here.

Yes they are, but smoldot would also need to support "whatever they are doing". We can also close this, but I think it is worth to have it here.

text/0043-storage-proof-size-hostfunction.md
```diff
@@ -0,0 +1,48 @@
...
+## Stakeholders
+- **Parachain Teams:** They must include this host function in their runtime and node.
+- **Light-client Implementors:** They should include this host function in their runtime and node.
+
+## Explanation
+This RFC proposes a new host function that exposes the storage-proof size to the runtime. As a result, runtimes can implement storage weight reclaiming mechanisms that improve block utilization.
+
+This RFC proposes the following host function signature:
+```rust
+fn ext_storage_proof_size_version_1() -> Option<u64>;
```
tomaka, line 32: Returning an `Option` will cause the host to allocate 5 bytes of memory that the runtime will then free. This is terribly inefficient.
I would suggest returning `-1` rather than `None`.

skunert: The goal was to get rid of the ambiguity between empty proof and proof recording disabled. 

> I would suggest returning -1 rather than None.
Assuming you mean `u32::MAX`, I am also fine with that :+1: .

tomaka: > Assuming you mean u64::MAX

Yes.
Substrate does some automagic interface adjustments, and under the hood this `u64` is actually a `i64`.

tomaka:
My concern hasn't really been addressed:

> I also would like to raise the concern that the algorithm used to generate a proof must now be precisely spec'ced (at least on paper).

> I'm not completely sure of what I'm saying because it's been a long time since I've looked into this, but if I remember correctly Substrate sometimes generates sub-optimal proofs (i.e. some proofs include slightly too much information), for example IIRC when deleting a prefix. Adding this host function means that other host implementers must now reproduce the exact same undocumented optimizations and flaws as Substrate, but also that adding these optimizations later would be difficult as it would break backwards-compatibility.

This change would make it very hard to fix the sub-optimal proof generation and to write an alternative implementation of a system parachain collator.

skunert:
> At the moment, you can generate a proof that contains unnecessary trie nodes and everything works fine. So it's fine to just include too many things in proofs "just in case", and as far as I understand (by looking at proofs while debugging smoldot) that's what Substrate sometimes does.

Okay, this indeed would be problematic. I was not aware of this behaviour. We should spec the proofs properly then.

tomaka:
Again maybe I'm wrong, or maybe it's been optimized since then. My concern mostly comes from the fact that this is pretty complicated.

From the top of my head:

- If the runtime calls `storage_exists` or `storage_read` with a size of 0, then you don't need to include the storage value when the trie version is V1.
- If the runtime calls `storage_append` but never reads back the value afterwards, the value that was there doesn't need to be included in the proof, but that's really contentious.
- If you call `storage_clear` you don't need to include the deleted keys in the proof, except that if you call `storage_next_key` where the parameter is before the ones you've deleted and the next key is after the ones you've deleted, then you need to include in the proof the keys that you've deleted.
- Calling `storage_root` means that you have to include in the proof all the modifications to the storage you've previously done. So for example if you call `storage_set` but not `storage_root` then the modified entry doesn't need to be included, but if you do call `storage_root` later then the modified entry does need to be included.

burdges:
Another approach would opaque newtype(s) vaguely resembling `std::time::{Duration,Instant,SystemTime}`.

`storage_proof_size` returns an `pub struct ProofSizeGuard<'a>(u32, PhantomData<&'a u32>);`, which satisfies `PartialOrd<u32>`, and maybe other traits, but does not satisfy `PartialEq<_>`.  We require drop `ProofSizeGuard` be dropped before invoking `storage_proof_size` again.  Alternatively `storage_proof_size` always returns monotonically increasing values until all

After collation, `storage_proof_size` returns an opaque actual proof size from the PoV block, but before collation it returns whatever approximation the system currently believes.  Any parachain could alter its internal `storage_proof_size` implementation, just like it could add custom storage, but we require `storage_proof_size(parablock+PoV) < storage_proof_size(parablock)` or else collators could produce invalid blocks (which backers reject, so no slashing).

In principle, `storage_proof_size` both increases and decreases unpredictably while building blocks, aka non-determinism, but only decreases could depend upon the block so far, while increases should be bounded by some function of whatever you just added. 

A parachain could break these invariants using unsafe code, but they could do much worse anyways.

skunert:
> Again maybe I'm wrong, or maybe it's been optimized since then. My concern mostly comes from the fact that this is insanely complicated.
> 
> Just from the top of my head:
> 
> * If the runtime calls `storage_exists` or `storage_read` with a size of 0, then you don't need to include the unhashed storage value when the trie version is V1.
> * If the runtime calls `storage_append` but never reads back the value afterwards, the value that was there doesn't need to be included in the proof, but I imagine that this is really hard to implement.
> * If you call `storage_clear` you don't need to include the deleted keys in the proof, except that if afterwards you call `storage_next_key` where the parameter is before the ones you've deleted and the next key is after the ones you've deleted, then you need to include in the proof the keys that you've deleted.
> * Calling `storage_root` means that you have to include in the proof all the modifications to the storage you've previously done. So for example if you call `storage_set` but not `storage_root` then the modified entry doesn't need to be included, but if you do call `storage_root` later then the modified entry does need to be included.
> * In a trie V1, if you call `storage_set` with a value identical to the one that was already there, then `storage_root`, then the proof doesn't need to include the unhashed storage value of the node that was set, since the receiver of the proof could compare the hash of the value being set with the existing hash.
> 
> In general trying to figure out what should be included in the case of `storage_root` makes my brain explode.

I thought about this some more and do not think that this is actually a blocker for this RFC.

Initially, I was under the impression that we were talking about a suboptimal storage-proof algorithm in substrate. But the cases you are referencing are hostfunction implementation related. If a hostfunction reads more than strictly needed, we include extra items in the proof.

However, that is separate from the changes proposed here. If you want to implement a custom collator node today, you already need to be super careful. Since the parachain runtime contains the `validate_block` function, which has externalities and hostfunction implementations baked in, implementors need to mimic the externalities and hostfunction behavior anyway. If a custom collator generates a smaller proof than what we expect in the PVF, it would be a problem. In other words, you already have to follow the quirks of the substrate implementation today. Otherwise, the PVF might reject your block.

Not a great state to be in, but I believe the proposed hostfunction here does not change much compared to the current state. We should still correctly specify the host functions, but this sounds like a separate effort to me.

tomaka:
> However, that is separate from the changes proposed here. If you want to implement a custom collator node today, you already need to be super careful. Since the parachain runtime contains the validate_block function, which has externalities and hostfunction implementations baked in, implementors need to mimic the externalities and hostfunction behavior anyway. If a custom collator generates a smaller proof than what we expect in the PVF, it would be a problem. In other words, you already have to follow the quirks of the substrate implementation today. Otherwise, the PVF might reject your block.

> Not a great state to be in, but I believe the proposed hostfunction here does not change much compared to the current state. We should still correctly specify the host functions, but this sounds like a separate effort to me.

I don't think that's accurate.

Yes, implementers need to mimic the externalities and host functions behavior anyway, but all the existing host functions are very clearly defined (maybe not explicitly in a document, but they're all pretty straight forward and their corner cases is easily figured).
In other words, mimicking the host function behavior is relatively easy. Smoldot does it, for example, and it wasn't that hard.

The one introduced in this RFC is the first host function that has tons of obscure corner cases with what seems to me to be over-the-roof implementation mimicking difficulty.

Your counter-argument also doesn't address the backwards compatibility concern.
Introducing this host function makes it impossible to modify the proof generation code without risking a netsplit. You'd have to do it in a centralized way by stopping all your collators at the same block, upgrading all of them simultaneously, then restarting all of them.

burdges:
We definitely want parachains to ship their own custom storage designs, while using substrate.  As a rule, these must iterate on their exact custom storage scheme.

We should iterate on the existing storage used in polkadot & system parachains too.  Although painful, I suppose migrations should be possible, but we've good reasons to be flexible before collation anyways:  You cannot know PoV size until you've built the full block, because you cannot know what merkle paths unify.  

It's worse:  If you've some KZG batching which does inner pairing products, then your PoV proof becomes very expensive in CPU time, so only the one collator ever computes.  It's also plausible other collators must learn the block first.  We then have a situation where the parachain itself cannot know what `storage_proof_size` should return on the relay chain. 

Again afaik there is no problem here with doing a `storage_proof_size` host call, but it should manage the inherent non-determinism correctly, not be another incorrect estimate.

skunert:
> The one introduced in this RFC is the first host function that has tons of obscure corner cases with what seems to me to be over-the-roof implementation mimicking difficulty.

Where do you see these corner cases/what is "proof generation code" to you?
- The trie crate and how it generates the proof from storage operations?
- The hostfunctions/externalities and what storage items they touch to provide their functionality?

If I understand correctly, you are suggesting that it is easy to mimic the host function functionality, but a custom implementation would not necessarily need to arrive at exactly the same proof layout.
This is what I don't understand. If your proof layout is any different from what substrate currently generates, your submitted block might fail. 

Taking your first example:
> If the runtime calls storage_exists or storage_read with a size of 0, then you don't need to include the unhashed storage value when the trie version is V1.

Whether or not you technically need to include this in the proof, as a implementor you need to exactly do what substrate does. If you produce a correct but smaller proof but substrate would include more nodes for whatever reason, the PVF will still try to call `storage_exists` with the baked in substrate stack. So what freedom do implementors lose with this host function? I don't yet see the wiggle room you are seeing (that is if you want to produce a block).

> Your counter-argument also doesn't address the backwards compatibility concern.

The above means to me that we are already in this situation. How would you modify the storage proof algorithm today? Since the parachain runtime essentially contains a whole trie/ext/hostfunction stack, you need to upgrade them in sync (and keep the old version around). At least this is how my mental model works.

librelois:
I agree that imposing determinism on such a complex algorithm poses problems, at the very least upgradeability problems.

One way of getting around the problem of the non-determinism of the proof generation algorithm is to ask the collator to inject the result of each call to this host function into the block, in the form of a vector of numbers, and to reuse this vector directly for importing and validating the block. This can be integrated into a new storage proof format, or injected into a special unsigned extrinsic at the end of the block (which would mean that the host function could not be used in on_finalize).

tomaka:
> If I understand correctly, you are suggesting that it is easy to mimic the host function functionality, but a custom implementation would not necessarily need to arrive at exactly the same proof layout.
This is what I don't understand. If your proof layout is any different from what substrate currently generates, your submitted block might fail.

What I'm saying is that it is easy to mimic every host function other than `ext_storage_proof_size_version_1`, but that mimicking `ext_storage_proof_size_version_1` is too complicated.

I am not suggesting that implementations should have freedom to generate different proofs. I'm actually saying the exact opposite: all implementations must generate the exact same proofs (apart from the ordering of items within the proof, which doesn't matter for the size), and it is too difficult to guarantee this.

Since "all implementations" also includes "multiple versions of Substrate", it blocks Substrate from ever improving the size of the proofs it generates.

What makes this dangerous is that since the code is at a very high level of abstraction compared to the actual proof being generated. It can therefore be really subtle: you can modify two small lines of code in the `trie` crate and accidentally break backwards compatibility. And you will find out the problem only when you want to author a block whose proof size is within the difference between implementations, which I guess will happen rarely, making this even more subtle.

Basically, this looks to me like a receipt for disaster. Either every full node uses the same version of Substrate, or one day some nodes but not all will refuse a block and someone will need to do a very long and painful cowboy-investigation of the problem.

librelois:
> Either every full node uses the same version of Substrate

This is not the case in production, and it never will be: it's far too restrictive. Think also about collators and full nodes for parachains, not only for relay, it's hard enough to update new substrate versions quickly, and it takes even longer for collators or full nodes for external partineraries (like block explorers).
In practice, on the Moonbeam side, for example, we currently have 4 different versions of full nodes in production!

More generally, I think that the storage proof generation algorithm should not be part of the consensus, the only requirement should be the validity of the proof and a size limit. If the host function is integrated as it stands, the proof generation algorithm will be  part of the consensus rules.

It would be great to take more time to evaluate alternative solutions, such as injecting the results of the host function into the block, or finding a way to satisfy the need without a host function.

burdges:
> One way of getting around the problem of the non-determinism of the proof generation algorithm is to ask the collator to inject the result of each call to this host function into the block, in the form of a vector of numbers, and to reuse this vector directly for importing and validating the block.

At some level, our whole goal here is to "reclaim" weight not really used by tx, as well as to have separate space and cpu weights.  If we could do what you say then we could just adjust the weights in roughly the current weight system, but this obviously does not work correctly.

Instead, we want the space weights to more accurately reflect the size in the final PoV block.  If someone ships an optimization that improves space under niche conditions then the blocks from collators who upgraded should benefit, while the blocks from other collators should not benefit.

This in inherently non-determinism is several ways:  Some collators do not have the new code.  Some collators have more CPU time to spend compressing the proof.  Some parachain designed will result in some collators not even being able to build the proof.  And all collators must consider the tx before they finish building the block.

Importantly, there are different classes of non-determinism:  (1) Any post-collation non-determinism is a treat to polkadot consensus, and must never be allowed.  (2) Pre-collation non-determinism in parablocks feels really weird, and complicates some things parachains might do, but should basically be fine for typical parachains who simply follow relay chain consensus.  (3) Now partial block non-determinism is annoying for UX, but it's also unavoidable for smart contracts, ala MEV.  We'll inherently increase our existing non-determinism of type (3) and add new some non-determinism of type (2).  Yes (2) requires some discussion, but if performance requires this then we should accept that and separately solve issues it creates for pre-finality grandpa.

librelois:
> our whole goal here is to "reclaim" weight not really used by tx

Yes, I know, and this will be the case, if the collators have the economic incentive to fill the block as much as possible, they have an interest in providing the smallest possible proof size to the runtime to be allowed to integrate more transactions and earn more fees.

> If we could do what you say then we could just adjust the weights in roughly the current weight system, but this obviously does not work correctly.

Maybe I'm missing something, but I don't understand why it wouldn't work. It seems to me that the purpose of this RFC is only to optimize the use of the PoV/proof_size resource, not the cpu_time.

With my proposal, if a collator generates more compact storage proofs, it's not a problem, because the relay and other collators and para full nodes won't need to regenerate the storage proof, just read it and check its validity.

burdges:
Incentives are irrelevant here:  Hosts are upgraded at different speeds.  Hosts have different hardware.  etc.  Adversaries can do whatever they can do which whatever non-determinism one permits, ala MEV.

> With my proposal, if a collator generates more compact storage proofs, it's not a problem, because the relay and other collators and para full nodes won't need to regenerate the storage proof, they just need to read it and check its validity (as it is now).

A relay chain validator already sees whatever proof the collator built, so they've no need for extra numbers, unless those numbers wind up being pessimistic over estimations, akin to our current weights system.

Parachain nodes cannot necessarily either see the full PoV block or compute the same storage proof, so they cannot produce correct values for the host calls, nor validate the numbers being inserted. 

We could simply accept the non-determinism at the parachain level since everything remains deterministic after collation at the relay chain level.  We'd encapsulate this non-determinism behind opaque types which discourage miss-use.  We'd warn parachains with unusual pre-collation consensus phases to either no use this mechanism, or else ensure it canno really break their consensus.

skunert:
> One way of getting around the problem of the non-determinism of the proof generation algorithm is to ask the collator to inject the result of each call to this host function into the block, in the form of a vector of numbers, and to reuse this vector directly for importing and validating the block.

This is very similar to what was originally proposed in the issue https://github.com/paritytech/polkadot-sdk/issues/209#issuecomment-1691687910 before we switched focus to the more "simple" solution proposed here. I guess it is an option to reevaluate the original approach.

 > Parachain nodes cannot necessarily either see the full PoV block or compute the same storage proof, so they cannot produce correct values for the host calls, nor validate the numbers being inserted.

The values delivered in the block would be used instead of the host calls. Its basically just the collator saying, "this is how much proof size was consumed at this point during block production.". Not sure how a collator would cheat by manipulating these number while going undetected. In the end, the weight during block import has to be correct and the PoV size has to be small enough for the relay to accept it.

burdges:
Alright, basti's comments there make more sense.  Yes, this complicates determinism, and fancy parachain consensus, but acceptably I guess.  If I understand..

We need this roughly host call in block building so that block producers know how much they've used.  It'll often return non-sensical and/or non-deterministic values but that's unavoidable, and acceptable in block production.

We do not need this host call after collation, aka in `validate_block`, because the relay chain enforces these limits elsewhere.  It's fine if the host call panics, returns `None`, or whatever here, so long as its deterministic.  

We're kinda stuck on what happens during parachain block imports.  As @tomaka says, we do not have deterministic execution if we return meaningful values.  Yet, we could theoretically make this deterministic in the distant future, or maybe even in the near future if we handled upgrade regimes more strictly, except we realistically impose such discipline upon parachain teams.

We've seemingly three choices for import:
1.  We make this host call panic, or return `None`, in both validation and import.  A parachain node cannot learn if the PoV block consumes too much space this way, but they could optimistically accept this, and adapt if rejected by the relay chain.  We do learn estimates during block production, so honest collators obey this.  
2.  We make this host call return local estimates during import.  It's break determinism for the prachain, which complicates fancy parachain consensus, but it's fine for standard parachains who obey the relay chain.  In validation, we'd either return the amount read so far from the PoV, or else panic or return `None`, both of which give determinism.
3.  We encode the block builder estimates somehow, which saves determinism during import by parachain nodes, and simplifies parachain debugging, but the value enforces nothing.  Ideally, these should be stripped when preparing the PoV block, since the relay chain has no use for them, so the host call then panics or returns `None` in the relay chain. 

We dislike complicating fancy parachain consensus, which rules out 2 probably.  Ain't much difference between 1 and 3, but 1 looks simplest, and avoids confusion.

skunert:
> We make this host call panic, or return None, in both validation and import. A parachain node cannot learn if the PoV block consumes too much space this way, but they could optimistically accept this, and adapt if rejected by the relay chain. We do learn estimates during block production, so honest collators obey this.

In your choices, 1 is just 3 without including the numbers in the block right? I don't think 1 can currently work. During import, nodes would need a way to calculate the overall consumed storage weight that goes into storage. If we don't deliver any values with the block, they will only have the benchmarked weights. This would also be a problem for the `on_idle` hook which can dynamically use the rest of the available weight. But they need to know how much can be reclaimed to arrive at the same result the collator did.

If we don't want 2 we should go with 3 IMO.

burdges:
In principle, there maybe advantages in leaving this up to the parachain team.  I suppose 2 works fine for simple parachains.  It's also basti's preference, or did he change his mind?  

If the host call returns `None` then any dependent code does nothing, meaning we forgo size weight checking, but yes alright `on_idle` requires the creator say the remainder.  It risks spam within the parachain too I guess.

If you do embed the values, then you could compare them with the amount read thus far during validation, so they must be correct or else the block becomes invalid.  In principle, parachain nodes could simply trust one another, which risks spam, but the relay chain would never approve unless correct. 

Approach:  You embed the values, do the comparison of read thus far in `validate_block`, and do no local comparisons on the parachain side.  As this risks spam, we document how parachain teams could compare vs local storage to thwart attacks, but suggest removing bad nodes instead. 

It's kinda annoying to embed a bunch of intermediate values when the relay chain only cares about the final total.

bkchr:
> Introducing this host function makes it impossible to modify the proof generation code without risking a netsplit. You'd have to do it in a centralized way by stopping all your collators at the same block, upgrading all of them simultaneously, then restarting all of them.

This is not any valid argument. You could say the same about adding new host function or whatever, that they risk a netsplit. However, we still handle them gracefully. The thing that we need to change is that we make the "proof generation" of each host function a part of the specification of each host function. This means that if you want to roll out any kind of optimization, you would need to roll out a new version of the host function and then the proof generation is versioned in the same way as the host function itself. This means, as long as the runtime uses the old host function, the proof generation doesn't change. I don't see any problem in doing so. 



> What makes this dangerous is that since the code is at a very high level of abstraction compared to the actual proof being generated. It can therefore be really subtle: you can modify two small lines of code in the `trie` crate and accidentally break backwards compatibility.

This problem already exists right now and it doesn't change with this proposal. The only thing I'm seeing here is to double down on testing to ensure that this doesn't happen in production (or better, reducing the possibility). And as @skunert already said, the Cumulus based PVF implementation is already sharing the same implementation details as the node implementation. This would mean that if you do some cowboy change, the PVF would stop accepting these blocks on the relay chain as a small detail changed. This also means that we already have this requirement of keeping this stuff in sync/upgradeable. 



> More generally, I think that the storage proof generation algorithm should not be part of the consensus, the only requirement should be the validity of the proof and a size limit. If the host function is integrated as it stands, the proof generation algorithm will be part of the consensus rules.

While we don't need it for this RFC and we could workaround with the proposal mentioned above, there will be a point in the future when the "proof generation" will be part of the consensus. When we bring "runtime weight metering", we will need to use the exact same proof generation everywhere to know if a transaction is running over the announced limits. Thus, I don't see any reason why we should go for some other way here, if we already know that we need this in the future any way. With runtime weight metering we will also need to track the proof size to abort a transaction if it used too much proof size. This needs to be deterministic as otherwise collators will be able to "reject" transactions, but still getting paid for it because it is in the block. 

We could argue that you don't need this at block import and you use the recorded proof size from the block producer, but you would still need to run the same logic in `validate_block`. As `validate_block` is the instance that ensures that the state transition is valid. This would remove some "headache" for historic blocks, but IMO we should stick to have the proof generation specced properly. I'm also down to help on this on @tomaka's new markdown spec.



> It's kinda annoying to embed a bunch of intermediate values when the relay chain only cares about the final total.

To make it clear again, this RFC is not touching any of the logic that is relevant to the relay chain.

skunert:
I agree to bastis points, the current proof algorithm is already deeply integrated and needs to stay.

I think the original approach is still feasible, and its up to the parachain to decide whether to ever call this function or not. I will open this up for voting tomorrow if no blockers are raised.

tomaka:
> This would remove some "headache" for historic blocks, but IMO we should stick to have the proof generation specced properly.

I'm personally not willing to retro-engineer Substrate in order to spec proof generation. I think it's way too insanely complicated, and I don't think people who have commented realize how complicated it is.

skunert:
> IMO for the sake of this RFC it should be assumed that speccing proof generation isn't possible.

I understand that it is complicated. But I don't think it is impossible. We could for example write some compliance tests in WASM that would ensure a host implementation matches what is expected. But that is just an idea on how to make it easier to test implementations.

The goal for this RFC was to raise awareness that such a hostfunction might come and gather feedback. We got feedback and the hostfunction proposed has been improved as a result. In the end, we already established that parachain teams are free to do as they like. If they have a fancy custom storage scheme, they might decide not to call the hostfunction or not even include it. If they don't like the approach as described in https://github.com/paritytech/polkadot-sdk/issues/209, they could even implement variant 3 (which would still require the presence of this hf). The solution is not perfect, but for a majority of the parachains I expect this to be a fitting mechanism.

xlc:
While I really need this RFC so that proof weight is not completely useless for parachains, I am also unsure if this RFC should be accepted as it is.
I still don't know how much technical details is needed for RFC but I am sure people won't be able to read this RFC and understand exactly the change and update alternative client to correctly implement the RFC.

xlc:
We also clearly don't have enough participation on this RFC. I also expect approval from alternative node implementation teams. If they find this RFC contains enough details, well, sure, that means it is enough.

bkchr:
> I also expect approval from alternative node implementation teams.

You would need to specify what kind of alternative node implementations? The relay chain? If yes, the relay chain node is never aware of this host function as already discussed in here.



> We also clearly don't have enough participation on this RFC (and most other RFCs).

Yeah and thus we are waiting until the participation magically appears? I mean we can then also just shutdown everything and move on.

xlc:
this RFC have already identified stakeholders so we should be able to ask them to take a read but yeah I agree we will need to just continue the work after we tried to get feedbacks and failed.

skunert:
> this RFC have already identified stakeholders so we should be able to ask them to take a read

Who do you mean by this? I mean you are right, this RFC alone does not contain all the details about the storage proof to implement it in a non substrate client.

But IMO the speccing of the proof generation does not need to be part of this RFC, it can be done in a separate document.

As you said the proof weight has currently some limitations that should be addressed. Since all collator implementations are currently substrate based anyway we should move forward in my opinion.

tomaka:
> Since all collator implementations are currently substrate based anyway we should move forward in my opinion.

This is a snake biting its tail. This kind of RFC is exactly why it won't be possible to create another implementation.

Please look at the low quality of the source code of Substrate. Do you think that Polkadot has a long term future if it keeps using this code? Every little change to the core takes years because there are literally less than 5 people in the world who are capable of working efficiently on Substrate.
Do you not think that it should be possible to at least leave the possibility to phase out Substrate and let other people build alternative, more clean, implementations in order to give a future to Polkadot'?

> But IMO the speccing of the proof generation does not need to be part of this RFC, it can be done in a separate document.

I would agree if this separate document had been written **before** this RFC.
This RFC is based on the assumption that it's even possible to write such a document, which, I repeat, I'm strongly skeptical about.

We all know that **nobody** is going to spec the proof generation. The idea of "speccing the proof generation in a separate document" is just a different way to sweep the problem under the carpet and delegate it to "someone else", while we all know that there won't be a "someone else".

skunert:
/rfc process 0x6386c7812bbed74cde165d0a074e69fa1b5d6845bf8aa914276305baf6f3b240

PR #44, opened 2023-11-06
Rent Based Registration Model

Szegoo:
This RFC proposes a new model for sustainable on-demand parachain registration, involving a smaller initial deposit and periodic rent payments. 

Initial discussion on this topic can be found [here](https://github.com/paritytech/polkadot-sdk/issues/1796)


text/0044-rent-based-registration.md
```diff
@@ -0,0 +1,199 @@
...
+
+With the support of on-demand parachains on Polkadot, there is a need to explore a new, more cost-effective model for registering validation code. In the current model, the parachain manager is responsible for reserving a unique `ParaId` and covering the cost of storing the validation code of the parachain. These costs can escalate, particularly if the validation code is large. We need a better, sustainable model for registering on-demand parachains on Polkadot to help smaller teams deploy more easily.
+
+This RFC suggests a new payment model to create a more financially viable approach to on-demand parachain registration. In this model, a lower initial deposit is required, followed by recurring payments upon parachain registration.
+
+This new model will coexist with the existing one-time deposit payment model, offering teams seeking to deploy on-demand parachains on Polkadot a more cost-effective alternative.
+
+## Requirements
+
+1. The solution SHOULD NOT affect the current model for registering validatoin code.
```
eskimor, line 23: typo

text/0044-rent-based-registration.md
```diff
@@ -0,0 +1,199 @@
...
+
+### On-demand parachain pruning
+
+If the rent is not paid, anyone has the option to prune the on-demand parachain and claim a portion of the initial deposit reserved for storing the validation code. Pruning only removes the validation code, while the head data and validation code hash are retained. The validation code hash is stored to allow anyone to register it again as well as to enable quicker re-registration by skipping the pre-checking process.
+
+The moment the rent is no longer paid, the parachain won't be able to purchase on-demand access, meaning no new blocks are allowed. This stage is called the "hibernation" stage, during which all the parachain-related data is still stored on-chain, but new blocks are not permitted. The reason for this is to ensure that the validation code is available in case it is needed in the dispute or approval checking subsystems. Waiting for one entire session will be enough to ensure it is safe to deregister the parachain.
```
eskimor, line 127: It is a soft form of deregistration though. ParaId is still claimed, only the PVF is pruned.

eskimor (commented):
Nice work!

phillux:
@Szegoo thanks for putting together this RFC. Overall, deposits need to be adjusted regularly and be as cheap as possible while maintaining sybil resistance and dos protection. I think the rent model makes sense but there are two issues I'm uncertain of:

1. Effectively this may be no different than lowering the cost-per-byte to a fraction of what it is now. From an attacker's point of view, they simply now can register multiple wasm blobs for the previous cost of one, then fail to pay rent later. Failing to pay the rent later is not an issue for an attacker in a DoS scenario.
2. Besides pruning, how is rent adjusted for runtime upgrades that add more bytes to their wasm blob?

For No. 2, this comes from a recent question I have not yet gotten a clear answer on, which is: 

Can an attacker pay one deposit, get a parachain registered, then perform runtime upgrades once per block (or are they only allowed one per epoch?) that spams the validators on the network without registering a new or modified deposit?

Overall, I think the RFC takes a good approach by increasing cost-per-byte while lowering initial setup costs. Provided the issues I've raised are clarified or addressed, I'd fully support this RFC.

text/0044-rent-based-registration.md
Szegoo: Yes, I think it can be called so. The main reason for this is to allow anyone to re-register the parachain under the same `ParaId` after it has been pruned.

Of course, there is always the option to perform a complete deregistration of a parachain by calling the `deregister` extrinsic.

Szegoo:
> Effectively this may be no different than lowering the cost-per-byte to a fraction of what it is now. From an attacker's point of view, they simply now can register multiple wasm blobs for the previous cost of one, then fail to pay rent later. Failing to pay the rent later is not an issue for an attacker in a DoS scenario.

I don't think this is a problem since the attacker is risking their initial deposit.

> Besides pruning, how is rent adjusted for runtime upgrades that add more bytes to their wasm blob?

This is a good point and I was a bit surprised to realize that the current implementation doesn't cover this scenario. The solution should probably be to simply request a deposit to cover the difference in the wasm blob size and based on that also re-adjust the recurring payment. I will add a new section to cover this.

text/0044-rent-based-registration.md
```diff
@@ -0,0 +1,201 @@
...
+	/// The duration is set in sessions instead of block numbers.
+	#[pallet::constant]
+	type RentDuration: Get<SessionIndex>;
+
+	/// The initial deposit amount for registering validation code.
+	///
+	/// This is defined as a percentage of the deposit that would be required in the regular
+	/// model.
+	#[pallet::constant]
+	type InitialRentDeposit: Get<Perbill>;
```
rphmeier, line 77: bikeshed: given that these are proportions, how about `RentalDepositProportion` and `RentalRecurringProportion`?

PR #45, opened 2023-11-07
Lowering Deposit Requirements on Polkadot and Kusama Asset Hub

poppyseedDev:
RFC-00xx-assethub, proposes lowering deposit requirements for NFT collection creation on Polkadot and Kusama Asset Hub to make it more accessible for artists. It draws upon the discussion in the [Polkadot Forum](https://forum.polkadot.network/t/polkadot-assethub-high-nft-collection-deposit/4262). While there are concerns about state bloat, the RFC also considers future governance models to dynamically adjust deposit requirements.


text/000x-assethub.md
```diff
@@ -0,0 +1,92 @@
...
+### Ergonomics
+The proposed change aims to enhance the user experience for artists, making Polkadot more accessible and user-friendly.
+
+### Compatibility
+The change doesn't impact compatibility.
+
+## Unresolved Questions
+- Determining the optimal new deposit amount that reduces the barrier to entry while preventing state bloat remains a subject for debate. What is the balance that allows for inclusivity without compromising the network's efficiency and security?
+
+## Future Directions and Related Material
```
ggwpez, line 90: I think a good future direction would be to establish some kind of process on how to agree on these values.  
Otherwise we will have the same discussion every 3-6 months i guess.

text/000x-assethub.md
```diff
@@ -0,0 +1,92 @@
...
+|---------------------------|----------------------------------|------------------------|-------------------------------|---------------------------|----------------------------------|--------------------------------------|
+| collectionDeposit         | 10 DOT                           | 50 $                   | 500 $                         | 0.1 DOT                   | 0.5 $                            | 5$                                   |
+| itemDeposit               | 0.01 DOT                         | 0.05 $                 | 0.5 $                         | 0.001 DOT                 | 0.005 $                          | 0.05$                                |
+| metadataDepositBase       | 0.20129 DOT                      | 1.00645 $              | 10.0645 $                     | 0.0020129 DOT             | 0.0100645 $                      | 0.100645$                            |
+| attributeDepositBase      | 0.2 DOT                          | 1 $                    | 10 $                          | 0.002 DOT                 | 0.01 $                           | 0.1$                                 |
```
ggwpez, line 36: It would be nice to see these values in relation to what is actually being stored on-chain.  
Then we can compare these values to proven spam-resistant storage prices (like in the preimage pallet).  
It should be possible to calculate the DOT/Byte to make it comparable.  

Otherwise they are just as random as at genesis, now just 1/10 😉

text/000x-assethub.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Motivation
+
+The current deposit of 10 DOT for collection creation on the Polkadot Asset Hub presents a significant financial barrier for many artists. By lowering the deposit requirements, we aim to encourage more artists to participate in the Polkadot NFT ecosystem, thereby enriching the diversity and vibrancy of the community and its offerings.
+
+## Stakeholders
+
+- **Artists**: Primary beneficiaries of the proposed change, particularly those who found the current deposit requirements prohibitive.
+- **KodaDot**: As the facilitator of artists' relations, KodaDot has a vested interest in making the platform more accessible.
+- **Polkadot Community**: Stands to benefit from an influx of artists, creators and diverse NFT collections, enhancing the overall ecosystem.
```
ggwpez, line 21: I think the community is also interested in the security of the network.

text/000x-assethub.md
```diff
@@ -0,0 +1,92 @@
+# RFC-0000: Lowering Deposit Requirements on Polkadot Asset Hub
+
+|                 |                                                                                                    |
+| --------------- | -------------------------------------------------------------------------------------------------- |
+| **Start Date**  | 2 November 2023                                                                                    |
+| **Description** | A proposal to reduce the minimum deposit required for collection creation on the Polkadot Asset Hub, making it more accessible and affordable for artists. |
+| **Authors**     | [Aurora Poppyseed](https://github.com/poppyseedDev), [Just_Luuuu](https://github.com/justLuuuu)    |
```
ggwpez, line 7: Further down you mention `KodaDot` a few times. I think it should be mentioned here as well - if it is part of a business effort.

text/000x-assethub.md
```diff
@@ -0,0 +1,92 @@
+# RFC-0000: Lowering Deposit Requirements on Polkadot Asset Hub
```
ggwpez, line 1:
```suggestion
# RFC-0000: Lowering NFT Deposits on Polkadot and Kusama Asset Hubs
```
better?

text/000x-assethub.md
```diff
@@ -0,0 +1,92 @@
...
+
+|                 |                                                                                                    |
+| --------------- | -------------------------------------------------------------------------------------------------- |
+| **Start Date**  | 2 November 2023                                                                                    |
+| **Description** | A proposal to reduce the minimum deposit required for collection creation on the Polkadot Asset Hub, making it more accessible and affordable for artists. |
```
ggwpez, line 6: Kusama should be mentioned as well.

text/000x-assethub.md
poppyseedDev: The proposed prices have been adjusted so that they are not too low, but still have a reasonable maximum at a historical maximum of prices of DOT and KSM.

burdges:
In theory, all prices should adjust downwards somewhat as more cores become available, due to ongoing optimizations.  We'll need Kusama to waste have it's core on gluttons of course.  We'll spend some cores on system parachain on Polkadot.  Other factors like async backing change the price structure the opposite direction.  etc.  All this has some floor defined by the number & specs of collator, which again depends upon how system parachains wind up being structured.  And elastic scaling releaves that slightly.  And all values increase like `log statesize` too.

All this says:  It's tricky to automate this too much, but it's still nice not to have the conversation too often.

It'd also be nice if there was a for-profit parachain which did specific things like NFTs cheaper than AssetHub.

cc @jonasW3F

text/000x-assethub.md
poppyseedDev: This initiative stemed from an independent effort, seeing how frustrated Luuu was with converting artists to Asset Hub, however I additionally added VikiiVal as a contributor since he is helping in gathering the requested data.

text/000x-assethub.md
poppyseedDev: Hmm, since this is talking more about the benefits for Stakeholders, would it be okay if I better highlight this point in the drawbacks?

vikiival:
Hello @ggwpez,
as I/we minted a majority of NFTs on AssetHub I can provide some statistic


1. Amount of collections created

```sql
select count(metadata) from collection_entity;
-- 40
```

2. Amount of nfts created

```sql
select count(metadata) from nft_entity;
-- 2313
```

3. Average and max length of metadata bytes stored on chain 

```sql
select avg(length(metadata)), max(length(metadata)) from nft_entity where metadata is not null;
-- 54.2
```

4. median of metadata bytes stored on chain 

```sql
SELECT PERCENTILE_DISC(0.5) WITHIN GROUP(ORDER BY length(metadata)) FROM nft_entity;
-- 46
```

5. Distribution of metadata lengths 

```sql
select length(metadata), count(*) from nft_entity where metadata is not null group by length(metadata) order by length(metadata) desc;
-- 74,1
-- 73,3
-- 71,651
-- 66,86
-- 59,33
-- 58,1
-- 53,66
-- 46,1472
```

Regarding attributes on-chain the current utilisation is **0**.

- https://github.com/polkadot-fellows/RFCs/pull/45#discussion_r1384846500



> It would be nice to see these values in relation to what is actually being stored on-chain.
Then we can compare these values to proven spam-resistant storage prices

Let's do checkup on utilization:

1. What is the median of NFTs minted in one collection?

```sql
SELECT PERCENTILE_DISC(0.5) WITHIN GROUP(ORDER BY supply)FROM collection_entity where supply > 0 and metadata is not null;`
-- 4
```

1. What is the average and max of NFTs minted in collection?

```sql
select avg(supply), max(supply) from collection_entity where metadata is not null;
-- 58.65, 1468
```

> ⚠️ Note that all number include both `uniques` and `nfts` pallet.

### What we can do to prevent spam on the AssetHub?

There are few ways to make offer better utilization for the assethub NFTs.
First we can migrate, deprecate and remove `uniques` pallet from the AssetHub chain. This will dramatically decrease the potential use of space on chain. 

Second most important is to decrease effective amount of items from `u32` to `u16`. 

Because `u32` is mapped as `assert_eq!(u32::MAX, 4_294_967_295);` in Rust. It is highly impossible to one will mint 4 billion NFTs per one collection. 

Third point collection could be potentially `forceDestroyed` (not-implemented) by the Open-Gov.

**TL;DR**: There are at least 3 effective ways to prevent 'waste-of-storage', prevent spam, and do effective utilization of the NFTs on AssetHub.

text/000x-assethub.md
vikiival: Fixed in https://github.com/poppyseedDev/RFCs/commit/0e49a4fd5dc3cbe4d4e326a6ede38c91f000da10

ggwpez:
Thanks for the data @vikiival, i normally have no introspection into how it is used. This was very helpful!  

I checked how many bytes it can take at most to create a collection, and it comes in at around 130 byte. So the 10 DOT seems definitely excessive. When i plug in the 130 byte into the `deposit` formula then it comes out around 0.2-0.4 DOT. For the `ItemDeposit` it seems to just be a constant as well, without justification.  
Maybe it would be possible to reduce some of [these constants](https://github.com/polkadot-fellows/runtimes/blob/7bd91e826e4b2d95f6d2ee7f6bb0f0f47d095859/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L716-L723) to lower the normal storage requirement.  

What do you think about the values @joepetrowski, @jsidorenko and @rphmeier?

> First we can migrate, deprecate and remove uniques pallet from the AssetHub chain

Yea we should do this. @jsidorenko do you have this planned?

text/000x-assethub.md
ggwpez: But the Polkadot Community also benefits from a secure network. Otherwise if the state gets bloated like on Moonbeam then it is not good.

jsidorenko:
Re migration, I'm waiting on the new runtime release to stabilise and then will release the web NFTs migratory.

As for the deposits, those enormous deposits don't make any sense and prevent us from having users. 
Last year we had the same issue on Kusama and I decreased the NFTs related deposits 10 times. As a result, we have now 200+ collections created over just 6 months after the release.

text/000x-assethub.md
```diff
@@ -0,0 +1,92 @@
...
+
+This RFC proposes a change to the current deposit requirements on the Polkadot and Kusama Asset Hub for creating NFT collections. The objective is to lower the barrier to entry for artists, fostering a more inclusive and vibrant ecosystem, while maintaining network integrity and preventing spam.
+
+## Motivation
+
+The current deposit of 10 DOT for collection creation on the Polkadot Asset Hub presents a significant financial barrier for many artists. By lowering the deposit requirements, we aim to encourage more artists to participate in the Polkadot NFT ecosystem, thereby enriching the diversity and vibrancy of the community and its offerings.
+
+## Stakeholders
+
+- **Artists**: Primary beneficiaries of the proposed change, particularly those who found the current deposit requirements prohibitive.
```
joepetrowski, line 19:
```suggestion
- **NFT Creators**: Primary beneficiaries of the proposed change, particularly those who found the current deposit requirements prohibitive.
```

text/000x-assethub.md
```diff
@@ -0,0 +1,92 @@
...
+- **KodaDot**: As the facilitator of artists' relations, KodaDot has a vested interest in making the platform more accessible.
```
joepetrowski, line 20: KodaDot may be an example, but I don't think RFCs should target specific entities.
```suggestion
- **NFT Platforms**: As the facilitator of artists' relations, KodaDot has a vested interest in making the platform more accessible.
```

text/000x-assethub.md
```diff
@@ -0,0 +1,92 @@
...
+- **Polkadot Community**: Stands to benefit from an influx of artists, creators and diverse NFT collections, enhancing the overall ecosystem.
+
+Previous discussions have been held within the KodaDot community, as well as with artists expressing their concerns about the deposit amounts. Referencing to [Polkadot Forum conversation](https://forum.polkadot.network/t/polkadot-assethub-high-nft-collection-deposit/4262).
+
+## Explanation
+This RFC suggests modifying the smart contract governing collection creation on the Polkadot Asset Hub to require a lower deposit. The exact amount of the reduced deposit is yet to be determined and should be discussed and agreed upon by the stakeholders. The implementation of this change requires careful consideration of the network's integrity and the prevention of spam, possibly through alternative means such as rate limiting or account verification.
+
+**Prices and Proposed Prices on Polkadot Asset Hub:**
```
joepetrowski, line 28: The explanation lacks discussion that [I was hoping for](https://forum.polkadot.network/t/polkadot-assethub-high-nft-collection-deposit/4262/17?u=joepetrowski) about the tradeoffs and how you arrived at these suggested deposits.

text/000x-assethub.md
```diff
@@ -0,0 +1,92 @@
...
+This change is not expected to have a significant impact on the overall performance of the Polkadot Asset Hub. However, it is crucial to monitor the network closely, especially in the initial stages after implementation, to identify and mitigate any potential issues.
+ 
+Additionally, a supplementary proposal aims to augment the network's adaptability:
+
+> Just from a technical perspective; I think the best we can do is to use a weak governance origin that is controlled by some consortium (ie. System Collective).
+> This origin could then update the NFT deposits any time the market conditions warrant it - obviously while honoring the storage deposit requirements.
+> To implement this, we need RFC#12 and the Parameters pallet from @xlc.
+>  -> OliverTY
+
+This dynamic governance approach would facilitate a responsive and agile economic model for deposit management, ensuring that the network remains both accessible and robust in the face of market volatility.
```
joepetrowski, line 79: I don't really agree with this approach for a few reasons. One, in attempting to provide stability it actually removes it because the DOT-based deposit can change so often that NFT issuers don't even know how much DOT they should acquire in order to plan a few weeks in advance.

Two, it does not scale. There are deposits all over the system. This group could then find themselves perpetually changing tons of deposits.

I don't see any need to update deposits more than every six months, which is well within the realm of runtime upgrades. If deposits should be tied to some fiat value, then on Asset Hub they should just directly declare the deposit in a chosen stablecoin. If the deposits are in DOT, and users want to hold stablecoins, then wallets can easily batch an Asset Conversion swap in front of a creation call.

joepetrowski (commented):
The good: I find the updated deposit suggestions quite reasonable and would accept a change toward those.

The bad: This RFC doesn't really provide structured discussion or generalize the issue. It seems like a dump of a bunch of quotes from the forum thread with a very specific goal of making KodaDot's life easier. While that is a nice goal, the RFC should be more general and discuss the economic, technical, and application tradeoffs of storage deposits.

joepetrowski:
> Second most important is to decrease effective amount of items from `u32` to `u16`.

This seems like an arbitrary imposition. The economics should provide reasonable bounds on usage, not data overflows.

> I checked how many bytes it can take at most to create a collection, and it comes in at around 130 byte. So the 10 DOT seems definitely excessive. When i plug in the 130 byte into the `deposit` formula then it comes out around 0.2-0.4 DOT. For the `ItemDeposit` it seems to just be a constant as well, without justification. Maybe it would be possible to reduce some of [these constants](https://github.com/polkadot-fellows/runtimes/blob/7bd91e826e4b2d95f6d2ee7f6bb0f0f47d095859/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L716-L723) to lower the normal storage requirement.
> 
> The other constants like `itemDeposit` etc are already calculated with the 1/100 storage price of Polkadot...

It looks like those values are just manually set. I'm in favor of generally using the `deposit()` function and calculating the storage footprint of items. This then treats pallets "equally". Maybe there are cases where you want to increase or decrease that, to increase or decrease the barrier to using some component, but those should be the exception. This [function](https://github.com/polkadot-fellows/runtimes/blob/dd7f86f0d50064481ed0b7c0218494a5cfad997e/system-parachains/asset-hubs/asset-hub-polkadot/src/constants.rs#L28) already accounts for its being in a parachain with a deposit lower by a factor of 100.

poppyseedDev:
We changed the PR quite a bit according to the requests.

@joepetrowski what are your thoughts?

text/000x-lowering-deposits-assethub.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Motivation
+
+The current deposit of 10 DOT for collection creation (along with 0.01 DOT for item deposit and 0.2 DOT for metadata and attribute deposit) on the Polkadot Asset Hub and 0.1 KSM on Kusama Asset Hub presents a significant financial barrier for many NFT creators. By lowering the deposit requirements, we aim to encourage more NFT creators to participate in the Polkadot NFT ecosystem, thereby enriching the diversity and vibrancy of the community and its offerings.
+
+The actual implementation of the deposit is an arbitrary number coming from [Uniques pallet](https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs#L757). It is not a result of any economic analysis. This proposal aims to adjust the deposit from constant to dynamic pricing based on the `deposit` function with respect to stakeholders.
```
joepetrowski, line 17: This is a testnet runtime (and the line number on master is already out of date). Can you update to a release tag from the Polkadot runtime?

text/000x-lowering-deposits-assethub.md
```diff
@@ -0,0 +1,197 @@
...
+Previous discussions have been held within the Polkadot Forum community and with artists expressing their concerns about the deposit amounts. [Link](https://forum.polkadot.network/t/polkadot-assethub-high-nft-collection-deposit/4262).
+
+## Explanation
+This RFC proposes a revision of the deposit constants in the nfts pallet on the Polkadot Asset Hub. The new deposit amounts would be determined by a standard deposit formula.
+
+This RFC suggests modifying deposit constants defined in the `nfts` pallet on the Polkadot Asset Hub to require a lower deposit. The reduced deposit amount should be determined by the `deposit` adjusted by the pricing mechanism (arbitrary number/another pricing function). 
+
+### Current code structure
+
+[Current deposit requirements are as follows](https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs#L757)
```
joepetrowski, line 39: Testnet

text/000x-lowering-deposits-assethub.md
```diff
@@ -0,0 +1,197 @@
...
+As we can see in the code definition above the current code does not use the `deposit` funtion when the collection in the following instances: `UniquesCollectionDeposit` and  `UniquesItemDeposit`.
+
+### Proposed Modification Using the Deposit Function
+
+This proposed modification adjusts the deposits to use the `deposit` function instead of using an arbitrary number.
+
+```rust
+parameter_types! {
+	pub const NftsCollectionDeposit: Balance = deposit(1, 130);
+	pub const NftsItemDeposit: Balance = deposit(1, 164);
```
joepetrowski, line 74: How did you generate these inputs? Not saying they are wrong, would just like to see a note about it.

text/000x-lowering-deposits-assethub.md
```diff
@@ -0,0 +1,197 @@
...
+	pub const NftsDepositPerByte: Balance = deposit(0, 1);
+}
+```
+
+Calculations viewed bellow were calculated by using the following repository [rfc-pricing](https://github.com/vikiival/rfc-pricing).
+**Polkadot**
+| **Name**                  | **Current price implementation** | **Proposed Modified by using the new deposit function** |
+|---------------------------|----------------------------------|-------------------------|
+| collectionDeposit         | 10 DOT                           | 0.20064 DOT             |
+| itemDeposit               | 0.01 DOT                         | 0.20081 DOT             |
```
joepetrowski, line 86: Interesting, 20x increase.

text/000x-lowering-deposits-assethub.md
```diff
@@ -0,0 +1,197 @@
...
+| itemDeposit               | 0.001 KSM                        | 0.000167 KSM              |
+| metadataDepositBase       | 0.006709666617 KSM               | 0.0006709666617 KSM       |
+| attributeDepositBase      | 0.00666666666 KSM                | 0.000666666666 KSM        |
+
+
+## Discussion of Other Proposals
+Several innovative proposals have been considered to enhance the network's adaptability and manage deposit requirements more effectively:
+
+### Enhanced Weak Governance Origin Model
+The concept of a weak governance origin, controlled by a consortium like the System Collective, has been proposed. This model would allow for dynamic adjustments of NFT deposit requirements in response to market conditions, adhering to storage deposit norms.
```
joepetrowski, line 137: What is the System Collective?

text/000x-lowering-deposits-assethub.md
```diff
@@ -0,0 +1,197 @@
...
+
+### Linking Deposit to USD(x) Value
+This approach suggests pegging the deposit value to a stable currency like the USD, introducing predictability and stability for network users.
+
+**Considerations and Challenges:**
+- **Market Dynamics**: One perspective is that fluctuations in native currency value naturally balance user participation and pricing, deterring network spam while encouraging higher-value collections. Conversely, there's an argument for allowing broader participation if the DOT/KSM value increases.
+- **Complexity and Risks**: Implementing a USD-based pricing system could add complexity and potential risks. The implementation needs to be carefully designed to avoid unintended consequences, such as excessive reliance on external financial systems or currencies.
+
+Each of these proposals offers unique advantages and challenges. The optimal approach may involve a combination of these ideas, carefully adjusted to address the specific needs and dynamics of the Polkadot and Kusama networks.
```
joepetrowski, line 158: I think we can probably do a combination of these two. It'd require some implementation, so switching to `deposit()` would be phase one, and then we could do something like:

- Function 1: Deposit (in DOT) as a function of number of collections.
- Function 2: Fixed USD(x) cost as a ceiling.
- Actual deposit: `min(F1, F2)`

joepetrowski (commented):
Generally in agreement with the direction, but a few questions:

text/000x-lowering-deposits-assethub.md
vikiival: Hey sure!

https://github.com/vikiival/rfc-pricing

text/000x-lowering-deposits-assethub.md
poppyseedDev: That's why we propose a further reduction for _itemDeposit_ in the next paragraph by 40 
```rust 
pub const NftsItemDeposit: Balance = deposit(1, 164) / 40;
```

text/000x-lowering-deposits-assethub.md
poppyseedDev:
```suggestion
The actual implementation of the deposit is an arbitrary number coming from [Uniques pallet]
(https://github.com/paritytech/polkadot-sdk/blob/afadb830e0a38365a5afec36fd1b2c77896a0a85/cumulus/parachains/runtimes/assets/asset-hub-rococo/src/lib.rs#L756-L762). It is not a result of any economic analysis. This proposal aims to adjust the deposit from constant to dynamic pricing based on the `deposit` function with respect to stakeholders.
```

poppyseedDev: Would that be branch `release-polkadot-v1.6.0` ?

poppyseedDev:
```suggestion
The actual implementation of the deposit is an arbitrary number coming from [Uniques pallet](https://github.com/polkadot-fellows/runtimes/blob/7515fc0385f8d48f2d08b2182a24174148e70135/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L686-L692). It is not a result of any economic analysis. This proposal aims to adjust the deposit from constant to dynamic pricing based on the `deposit` function with respect to stakeholders.
```

poppyseedDev: Or on [polkadot fellowship](https://github.com/polkadot-fellows/runtimes)?

text/000x-lowering-deposits-assethub.md
poppyseedDev:
```suggestion
[Current deposit requirements are as follows](https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L686-L692)
```

text/000x-lowering-deposits-assethub.md
poppyseedDev: https://forum.polkadot.network/t/polkadot-assethub-high-nft-collection-deposit/4262/16

text/000x-lowering-deposits-assethub.md
joepetrowski: Yeah exactly, the Fellowship one

text/000x-lowering-deposits-assethub.md
joepetrowski: Yeah, but with a commit instead of `main`.

poppyseedDev:
```suggestion
[Current deposit requirements are as follows](https://github.com/polkadot-fellows/runtimes/commit/c2a3c9881c0ee6393d1665fefd9671d040215188#diff-32714ab945a24cbcc5979fd9065e0c13e0beb59f7533b38e2e97a872b23cb125R596-R600)
```

poppyseedDev: so I guess like this?

joepetrowski: yeah

text/000x-lowering-deposits-assethub.md
muharem: I'm curious about the motivation behind increasing the deposit based on the number of collections. If the proposed floor deposits are too low to prevent spam, in theory, the function will adjust the deposit value to a more reasonable one and continue to grow, potentially raising a barrier for creators again.

This leads me to think that the primary challenge is to find an optimal deposit value, which depends on the DOT/KSM price and our current technology state (as mentioned by @burdges [here](https://github.com/polkadot-fellows/RFCs/pull/45#issuecomment-1798950680)). If there is no optimal value, we will need to explore technical solutions.

If what I've said withstands criticism, we could have a scale factor parameter that can be changed through a referendum and deposits could also be linked to the USD price. Additionally, we should explore and learn how to estimate such an optimal deposit value.

vikiival:
@joepetrowski as per commit 3ad68dd73658c8649c3f8d635f987346297b0078

can we keep the deposit as 0.01 DOT per item deposit?

joepetrowski:
> can we keep the deposit as 0.01 DOT per item deposit?

It was 0.005 before, you want it back to 0.01?

vikiival:
> It was 0.005 before, you want it back to 0.01?

Ah, I must have misread the `git diff`

poppyseedDev:
Thank you so much @joepetrowski for your changes!!

text/0045-nft-deposits-asset-hub.md
```diff
@@ -0,0 +1,246 @@
...
+In the long term, an implementation should account for what should happen to the deposit rates
+assuming that the subsidy is successful and attracts a lot of deployments. Many options are
+discussed in the [Addendum](#addendum).
+
+The deposit should be calculated as a function of the number of existing collections with maximum
+DOT and stablecoin values limiting the amount. With asset rates available via the Asset Conversion
+pallet, the system could take the lower value required. A sigmoid curve would make sense for this
+application to avoid sudden rate changes, as in:
+
+$$ \frac{\mathrm{min(DOT deposit, stable deposit)} }{\mathrm{1 + e^{a - b * x}} }$$
```
ggwpez, line 135: Yea LGTM. Made some [script](https://github.com/ggwpez/substrate-scripts/blob/a0799650caaf478d7bf1eaf8961f561cf070c1a1/nft-deposit-curve.py) to inspect the params.    
</img>

text/0045-nft-deposits-asset-hub.md
```diff
@@ -0,0 +1,246 @@
...
+
+As noted above, state bloat is a security concern. In the case of abuse, governance could adapt by
+increasing deposit rates and/or using `forceDestroy` on collections agreed to be spam.
+
+## Performance, Ergonomics, and Compatibility
+
+### Performance
+
+The primary performance consideration stems from the potential for state bloat due to increased
+activity from lower deposit requirements. It's vital to monitor and manage this to avoid any
```
ggwpez, line 180: Do we have monitoring in place?

ggwpez (approved):

text/0045-nft-deposits-asset-hub.md
```diff
@@ -116,6 +116,28 @@ This adjustment would result in the following DOT and KSM deposit values:
...
+In the long term, an implementation should account for what should happen to the deposit rates
+assuming that the subsidy is successful and attracts a lot of deployments. Many options are
+discussed in the [Addendum](#addendum).
+
+The deposit should be calculated as a function of the number of existing collections with maximum
+DOT and stablecoin values limiting the amount. With asset rates available via the Asset Conversion
+pallet, the system could take the lower value required. A sigmoid curve would make sense for this
+application to avoid sudden rate changes, as in:
+
+$$ \frac{\mathrm{min(DOT deposit, stable deposit)} }{\mathrm{1 + e^{a - b * x}} }$$
```
poppyseedDev, line 135:
```suggestion
$$ startdeposit + \frac{\mathrm{min(DOTdeposit, stabledeposit) - startdeposit} }{\mathrm{1 + e^{b * a - b * x}} }$$
```

text/0045-nft-deposits-asset-hub.md
poppyseedDev: @ggwpez I did a PR ([link](https://github.com/ggwpez/substrate-scripts/pull/4)) to your code so that the function is a bit adjusted to reflect the changes in the function below.

https://github.com/polkadot-fellows/RFCs/assets/30662672/d53f4b7c-f011-4ae2-84e1-e63f20550891

ggwpez: Perfect, yea we can use this to find good params for making it auto-adjust. Thanks!

text/0045-nft-deposits-asset-hub.md
```diff
@@ -0,0 +1,246 @@
+# RFC-0045: Lowering NFT Deposits on Asset Hub
+
+|                 |                                                                               |
+| --------------- | ----------------------------------------------------------------------------- |
+| **Start Date**  | 2 November 2023                                                               |
+| **Description** | A proposal to reduce the minimum deposit required for collection creation on the Polkadot and Kusama Asset Hubs. |
+| **Authors**     | [Aurora Poppyseed](https://github.com/poppyseedDev), [Just_Luuuu](https://github.com/justLuuuu), [Viki Val](https://github.com/vikiival) |
```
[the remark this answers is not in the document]
joepetrowski: I will leave it to @poppyseedDev , I was happy to help out but the initiative came from her.

text/0045-nft-deposits-asset-hub.md
```diff
@@ -0,0 +1,246 @@
...
+a value based on the `deposit` function, which calculates a deposit based on the number of keys
+introduced to storage and the size of corresponding values stored.
+
+Further, it suggests a direction for a future of calculating deposits variably based on adoption
+and/or market conditions. There is a discussion on tradeoffs of setting deposits too high or too
+low.
+
+### Requirements
+
+- Deposits SHOULD be derived from `deposit` function, adjusted by correspoding pricing mechansim.
```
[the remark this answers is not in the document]
joepetrowski: > The terms "MUST", "MUST NOT", "SHALL", "SHALL NOT", or "REQUIRED" mean that the requirement is fixed and must be adhered to by implementations. These statements should be limited to those required for interoperability and security.
> The terms "SHOULD", "RECOMMENDED", "SHOULD NOT", or "NOT RECOMMENDED" mean that there are only limited valid circumstances in which a requirement may be ignored.

I would lean toward "SHOULD" here because it's not like a message format or protocol that other systems depend on. There _may be_ cases in the future where some deposit item is not based on such a function, and it's certainly not a technical limitation that other parts of the runtime depend on.

But either one works for me, I think this is a case where the both are OK.

text/0045-nft-deposits-asset-hub.md
```diff
@@ -0,0 +1,246 @@
...
+This RFC proposes a revision of the deposit constants in the configuration of the NFTs pallet on the
+Polkadot Asset Hub. The new deposit amounts would be determined by a standard deposit formula.
+
+As of v1.1.1, the Collection Deposit is 10 DOT and the Item Deposit is 0.01 DOT (see
+[here](https://github.com/polkadot-fellows/runtimes/blob/v1.1.1/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L687)).
+
+Based on the storage footprint of these items, this RFC proposes changing them to:
+
+```rust
+pub const NftsCollectionDeposit: Balance = system_para_deposit(1, 130);
```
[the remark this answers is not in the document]
joepetrowski: If there are two items then we should probably be consistent. It should still be a large reduction (from 10 to 0.4 instead of 0.2).

Broadly speaking though, and w/r/t the idea of finding an "optimal" deposit, it can never be perfect. As long as you can't start adding more storage items for free (i.e. deposit is directly proportional to storage items added) then it's well-designed. Factors of 2x are just not significant (and it's a bit arrogant to think we can get it so precise). I think probably 10x is the best we can hope for with these parameters.

text/0045-nft-deposits-asset-hub.md
joepetrowski: Maybe @karim or someone from KodaDot is tracking NFT activity?

text/0045-nft-deposits-asset-hub.md
joepetrowski: Ah, I see what you are doing. But you don't need `b * a` because its evaluation is implicit in the value of `a`.

text/0045-nft-deposits-asset-hub.md
```diff
@@ -0,0 +1,246 @@
...
+Kusama and Polkadot Asset Hubs, making Polkadot and Kusama more accessible and user-friendly.
+
+### Compatibility
+
+The change does not impact compatibility as a `redeposit` function is already implemented.
+
+## Unresolved Questions
+
+If this RFC is accepted, there should not be any unresolved questions regarding how to adapt the
+implementation of deposits for NFT collections.
```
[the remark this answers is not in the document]
vikiival: I think that @joepetrowski perfectly summarised point of this RFC 

> The plan presented above is recommended as an immediate step

Other strategies can be done as a follow-up 🚀

text/0045-nft-deposits-asset-hub.md
poppyseedDev: I think @vikiival here is in a good position to answer this question, since he created indexer for AssetHub.

text/0045-nft-deposits-asset-hub.md
poppyseedDev: Okay but then the code implementor has to know that a = x0 * b, where x0 is the threshold level of collections or items.

text/0045-nft-deposits-asset-hub.md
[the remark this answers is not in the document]
ggwpez: Yea okay. Its not an implementation detail anyway.

text/0045-nft-deposits-asset-hub.md
vikiival: > Someone from KodaDot is tracking NFT activity?

```sql
-- current nfts 9792
select count(*) from nft_entity where burned = false;

-- current collection 98
select count(*) from collection_entity where burned = false;

-- all nfts 9883 (~ 100 burned)
select count(*) from nft_entity;

-- all collections 138 (40 destroyed)
select count(*) from collection_entity;

-- all interactions that is not mint (list, buy, burn) 16_019
select count(*) from event where interaction <> 'MINT';
```

Top 3 creators on AHP
```json
  [{
    "count": 41,
    "issuer": "15CoYMEnJhhWHvdEPXDuTBnZKXwrJzMQdcMwcHGsVx5kXYvW"
  },
  {
    "count": 24,
    "issuer": "1ULZhwpUPLLg5VRYiq6rBHY8XaShAmBW7kqGBfvHBqrgBcN"
  },
  {
    "count": 6,
    "issuer": "1nYXo6fd1fjVm1ixUwycELfjTyaxxzVHAVgZ8S8H6cZMfY7"
  }]
  ```
  
Data are public via our @subsquid indexer or [Typescript API (kodadot/uniquery)](https://github.com/kodadot/uniquery)

poppyseedDev:
Hey @ggwpez I was told that we need someone with rank above 3 to approve this PR so that the voting can start.

Could you help us with that?

ggwpez:
>Hey ggwpez could you tell us what are the next steps with this PR?

We will soon propose this for on-chain voting 🙏

joepetrowski (approved):

joepetrowski:
/rfc process 0x3ac549f24ecf451eb4d45f16f4fff694b0a103f8006ff9ab7f16c484bcfb535c

anaelleltd:
@poppyseedDev, do you have an update on the status of this RFC?
Thanks.

joepetrowski:
@anaelleltd it's done and deployed
https://github.com/polkadot-fellows/runtimes/pull/237

PR #46, opened 2023-11-08
Metadata for offline signers

Slesarew:
Add a digest of metadata to signed extensions as mentioned here https://github.com/paritytech/polkadot-sdk/issues/2016

### DIFF
--- a/text/0000-metadata-for-offline-signers.md
+++ b/text/0000-metadata-for-offline-signers.md
@@ -1,0 +1,137 @@
+# RFC-0000: Metadata for offline signers
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 2023-10-31                                                                                  |
+| **Description** | Add SignedExtension to check Metadata Root Hash                                             |
+| **Authors**     | Alzymologist Oy, Zondax LLC, Parity GmbH                                                    |
+
+## Summary
+
+Add a metadata digest value (33-byte constant within fixed `spec_version`) to Signed Extensions to supplement signer party with proof of correct extrinsic interpretation. The digest value is generated once before release and is well-known and deterministic. The digest mechanism is designed to be modular and flexible and to support partial metadata transfer as needed by the signing party's extrinsic decoding mechanism and taking into account signing devices potentially limited communication bandwidth and memory capacity.
+
+## Motivation
+
+### Background
+
+While all blockchain systems support (at least in some sense) offline signing used in air-gapped wallets and lightweight embedded devices, only few allow simultaneously complex upgradeable logic and full message decoding on the cold off-line signer side; Substrate is one of these heartening few, and therefore - we should build on this feature to greatly improve transaction security, and thus in general, network resilience.
+
+As a starting point, it is important to recognise that prudence and due care are naturally required. As we build further reliance on this feature we should be very careful to make sure it works correctly every time so as not to create false sense of security.
+
+In order to enable decoding that is small and optimized for chain storage transactions, a metadata entity is used, which is not at all small in itself (on the order of half-MB for most networks). This is a dynamic data chunk which completely describes chain interfaces and properties that could be made into a portable scale-encoded string for any given network version and passed along into an off-chain device to familiarize it with latest network updates. Of course, compromising this metadata anywhere in the path could result in differences between what user sees and signs, thus it is essential that we protect it.
+
+Therefore, we have 2 problems to be solved:
+
+1. Metadata is large, takes long time to be passed into a cold storage device with memory insufficient for its storage; metadata SHOULD be shortened and transmission SHOULD be optimized.
+2. Metadata authenticity SHOULD be ensured.
+
+As of now, there is no working solution for (1), as the whole metadata has to be passed to the device. On top of this, the solution for (2) heavily relies on a trusted party managing keys and ensuring metadata is indeed authentic: creating poorly decentralized points of potential failure.
+
+### Solution requirements
+
+#### Include metadata digest into signature
+
+Some cryptographically strong digest of metadata MUST be included into signable blob. There SHALL NOT be storage overhead for this blob, nor computational overhead, on the node side; thus MUST be a constant within given runtime version, deterministically defined by metadata.
+
+ - Metadata information that could be used in signable extrinsic decoding MUST be be included in digest;
+ - Digest MUST be deterministic with respect to metadata;
+ - Digest MUST be cryptographically strong against pre-image, both first and second;
+ - Extra-metadata information necessary for extrinsic decoding and constant within runtime version MUST be included in digest;
+ - Digest format SHOULD be versioned to allow rapid withdrawal of cold signing devices in case severe security vulnerability is found in shortener mechanism;
+ - Work necessary for proving metadata authenticity MAY be omitted at discretion of signer device design (to support automation tools).
+
+#### Reduce metadata size
+
+Metadata should be stripped from parts that are not necessary to parse a signable extrinsic, then it should be separated into a finite set of self-descriptive chunks. Thus, a subset of chunks necessary for signable extrinsic decoding and rendering could be sent, possibly in small portions (ultimately - one at a time), to cold device together with proof.
+
+ - Single chunk with proof payload size SHOULD fit within few kB;
+ - Chunks handling mechanism SHOULD support chunks being sent in any order without memory utilization overhead;
+ - Unused enum variants MUST be stripped (this has great impact on transmitted metadata size; examples: era enum, enum with all calls for call batching).
+
+## Stakeholders
+
+This feature is essential for **all** offline signer tools; many regular signing tools might make use of it. In general, this RFC greatly improves security of any network implementing it, as many governing keys are used with offline signers.
+
+Implementing this RFC would remove requirement to maintain metadata portals manually, as task of metadata verification would be effectively moved to consensus mechanism of the chain.
+
+## Explanation
+
+Detailed description of metadata shortening and digest process is provided in [metadata-shortener](https://github.com/Alzymologist/metadata-shortener) crate (see `cargo doc --open` and examples).
+
+### Metadata descriptor
+
+Values for metadata shortening protocol version, `ExtrinsicMetadata`, SCALE-encoded `spec_version` and `spec_name` Strings, SCALE-encoded base58 prefix, SCALE-encoded decimals value, SCALE-encoded token unit String, should be prepared and combined as metadata descriptor.
+
+### Metadata modularization
+
+1. Types registry is stripped from `docs` fields.
+2. Types records are separated into chunks, with enum variants being individual chunks differing by variant index; each chunk consisting of `id` (same as in full metadata registry) and SCALE-encoded 'Type' description (reduced to 1-variant enum for enum variants)
+
+### Digest
+
+1. Blake3 hash is computed for each chunk of modular short metadata registry.
+2. Hashes are sorted and constructed into static Merkle tree as implemented in `merkle_cbt` crate using blake3 digest of concatenated child nodes values for merging.
+3. Root hash of this tree is merged with metadata descriptor blake3 hash; this is metadata digest.
+
+Product of concatenation of porotocol version number with resulting metadata digest MUST be included into Signed Extensions
+
+### Shortening
+
+Shortened metadata is produced as sorted subset of modular metadata chunks selected from full set using protocol `cut_metadata` proposed in [metadata-shortener](https://github.com/Alzymologist/metadata-shortener) crate or equivalent operation that provides sufficient information for both decoding of extrinsic.
+
+### Transmission
+
+Shortened metadata chunks MAY be trasmitted into cold device together with Merkle proof in its entirety or in parts, depending on memory capabilities of the cold device and it ability to reconstruct larger fraction of tree. This document does not specify the manner of transmission. The order of metadata chunks MAY be arbitrary, the only requirement is that indices of nodes corresponding to chunks MUST be communicated.
+
+### Offline verification
+
+The transmitted metadata chunks are hashed together with proof lemmas to obtain root that MAY be transmitted along with the rest of payload. Verification that the root transmitted with message matches with calculated root is optional; the transmitted root SHOULD NOT be used in signature, calculated root MUST be used; however, there is no mechanism to enforce this - it should be done during cold signers code audit.
+
+### Chain verification
+
+The root of metadata computed by cold device MUST be included into Signed Extensions; this way the transaction will pass as valid iff hash of metadata as seen by cold storage device is identical to consensus hash of metadata, ensuring fair signing protocol.
+
+## Drawbacks
+
+### Increased transaction size
+
+Depending on implementation details, an extra byte may be needed to indicate whether the new version of metadata verification was used; this may be needed during transaction period, or the same byte may store the version of metadata hashing protocol
+
+### Transition overhead
+
+Some slightly out of spec systems might experience breaking changes as new content of signed extensions is added. It is important to note, that there is no real overhead in processing time nor complexity, as the metadata checking mechanism is voluntary. The only drawbacks are expected for tools that do not implement MetadataV14 self-descripting features.
+
+## Testing, Security, and Privacy
+
+The metadata shortening protocol should be extensively tested on all available examples of metadata before releasing changes to either metadata or shortener. Careful code review should be performed on shortener implementation code to ensure security. The main metadata tree would inevitably be constructed on runtime build which would also ensure correctness.
+
+To be able to recall shortener protocol in case of vulnerability issues, a version byte is included.
+
+## Performance, Ergonomics, and Compatibility
+
+### Performance
+
+This is negligibly short pessimization during build time on the chain side. Cold wallets performance would improve mostly as metadata validity mechanism that was taking most of effort in cold wallet support would become trivial.
+
+### Ergonomics
+
+The proposal was optimized for cold storage wallets usage with minimal impact on all other parts of the ecosystem
+
+### Compatibility
+
+Proposal in this form is not compatible with older toold that do not implement proper MetadataV14 self-descriptive features; those would have to be upgraded to include a new signed extensions field.
+
+## Prior Art and References
+
+This project was developed as Polkadot Treasury grant; relevant development links are located in [metadata-offline-project](https://github.com/Alzymologist/metadata-offline-project) repository.
+
+## Unresolved Questions
+
+1. Should hash inclusion bit be added to signed extensions?
+2. How would polkadot-js handle the transition?
+3. Where would non-rust tools like Ledger apps get shortened metadata content?
+
+## Future Directions and Related Material
+
+Changes to code of all cold signers to implement this mechanism SHOULD be done when this is enabled; non-cold signers may perform extra metadata check for better security. Ultimately, signing anything without decoding it with verifiable metadata should become discouraged in all situations where a decision-making mechanism is involved (that is, outside of fully automated blind signers like trade bots or staking rewards payout tools).
+


text/0000-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Stakeholders
+
+This feature is essential for **all** offline signer tools; many regular signing tools might make use of it. In general, this RFC greatly improves security of any network implementing it, as many governing keys are used with offline signers.
+
+Implementing this RFC would remove requirement to maintain metadata portals manually, as task of metadata verification would be effectively moved to consensus mechanism of the chain.
+
+## Explanation
+
+Detailed description of metadata shortening and digest process is provided in [metadata-shortener](https://github.com/Alzymologist/metadata-shortener) crate (see `cargo doc --open` and examples).
```
tomaka, line 59: The point of writing an RFC is to describe the design in plain English *here*.
You can't just link to a third party repository that you modify whenever you want. The fellowship votes to accept specific design details.

text/0000-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,137 @@
...
+Values for metadata shortening protocol version, `ExtrinsicMetadata`, SCALE-encoded `spec_version` and `spec_name` Strings, SCALE-encoded base58 prefix, SCALE-encoded decimals value, SCALE-encoded token unit String, should be prepared and combined as metadata descriptor.
+
+### Metadata modularization
+
+1. Types registry is stripped from `docs` fields.
+2. Types records are separated into chunks, with enum variants being individual chunks differing by variant index; each chunk consisting of `id` (same as in full metadata registry) and SCALE-encoded 'Type' description (reduced to 1-variant enum for enum variants)
+
+### Digest
+
+1. Blake3 hash is computed for each chunk of modular short metadata registry.
```
tomaka, line 72: Given that Polkadot and Substrate use blake2 basically everywhere, is there a reason to prefer blake3?
We're not going to migrate from blake2 to blake3 any time soon, so whatever program implements this would most likely need to support both blake2 and blake3.

text/0000-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,137 @@
...
+2. Hashes are sorted and constructed into static Merkle tree as implemented in `merkle_cbt` crate using blake3 digest of concatenated child nodes values for merging.
```
tomaka, line 73: Same remark. We want the format to be clear and precise, we can't say "we'll use whatever format this third party library does". There are implementers that don't even use Rust.

bkchr: Yeah, I was also expecting that we get the details/spec of how the merkelization works.

text/0000-metadata-for-offline-signers.md
Slesarew: Sure, I'll be adding the detailed description soon. This is first draft mostly for the rest of the team to view.

text/0000-metadata-for-offline-signers.md
Slesarew: The two bottleneck tools support blake3 nicely (I understand that Ledger has hardware support or something comparably low-footprint), it is somewhat faster than blake2 even on single core, it is better defined, and we expect this standard to outlive several Metadata revisions ideally (as it has so much fewer features than metadata) thus it might end up living long enough to see ecosystem transition. We've discussed this overhead with Zondax and both teams see this as advantage in our cold devices; adding blake3 to Polkadot Vault is trivial, and we have no other tools that need this functionality at the moment - best guess is that if Ledger and Kampela can run both blakes, everything that's released later sure would be able too.

text/0000-metadata-for-offline-signers.md
Slesarew: Of course, I'll be adding algorithm to spec soon.

text/0000-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,150 @@
...
+
+Implementing this RFC would remove requirement to maintain metadata portals manually, as task of metadata verification would be effectively moved to consensus mechanism of the chain.
+
+## Explanation
+
+Detailed description of metadata shortening and digest process is provided in [metadata-shortener](https://github.com/Alzymologist/metadata-shortener) crate (see `cargo doc --open` and examples). Below are presented algorithms of the process.
+
+### Metadata descriptor
+
+Values for metadata shortening protocol version, `ExtrinsicMetadata`, SCALE-encoded `spec_version` and `spec_name` Strings, SCALE-encoded base58 prefix, SCALE-encoded decimals value, SCALE-encoded token unit String, should be prepared and combined as metadata descriptor.
```
carlosala, line 63: `base58prefix` is encoded as u16 and `decimals` as u8 in the reference implementation. It seems that fixed-width uints work better here since their size is known.
base58prefix is u16 from spec and using more than 255 decimal values is not rational since max value for u256 is 10e78.

text/0000-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Metadata modularization
+
+1. Types registry is stripped from `docs` fields.
+2. Types records are separated into chunks, with enum variants being individual chunks differing by variant index; each chunk consisting of `id` (same as in full metadata registry) and SCALE-encoded 'Type' description (reduced to 1-variant enum for enum variants).
```
carlosala, line 68: Would be great to mention that 0-variant enums are treated as regular types (as 1-variant enums).

text/0000-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,150 @@
...
+4. Right node and then left node is popped from the front of the nodes queue and merged; the result is sent to the end of the queue.
+5. Step (4) is repeated until only one node remains; this is tree root.
+
+### Digest
+
+1. Blake3 hash is computed for each chunk of modular short metadata registry.
+3. Complete Binary Merkle Tree is constructed as described above.
+4. Root hash of this tree (left) is merged with metadata descriptor blake3 hash (right); this is metadata digest.
+
+Product of concatenation of porotocol version number with resulting metadata digest MUST be included into Signed Extensions.
```
carlosala, line 89:
```suggestion
Product of concatenation of protocol version number with resulting metadata digest MUST be included into Signed Extensions.
```

text/0000-metadata-for-offline-signers.md
Slesarew:
```suggestion
Values for metadata shortening protocol version, `ExtrinsicMetadata`, SCALE-encoded `spec_version` and `spec_name` Strings, `u16` base58 prefix, `u8` decimals value, SCALE-encoded token unit String, should be prepared and combined as metadata descriptor.
```

text/0000-metadata-for-offline-signers.md
Slesarew:
```suggestion
2. Types records are separated into chunks, with enum variants being individual chunks differing by variant index; each chunk consisting of `id` (same as in full metadata registry) and SCALE-encoded 'Type' description (reduced to 1-variant enum for enum variants). Enums with 0 variants are treated as regular types.
```

Slesarew:
> It'd be great to define how lemmas and indices are sorted when computing `ShortMetadata`.

Isn't it outside of RFC scope? This is not related to chain behaviour, it's purely on cold protocol side. We can spec it out elsewhere.

carlosala:
> Isn't it outside of RFC scope? This is not related to chain behaviour, it's purely on cold protocol side. We can spec it out elsewhere.

Yep, it's probably outside of the chain's scope. Let's keep it in mind, it's actually important to have a standardized way of transmitting data.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,150 @@
...
+### Solution requirements
+
+#### Include metadata digest into signature
+
+Some cryptographically strong digest of metadata MUST be included into signable blob. There SHALL NOT be storage overhead for this blob, nor computational overhead, on the node side; thus MUST be a constant within given runtime version, deterministically defined by metadata.
+
+ - Metadata information that could be used in signable extrinsic decoding MUST be be included in digest;
+ - Digest MUST be deterministic with respect to metadata;
+ - Digest MUST be cryptographically strong against pre-image, both first and second;
+ - Extra-metadata information necessary for extrinsic decoding and constant within runtime version MUST be included in digest;
```
xlc, line 39: I don't 100% understand this. Do you mean something like:
Additional information (on top of the metadata) that's required to decode extrinsics and runtime constants for a given runtime version must be included in the digest?
But what are those information? Shouldn't metadata itself includes everything?

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,150 @@
...
+ - Digest format SHOULD be versioned to allow rapid withdrawal of cold signing devices in case severe security vulnerability is found in shortener mechanism;
```
xlc, line 40: This is a requirement, and a solution, put together. There are many alternative solutions to this particular issue. There are also other reasons for versioning.
I suggest to only describe the requirements, without linking it to a particular solution.

xlc (commented):
I don't think I will be able to reimplement this work purely using the details provided by this RFC. Maybe more pseudo code will help.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,150 @@
...
+ - Work necessary for proving metadata authenticity MAY be omitted at discretion of signer device design (to support automation tools).
+
+#### Reduce metadata size
+
+Metadata should be stripped from parts that are not necessary to parse a signable extrinsic, then it should be separated into a finite set of self-descriptive chunks. Thus, a subset of chunks necessary for signable extrinsic decoding and rendering could be sent, possibly in small portions (ultimately - one at a time), to cold device together with proof.
```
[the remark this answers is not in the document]
Slesarew: There should be all varieties of options ideally, to fit growing metadata into as small footprint systems as possible. There should also be options to send it in bulk data-efficient format where transmission channel is limiting. This will ensure that we are future-compatible with unpredictably developing hardware (well, and software) systems. Proposed solution solves this challenge.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,150 @@
...
+
+Implementing this RFC would remove requirement to maintain metadata portals manually, as task of metadata verification would be effectively moved to consensus mechanism of the chain.
+
+## Explanation
+
+Detailed description of metadata shortening and digest process is provided in [metadata-shortener](https://github.com/Alzymologist/metadata-shortener) crate (see `cargo doc --open` and examples). Below are presented algorithms of the process.
+
+### Metadata descriptor
+
+Values for metadata shortening protocol version, `ExtrinsicMetadata`, SCALE-encoded `spec_version` and `spec_name` Strings, `u16` base58 prefix, `u8` decimals value, SCALE-encoded token unit String, should be prepared and combined as metadata descriptor.
```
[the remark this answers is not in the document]
Slesarew: Chains without token would naturally have empty fields for token name and zero as unit value. This is how Polkadot Vault handles this now, nobody came up with better solution. These are not contained in metadata so must be transferred separately. It is not possible to decode a transaction without this information in Polkadot network. If you have better solution that would support:
1. chains with units
2. chains without units
3. chains with multiple units
and be simple enough, let's discuss and integrate it.

text/0046-metadata-for-offline-signers.md
Slesarew:
```suggestion
 - It SHOULD be possible to quickly withdraw offline signing mechanism without access to cold signing devices;
 - Digest format SHOULD be versioned.
```

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,150 @@
...
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 2023-10-31                                                                                  |
+| **Description** | Add SignedExtension to check Metadata Root Hash                                             |
+| **Authors**     | Alzymologist Oy, Zondax LLC, Parity GmbH                                                    |
+
+## Summary
+
+Add a metadata digest value (33-byte constant within fixed `spec_version`) to Signed Extensions to supplement signer party with proof of correct extrinsic interpretation. The digest value is generated once before release and is well-known and deterministic. The digest mechanism is designed to be modular and flexible and to support partial metadata transfer as needed by the signing party's extrinsic decoding mechanism and taking into account signing devices potentially limited communication bandwidth and memory capacity.
```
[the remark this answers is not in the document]
Slesarew:
```suggestion
Add a metadata digest value (33-byte constant within fixed `spec_version`) to Signed Extensions to supplement signer party with proof of correct extrinsic interpretation. The digest value is generated once before release and is well-known and deterministic. The digest mechanism is designed to be modular and flexible. It also supports partial metadata transfer as needed by the signing party's extrinsic decoding mechanism. This considers signing devices potentially limited communication bandwidth and/or memory capacity.
```

text/0046-metadata-for-offline-signers.md
xlc: If metadata does not contain all the necessary metadata, I will suggest just fix the metadata. I raised a somewhat related issue before https://github.com/paritytech/substrate/issues/3686

tomaka:
I've now re-read this RFC multiple times.
I know what a trie is, I know what a Merkle proof is and the details of the format of a Merkle proof in Substrate, and I know what the metadata is and what is found inside.
Yet [as I've already commented](https://github.com/polkadot-fellows/RFCs/pull/46#discussion_r1387581546), I don't see how one could possibly implement either generating or parsing this compact metadata by reading this RFC alone.

This RFC uses a lot of terminology that I have no idea what they mean (either generally or in this context), such as "token unit String", "metadata descriptor", "modularization", "merging protocol", or "shortening". All these terms should really be introduced or explained.

In the middle of the RFC there's suddenly the word "leaves", but I don't think I've seen where it ever introduces the fact that there's a trie. I understand that the metadata is split in a trie, but the reason why I understand that is because I've followed in the past the development of this feature, because I don't think that's even mentioned in the RFC.

It seems that this RFC is just a collection of notes that were written down while the implementation was taking place, rather than an actual explanation of the design. It seems to assume that the reader is already familiar with the implementation, and in general makes no effort to be accessible.

text/0046-metadata-for-offline-signers.md
Slesarew: This is a nice solution; however, for things like "units" we'll need more robust automation, those are not arbitrary values, but items declared in runtime. We'll need to make them properly coded part of metadata, otherwise parachain upgrading will end up in mess.

I'm all for fixing metadata issues, but I think this should be done one at a time. This particular issue that units are not mentioned anywhere was discussed for a long time and no effective solution was proposed. I think we should move on with this project and then, if there is a proper way to include units in metadata for all chains, we'll keep wasting couple byte in transfer protocol for empty values and then eventually remove this extra information if an upgrade to shortened metadata happens.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,170 @@
...
+
+As of now, there is no working solution for (1), as the whole metadata has to be passed to the device. On top of this, the solution for (2) heavily relies on a trusted party managing keys and ensuring metadata is indeed authentic: creating poorly decentralized points of potential failure.
+
+### Solution requirements
+
+#### Include metadata digest into signature
+
+Some cryptographically strong digest of metadata MAY be included into signable blob. There SHALL NOT be storage overhead for this blob, nor computational overhead, on the node side; thus MUST be a constant within given runtime version, deterministically defined by metadata.
+
+ - Metadata information that could be used in signable extrinsic decoding MAY be be included in digest, its inclusion MUST be indicated in signed extensions;
```
agryaznov, line 36:
```suggestion
 - Metadata information that could be used in signable extrinsic decoding MAY be included in digest, its inclusion MUST be indicated in signed extensions;
```

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,170 @@
...
+6. `u8` decimals value or `0u8` if no units are defined,
+7. SCALE-encoded token unit String or empty string if no base units are defined,
+
+constitute metadata descriptor. This is minimal information sufficient to decode any signable transaction. Proof that the signing device was presented with genuine subset of Metadata Descriptor should be validated by chain.
+
+### Metadata modularization
+
+1. Types registry is stripped from `docs` fields.
+2. Types records are separated into chunks, with enum variants being individual chunks differing by variant index; each chunk consisting of `id` (same as in full metadata registry) and SCALE-encoded 'Type' description (reduced to 1-variant enum for enum variants). Enums with 0 variants are treated as regular types.
+3. Chunks are sorted by `id` in accending order; chunks with same `id` are sorted by enum vainant index in accending order.
```
agryaznov, line 82:
```suggestion
3. Chunks are sorted by `id` in ascending order; chunks with same `id` are sorted by enum variant index in ascending order.
```

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,170 @@
...
+
+This is negligibly short pessimization during build time on the chain side. Cold wallets performance would improve mostly as metadata validity mechanism that was taking most of effort in cold wallet support would become trivial.
+
+### Ergonomics
+
+The proposal was optimized for cold storage wallets usage with minimal impact on all other parts of the ecosystem
+
+### Compatibility
+
+Proposal in this form is not compatible with older toold that do not implement proper MetadataV14 self-descriptive features; those would have to be upgraded to include a new signed extensions field.
```
agryaznov, line 156:
```suggestion
Proposal in this form is not compatible with older tools that do not implement proper MetadataV14 self-descriptive features; those would have to be upgraded to include a new signed extensions field.
```

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Prior Art and References
+
+This project was developed as Polkadot Treasury grant; relevant development links are located in [metadata-offline-project](https://github.com/Alzymologist/metadata-offline-project) repository.
```
agryaznov, line 160:
```suggestion
This project was developed upon a Polkadot Treasury grant; relevant development links are located in [metadata-offline-project](https://github.com/Alzymologist/metadata-offline-project) repository.
```

bkchr:
I support what @tomaka is saying. IMO it requires much more information, like the actual data structures in pseudo code etc.

> It would also have been nice to give a brief overview of what the metadata contains, rather than assume that the reader is familiar with that, because I don't think that there is a single document that can be found online that describes the metadata in details. This would make it possible to review what is left and what is stripped.

I would say that the actual metadata is out of scope. However, we should explain what it is as Pierre said and then explain the structures that are put into the trie etc.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+1. The metadata is converted into lean modular form (vector of chunks)
+2. A Merkle tree is constructed from the metadata chunks
+3. A root of tree (as a left element) is merged with Metadata Descriptor (as a right element)
+4. Resulting value is a constant to be included in `additionalSigned` to prove that the metadata seen by cold device is genuine
+
+### Metadata modularization
+
+1. Types registry is stripped from `docs` fields.
+2. Types records are separated into chunks, with enum variants being individual chunks differing by variant index; each chunk consisting of `id` (same as in full metadata registry) and SCALE-encoded 'Type' description (reduced to 1-variant enum for enum variants). Enums with 0 variants are treated as regular types.
+3. Chunks are sorted by `id` in ascending order; chunks with same `id` are sorted by enum variant index in ascending order.
```
bkchr, line 146: We clearly should not mention the `docs` field, as I assume that this isn't present in the "shortened metadata" representation. In general this algorithm does not make that much difference, if you don't know the types.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+3   4
+```
+
+### Digest
+
+1. Blake3 hash is computed for each chunk of modular short metadata registry.
+3. Complete Binary Merkle Tree is constructed as described above.
+4. Root hash of this tree (left) is merged with metadata descriptor blake3 hash (right); this is metadata digest.
+
+Version number and corresponding resulting metadata digest MUST be included into Signed Extensions as specified in Chain Verification section below.
```
bkchr, line 206: IMO it should be fine to just say that you use the root hash plus the hash of the metadata descriptor as digest.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+```
+
+### Merging protocol
+
+`blake3` transformation of concatenated child nodes (`blake3(left + right)`) as merge procedure;
+
+### Complete Binary Merkle Tree construction protocol
+
+1. Leaves are numbered in ascending order. Leaf index is associated with corresponding chunk.
```
bkchr, line 182: Isn't it the other way around, you associate chunks with leaf indices? 

In general I would also by intuition, order all leaves in ascending order. Then I would chunk them into two and hash them. The output of this would follow the same procedure until only one hash (the root hash) is left.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Transmission
+
+Shortened metadata chunks MAY be trasmitted into cold device together with Merkle proof in its entirety or in parts, depending on memory capabilities of the cold device and it ability to reconstruct larger fraction of tree. This document does not specify the manner of transmission. The order of metadata chunks MAY be arbitrary, the only requirement is that indices of leaf nodes in Merkle tree corresponding to chunks MUST be communicated. Community MAY handle proof format standartization independently.
+
+### Offline verification
+
+The transmitted metadata chunks are hashed together with proof lemmas to obtain root that MAY be transmitted along with the rest of payload. Verification that the root transmitted with message matches with calculated root is optional; the transmitted root SHOULD NOT be used in signature, calculated root MUST be used; however, there is no mechanism to enforce this - it should be done during cold signers code audit.
```
bkchr, line 218: Same for these.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Chain verification
+
+The root of metadata computed by cold device MAY be included into Signed Extensions; this way the transaction will pass as valid iff hash of metadata as seen by cold storage device is identical to consensus hash of metadata, ensuring fair signing protocol.
```
bkchr, line 222: I mean this isn't entirely correct, because below you also explain that passing `0` means that no hash is compared.

bkchr (commented):
Goes into a good direction. However, I still miss a real explanation of the types that are part of the shortened metadata, as they are crucial for the protocol being described here.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+Some cryptographically strong digest of metadata MAY be included into signable blob. There SHALL NOT be storage overhead for this blob, nor computational overhead, on the node side; thus MUST be a constant within given runtime version, deterministically defined by metadata.
+
+ - Metadata information that could be used in signable extrinsic decoding MAY be included in digest, its inclusion MUST be indicated in signed extensions;
+ - Digest MUST be deterministic with respect to metadata;
+ - Digest MUST be cryptographically strong against pre-image, both first and second;
+ - Extra-metadata information necessary for extrinsic decoding and constant within runtime version MUST be included in digest;
+ - It SHOULD be possible to quickly withdraw offline signing mechanism without access to cold signing devices;
+ - Digest format SHOULD be versioned.
+ - Work necessary for proving metadata authenticity MAY be omitted at discretion of signer device design (to support automation tools).
```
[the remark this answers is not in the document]
carlosala: I'd say that the device may choose not to verify the data. It'd be like current situation with PJS extension for example, that gets the metadata and uses it to interpret the extrinsic, without verifying its authenticity (probably because it can't).

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+#### Reduce metadata size
+
+Metadata should be stripped from parts that are not necessary to parse a signable extrinsic, then it should be separated into a finite set of self-descriptive chunks. Thus, a subset of chunks necessary for signable extrinsic decoding and rendering could be sent, possibly in small portions (ultimately - one at a time), to cold device together with proof.
```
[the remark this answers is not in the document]
carlosala: Indeed, they are if you accept that under SCALE-encoding nothing is self-descriptive by itself, but it relies on `RuntimeMetadataV14` registry type and basic SCALE types.
Both parts should know what `Registry` and SCALE is.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+ - Single chunk with proof payload size SHOULD fit within few kB;
+ - Chunks handling mechanism SHOULD support chunks being sent in any order without memory utilization overhead;
+ - Unused enum variants MUST be stripped (this has great impact on transmitted metadata size; examples: era enum, enum with all calls for call batching).
```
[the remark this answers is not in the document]
carlosala: Not necessarily. When sending data to the device, only used variants are sent (all in one type). This differs from the proof generation, which is explained afterwards.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+1. `u8` metadata shortening protocol version, 
+2. SCALE-encoded `ExtrinsicMetadata`,
+3. SCALE-encoded `spec_version` `String`,
+4. SCALE-encoded `spec_name` `String`,
+5. `u16` base58 prefix,
+6. `u8` decimals value or `0u8` if no units are defined,
+7. SCALE-encoded `tokenSymbol` `String` defined on chain to identify the name of currency (available for example through `system.properties()` RPC call) or empty string if no base units are defined,
+
+```
+struct MetadataDescriptor { // really a scale-encoded enum, thus first field is enum value - only 0x01 currently supported.
```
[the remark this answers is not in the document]
carlosala: I agree. `V0` used to be the version that we used during development, that's why it starts with `1` (`0` was already burned let's say)
I'd clarify that.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+  protocol_version: u8,
```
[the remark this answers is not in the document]
carlosala: AFAIK the byte that we have in the digest is the same as this one. We're then 100% sure that both in runtime build and in metadata shortening process (that happen in different envs and different times) the same process was used.

It seems to be the `index` of the enum actually. Changing the declaration to an enum would simplify the understanding of it.
I'd as well mention that `V0` is already burned.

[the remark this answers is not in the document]
bkchr: > AFAIK the byte that we have in the digest is the same as this one. We're then 100% sure that both in runtime build and in metadata shortening process (that happen in different envs and different times) the same process was used.

But we already put this byte into the tx and it will be signed. So, we already ensure this?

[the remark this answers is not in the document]
carlosala: Yeh, true... I still see value for the devices parsing the metadata (since it's scale, they need to know the structure). With that being an enum and not a simple struct, we're sure that we are parsing the info correctly (in terms of metadata, not blob), and it also gives us the flexibility to change it in any chain and keep backwards-compatibility as well with the previous versions.
Otherwise, we need the blob to know how to parse the metadata, and we need the metadata to know how to parse the blob. It's kind of a vicious circle.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 2023-10-31                                                                                  |
+| **Description** | Add SignedExtension to check Metadata Root Hash                                             |
+| **Authors**     | Alzymologist Oy, Zondax LLC, Parity GmbH                                                    |
+
+## Summary
+
+Add a metadata digest value (33-byte constant within fixed `spec_version`) to Signed Extensions to supplement signer party with proof of correct extrinsic interpretation. The digest value is generated once before release and is well-known and deterministic. The digest mechanism is designed to be modular and flexible. It also supports partial metadata transfer as needed by the signing party's extrinsic decoding mechanism. This considers signing devices potentially limited communication bandwidth and/or memory capacity.
```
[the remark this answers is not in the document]
Slesarew:
```suggestion
Add a metadata digest value (1-byte per transaction cost + 32-byts constant in `additionalSigned` within fixed `spec_version`) to Signed Extensions to supplement signer party with proof of correct extrinsic interpretation. The digest value is generated once before release and is well-known and deterministic. The digest mechanism is designed to be modular and flexible. It also supports partial metadata transfer as needed by the signing party's extrinsic decoding mechanism. This considers signing devices potentially limited communication bandwidth and/or memory capacity.
```

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+In order to enable decoding that is small and optimized for chain storage transactions, a metadata entity is used, which is not at all small in itself (on the order of half-MB for most networks). This is a dynamic data chunk which completely describes chain interfaces and properties that could be made into a portable scale-encoded string for any given network version and passed along into an off-chain device to familiarize it with latest network updates. Of course, compromising this metadata anywhere in the path could result in differences between what user sees and signs, thus it is essential that we protect it.
+
+Therefore, we have 2 problems to be solved:
+
+1. Metadata is large, takes long time to be passed into a cold storage device with memory insufficient for its storage; metadata SHOULD be shortened and transmission SHOULD be optimized.
+2. Metadata authenticity SHOULD be ensured.
+
+As of now, there is no working solution for (1), as the whole metadata has to be passed to the device. On top of this, the solution for (2) heavily relies on a trusted party managing keys and ensuring metadata is indeed authentic: creating poorly decentralized points of potential failure.
```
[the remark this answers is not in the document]
Slesarew: this is not exactly true; there was an attempt to decentralize it at the very beginning: anyone can become a trusted party. As far as I know, at least a few big parties emerged that handled certificates routine correctly (Parity, Nova, and some parachains issued their certificate collections). Furthermore, many people verified their metadata manually creating their personal trust sources - which is a perfectly valid and safe strategy given enough time and discipline (although few people can afford this and there are downsides as well as benefits when compared to centralized distribution). So this was not centralized after all, but decentralization was clearly insufficient.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Solution requirements
+
+#### Include metadata digest into signature
+
+Some cryptographically strong digest of metadata MAY be included into signable blob. There SHALL NOT be storage overhead for this blob, nor computational overhead, on the node side; thus MUST be a constant within given runtime version, deterministically defined by metadata.
+
+ - Metadata information that could be used in signable extrinsic decoding MAY be included in digest, its inclusion MUST be indicated in signed extensions;
+ - Digest MUST be deterministic with respect to metadata;
+ - Digest MUST be cryptographically strong against pre-image, both first and second;
```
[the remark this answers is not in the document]
Slesarew: First pre-image attack means finding an input that results in given digest. Second pre-image attack means that finding an input that results in same digest as some other input given.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+ - Extra-metadata information necessary for extrinsic decoding and constant within runtime version MUST be included in digest;
+ - It SHOULD be possible to quickly withdraw offline signing mechanism without access to cold signing devices;
```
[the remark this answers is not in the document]
Slesarew: yes, and all wallets would have to explicitly change this value, no chance of partial compatibility here. There are many strategies for communicating correct version per chain to cold signer if these diverge for some reason, I think it's out of scope here.

text/0046-metadata-for-offline-signers.md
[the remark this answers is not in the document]
Slesarew: It's something we have no chance of enforcing here really. So we just state that it is not provable at this part of the architecture and expect the possibility of absence of this check when designing security strategies (for example, this would be required check in app code audits).

Also, tools like staking payout machine do not really need to check any metadata authenticity since they are limited to small number of calls and are often connected to petty cash accounts that are not worth being attacked anyway. We have an expetimental project ourselves where a signer resides on a physical key for physical lock, it does not need metadata verification as it only signs very narrow range of extrinsics (and it would severely degrade performance to do the check as hardware there is really humble); this is alternative - and very centralized in a sense, you just trust specialized firmware developer - approach to the same problem. To support signers like these we need to keep this optional.

text/0046-metadata-for-offline-signers.md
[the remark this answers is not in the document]
Slesarew: The manner of these being sent to device is not specified really, they could be sent as separate chunks or as fewer variant enums or even as full enums, why not. What is important, they are indeed separate chunks for proof generation.

text/0046-metadata-for-offline-signers.md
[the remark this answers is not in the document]
Slesarew: It's just one byte, but a byte that makes implementations a bit more ideomatic and helps developers explicitly state which version of shortener protocol they are using at the moment in their code. I think it's worth it. It's not even mandatory to send it every time - signer device could add it itself - if sending one extra byte matters.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+  spec_name: Vec<u8>, // SCALE from `String`
+  base58_prefix: u16,
+  decimals: u8,
+  token_symbol: Vec<u8>, // SCALE from `String`
+}
+```
+
+constitute metadata descriptor. This is minimal information that is, together with (shortened) types registry, sufficient to decode any signable transaction.
+
+#### Merkle tree
```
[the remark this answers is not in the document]
Slesarew: This is just for conventions - what exactly notation would be used further. Not sure it is used much right now, but I don't see how this hurts either. Might help ask questions for reviewers in same definitions to speed up this process a bit.

text/0046-metadata-for-offline-signers.md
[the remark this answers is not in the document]
Slesarew: It's struct that encodes an enum. I thought that it would be more clear at this point but have nothing against changing it back. (this involves next line so I'll just commit, not suggest)

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+Node 0 is root
+
+```
+
+### General flow
+
+1. The metadata is converted into lean modular form (vector of chunks)
+2. A Merkle tree is constructed from the metadata chunks
+3. A root of tree (as a left element) is merged with Metadata Descriptor (as a right element)
+4. Resulting value is a constant to be included in `additionalSigned` to prove that the metadata seen by cold device is genuine
```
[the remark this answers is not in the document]
Slesarew: This is a general flow of the whole construction; the (4) just indicates final destination of result, as mentioned elsewhere (we say that we are adding a constant to `additionalSigned` and do not give any name to that step; thus here we just repeat those words to connect flow).

text/0046-metadata-for-offline-signers.md
Slesarew: Types registry comes directly from full, non-shortened metadata, that has all fields, and this (1) is the stage where it loses `docs`, not earlier. As we describe algorithm here, this is an essential step. The only other way to describe this would be copying all other fields to new structure, but that is just longer description of the same step.

Also, there is ambiguity about docs field missing altogether vs being blank that would probably surface across implementations; this statement should resolve it.

text/0046-metadata-for-offline-signers.md
Slesarew: Are you suggesting removal of line 206? I was kind of afraid this would be a bit difficult to follow if we don't state what's the next step, but yes, we can remove it. Should we?

text/0046-metadata-for-offline-signers.md
Slesarew: There are other approaches; this is kind of canonical for CBMT, but there could be implementers that do not know canonical stuff. I know we've independently made exactly same implementation for hot-to-cold proof transfer protocol with @carlosala based on natural order of universe, but I'm afraid here we should be a bit more explicit as canonical here is not the best construction path, just a convenient one. Ordering matters here, and also there is an issue of pairing last odd leaf that could be resolved in more awkward or even slightly less efficient way (making tree incomplete, but I'm not sure it's obvious for everyone).

Also, just following my description verbatim, it's possible to construct the tree without knowing much of background. That should just make implementations easier hopefully. I was thinking about also writing pseudocode representation, but that's probably too much already.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+1. Blake3 hash is computed for each chunk of modular short metadata registry.
+3. Complete Binary Merkle Tree is constructed as described above.
+4. Root hash of this tree (left) is merged with metadata descriptor blake3 hash (right); this is metadata digest.
+
+Version number and corresponding resulting metadata digest MUST be included into Signed Extensions as specified in Chain Verification section below.
+
+### Shortening
+
+For shortening, an attempt to decode transaction completely using provided metadata is performed with the same algorithm that would be used on the cold side. All chunks are associated with their leaf indices. An example of this protocol is proposed in [metadata-shortener](https://github.com/Alzymologist/metadata-shortener) that is based on [substrate-parser](https://github.com/Alzymologist/substrate-parser) decoding protocol; any decoding protocol could be used here as long as cold signer's design finds it appropriate for given security model.
```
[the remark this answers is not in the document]
Slesarew: Do you mean this?

```suggestion

For shortening, an attempt to decode transaction completely using provided metadata is performed with the same algorithm that would be used on the cold side. All chunks are associated with their leaf indices. Any decoding protocol could be used here as long as cold signer's design finds it appropriate for given security model.
```

text/0046-metadata-for-offline-signers.md
Slesarew: Or are you suggesting removing everything on partial proofs completely? This kind of explains why we cut things this way, although it does seem obvious.

The other reason to keep it here is just suggested security considerations that might be useful for implementors to understand what's happening here. Maybe it's not something RFC should be taking care of indeed.

 Please let me know (or just throw in a suggestion to remove these)

text/0046-metadata-for-offline-signers.md
Slesarew: if `0` is passed then either root was not included into Signer Extensions ("MAY be included but it is not"), or the signature would be wrong as chain expects it to not be included ("MAY be included and is but incorrectly"); thus if what MAY be IS, we have a valid iff condition, otherwise this mechanism is just not used.

Another thing is that cold signer might just cheat and keep passing a constant, but that's out of scope here I suppose.

Maybe this clarification would be nice?

```suggestion
The root of metadata computed by cold device MAY be included into Signed Extensions; if it is included, the transaction will pass as valid iff hash of metadata as seen by cold storage device is identical to consensus hash of metadata, ensuring fair signing protocol.
```

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+Proposal in this form is not compatible with older tools that do not implement proper MetadataV14 self-descriptive features; those would have to be upgraded to include a new signed extensions field.
+
+## Prior Art and References
+
+This project was developed upon a Polkadot Treasury grant; relevant development links are located in [metadata-offline-project](https://github.com/Alzymologist/metadata-offline-project) repository.
+
+## Unresolved Questions
+
+2. How would polkadot-js handle the transition?
```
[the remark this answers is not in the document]
Slesarew: It should pass `0` in general, but it has compatibility layer with cold signers that require other options, and even to pass `0` it needs some patches.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+### Offline verification
+
+The transmitted metadata chunks are hashed together with proof lemmas to obtain root that MAY be transmitted along with the rest of payload. Verification that the root transmitted with message matches with calculated root is optional; the transmitted root SHOULD NOT be used in signature, calculated root MUST be used; however, there is no mechanism to enforce this - it should be done during cold signers code audit.
+
+### Chain verification
+
+The root of metadata computed by cold device MAY be included into Signed Extensions; this way the transaction will pass as valid iff hash of metadata as seen by cold storage device is identical to consensus hash of metadata, ensuring fair signing protocol.
+
+The Signed Extension representing metadata digest is a single byte representing both digest vaule inclusion and shortening protocol version; this MUST be included in Signed Extensions set. Depending on its value, a digest value is included as `additionalSigned` to signature computation according to following specification:
```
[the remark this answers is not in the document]
Slesarew: we add 1 byte that is version/inclusion byte, and then we add 32-byte digest to that part of data. Not explaining that it is part of `additionalSigned` would just add to unmeasurable amount of confusion around this obscure entity. I lost count of times I've been asked how Signed Extensions work and I spent way too much time figuring it out myself first, probably massively pestered you as well, I don't remember. I'm pretty solid that this should stay here if we want to have any hope of more implementations appearing eventually.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+Proposal in this form is not compatible with older tools that do not implement proper MetadataV14 self-descriptive features; those would have to be upgraded to include a new signed extensions field.
+
+## Prior Art and References
+
+This project was developed upon a Polkadot Treasury grant; relevant development links are located in [metadata-offline-project](https://github.com/Alzymologist/metadata-offline-project) repository.
+
+## Unresolved Questions
+
+2. How would polkadot-js handle the transition?
+3. Where would non-rust tools like Ledger apps get shortened metadata content?
```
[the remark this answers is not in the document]
Slesarew: This is being discussed; there is a proposal to just wrap it in wasm. There also was a proposal to have a centralized untrusted services to provide these blobs. As far as I understand, these proposals are still alive

text/0046-metadata-for-offline-signers.md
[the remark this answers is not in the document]
bkchr: I meant just to drop `additionalSigned` and describe it in a more human friendly way. Because outside of Substrate it is maybe not called `additionalSigned`.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+
+## Drawbacks
+
+### Increased transaction size
+
+A 1-byte increase in transaction size due to signed extension value. Digest is not included in transferred transaction, only in signing process.
+
+### Transition overhead
+
+Some slightly out of spec systems might experience breaking changes as new content of signed extensions is added. It is important to note, that there is no real overhead in processing time nor complexity, as the metadata checking mechanism is voluntary. The only drawbacks are expected for tools that do not implement MetadataV14 self-descripting features.
```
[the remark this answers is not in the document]
bkchr: But this is not related to the metadata version. Even if you know how to decode all the signed extensions using the type info of the metadata, you don't know how to construct the signed extension. It makes no sense that any tool would just pass `0` there, it should reject because it doesn't know the signed extension, which pjs is doing.

bkchr:
> Just realized this was the main question. The types are exactly same as they are defined in `frame-metadata`, so that we don't have to re-implement every single decoding tool. The only difference is that docs are left blank.

But the types in `frame-metadata` probably change per version. There should be a fixed representation of the types being used in this RFC.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,294 @@
...
+```
+
+### Merging protocol
+
+`blake3` transformation of concatenated child nodes (`blake3(left + right)`) as merge procedure;
+
+### Complete Binary Merkle Tree construction protocol
+
+1. Leaves are numbered in ascending order. Leaf index is associated with corresponding chunk.
+2. Merge is performed using the leaf with highest index as right and node with second to highest index as left children; result is pushed to the end of nodes queue and leaves are discarded.
```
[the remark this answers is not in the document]
carlosala: Left and right always mean the order in hashing: `hash(<left><right>)`

text/0046-metadata-for-offline-signers.md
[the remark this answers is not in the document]
Slesarew:
```suggestion
Add a metadata digest value to signed data to supplement signer party with proof of correct extrinsic interpretation. This would ensure that hardware wallets always use correct metadata to decode the information for the user.

The digest value is generated once before release and is well-known and deterministic. The digest mechanism is designed to be modular and flexible. It also supports partial metadata transfer as needed by the signing party's extrinsic decoding mechanism. This considers signing devices potentially limited communication bandwidth and/or memory capacity.
```

text/0046-metadata-for-offline-signers.md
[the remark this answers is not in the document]
Slesarew:
```suggestion
 - Digest MUST be cryptographically strong against pre-image, both first (finding an input that results in given digest) and second (finding an input that results in same digest as some other input given);
```

text/0046-metadata-for-offline-signers.md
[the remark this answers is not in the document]
Slesarew:
```suggestion
Some slightly out of spec systems might experience breaking changes as new content of signed extensions is added - tools that delay their transition instead of preparing ahead of time would break for the duration of delay. It is important to note, that there is no real overhead in processing time nor complexity, as the metadata checking mechanism is voluntary.
```

text/0046-metadata-for-offline-signers.md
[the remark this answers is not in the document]
Slesarew: How would I explain it differently? It's a field that's called that way in metadata and at least in major Rust and JS codebases, if we drop the name or change it nobody would know where this belongs. Things that go into signature and their order are extremely confusing. It's not described anywhere (yet) in full. So if we just say that it is added to data that gets signed, the manner of addition would remain unclear - and I'm not describing this step anywhere else in the document.

text/0046-metadata-for-offline-signers.md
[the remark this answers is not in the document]
Slesarew:
```suggestion
```

text/0046-metadata-for-offline-signers.md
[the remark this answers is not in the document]
Slesarew: indeed. Looks like we are out of questions. Cool!

```suggestion
```

text/0046-metadata-for-offline-signers.md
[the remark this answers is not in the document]
Slesarew: Just as a reminder to everybody that this is a fact - you should not rely on this actually happening. Explicit statement that proving that the check actually happened is not planned at all, just that the tool that could have done it had all required information in genuine form. I think this is important for future security analysts.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,274 @@
...
+#### Metadata structure
+
+Metadata in general consists of four sections:
+
+1. Types registry
+2. Pallets
+3. Extrinsic metadata
+4. Runtime type
+
+Of these, only sections 1-3 contain information required for extrinsic decoding. The most important section is (1) Types registry, that is mostly used in extrinsic decoding. It is also the largest part, thus it is modularized for fractional transmission. Part (2) contains runtime version and is otherwise useless for transaction decoding; thus its contents are reduced to this parameter and included into Metadata Descriptor. Part (3) is included into Metadata Descriptor verbatim.
```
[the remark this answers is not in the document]
Slesarew:
```suggestion
## Explanation

Detailed description of metadata shortening and digest process is provided in [metadata-shortener](https://github.com/Alzymologist/metadata-shortener) crate (see `cargo doc --open` and examples). Below are presented algorithms of the process.

### Definitions

```

Slesarew:
> > Just realized this was the main question. The types are exactly same as they are defined in `frame-metadata`, so that we don't have to re-implement every single decoding tool. The only difference is that docs are left blank.
> 
> But the types in `frame-metadata` probably change per version. There should be a fixed representation of the types being used in this RFC.

Ok, I've mentioned that we imitate MetadataV14 and also added the structure.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,295 @@
...
+
+#### Metadata descriptor
+
+Values for:
+
+1. `u8` metadata shortening protocol version, 
+2. SCALE-encoded `ExtrinsicMetadata`,
+3. SCALE-encoded `spec_version` `String`,
+4. SCALE-encoded `spec_name` `String`,
+5. `u16` base58 prefix,
```
xlc, line 77:
```suggestion
5. `u16` ss58 prefix,
```

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,295 @@
...
+enum MetadataDescriptor {
+  V0,
+  V1(MetadataDescriptorV1),
+}
+
+struct MetadataDescriptorV1 {
+  extrinsic_metadata: Vec<u8>, // SCALE from `ExtrinsicMetadata`
+  spec_version: Vec<u8>, // SCALE form `String`
+  spec_name: Vec<u8>, // SCALE from `String`
+  base58_prefix: u16,
```
xlc, line 91:
```suggestion
  ss58_prefix: u16,
```

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,295 @@
...
+### Definitions
+
+
+#### Metadata descriptor
+
+Values for:
+
+1. `u8` metadata shortening protocol version, 
+2. SCALE-encoded `ExtrinsicMetadata`,
+3. SCALE-encoded `spec_version` `String`,
```
xlc, line 75: why this is string, not u32?

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,295 @@
...
+4. SCALE-encoded `spec_name` `String`,
+5. `u16` base58 prefix,
+6. `u8` decimals value or `0u8` if no units are defined,
+7. SCALE-encoded `tokenSymbol` `String` defined on chain to identify the name of currency (available for example through `system.properties()` RPC call) or empty string if no base units are defined,
+
+```
+enum MetadataDescriptor {
```
xlc, line 82: this is SCALE encoded right?

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,295 @@
...
+  V0,
+  V1(MetadataDescriptorV1),
+}
+
+struct MetadataDescriptorV1 {
+  extrinsic_metadata: Vec<u8>, // SCALE from `ExtrinsicMetadata`
+  spec_version: Vec<u8>, // SCALE form `String`
+  spec_name: Vec<u8>, // SCALE from `String`
```
xlc, line 90: SCALE encoded String have length prefix. SCALE encoded Vec have length prefix. Are we double SCALE encoding here or not? The comment makes this more confusing.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,295 @@
...
+7. SCALE-encoded `tokenSymbol` `String` defined on chain to identify the name of currency (available for example through `system.properties()` RPC call) or empty string if no base units are defined,
+
+```
+enum MetadataDescriptor {
+  V0,
+  V1(MetadataDescriptorV1),
+}
+
+struct MetadataDescriptorV1 {
+  extrinsic_metadata: Vec<u8>, // SCALE from `ExtrinsicMetadata`
```
xlc, line 88:
```suggestion
  extrinsic_metadata: ExtrinsicMetadata,
```

can we just do this?

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,295 @@
...
+  spec_version: Vec<u8>, // SCALE form `String`
+  spec_name: Vec<u8>, // SCALE from `String`
+  base58_prefix: u16,
+  decimals: u8,
```
xlc, line 92: what about for the chain that does not have native token or multiple native tokens? I guess we can pass 0 and empty vec for symbol but best to be explicit

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Definitions
+
+
+#### Metadata descriptor
+
+Values for:
+
+1. `u8` metadata shortening protocol version, 
+2. SCALE-encoded `ExtrinsicMetadata`,
```
xlc, line 74: what is `ExtrinsicMetadata`?

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,295 @@
...
+### General flow
+
+1. The metadata is converted into lean modular form (vector of chunks)
+2. A Merkle tree is constructed from the metadata chunks
+3. A root of tree is merged with the hash of the `MetadataDescriptor`
+4. Resulting value is a constant to be included in `additionalSigned` to prove that the metadata seen by cold device is genuine
+
+### Metadata modularization
+
+Structure of types in shortened metadata exactly matches structure of types in `scale-info` at MetadataV14 state, but `doc` field is always empty
```
xlc, line 140: do we want to use V15?

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,295 @@
...
+2. Merge is performed using the leaf with highest index as right and leaf with second to highest index as left children; result is pushed to the end of nodes queue and leaves are discarded.
+3. Step (2) is repeated until no leaves or just one leaf remains; in latter case, the last leaf is pushed to the front of the nodes queue.
+4. Right node and then left node is popped from the front of the nodes queue and merged; the result is sent to the end of the queue.
+5. Step (4) is repeated until only one node remains; this is tree root.
+
+
+```
+queue = empty_queue
+
+while (leaves.length>1) {
```
xlc, line 204:
```suggestion
while leaves.length > 1 {
```

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,295 @@
...
+|------------------------|----------------|------------------------------------|
+| `0x00`                 |                | digest is not included             |
+| `0x01`                 | 32-byte digest | this represents protocol version 1 |
+| `0x02` - `0xFF`        | *reserved*     | reserved for future use            |
+
+## Drawbacks
+
+### Increased transaction size
+
+A 1-byte increase in transaction size due to signed extension value. Digest is not included in transferred transaction, only in signing process.
```
xlc, line 259: we do already have a version byte in extrinsic format. maybe we can just use it instead of a new byte?

bkchr:
I already posted this yesterday internally in some chat, I also want to make it public here. I'm still not convinced that you can implement this Rfc by reading the Rfc. To actually implementing you will need to lookup all the types that are not mentioned here in the scale codec crate. I would personally also layout a more customized format for the types that would for example remove the docs field everywhere. Currently the docs are emptied everywhere, but it means that there is everytime a byte being used for each encoded type information.

text/0046-metadata-for-offline-signers.md
carlosala: That's a typo, it should be `form` @Slesarew

text/0046-metadata-for-offline-signers.md
Slesarew: Because we've seen parachains and standalone chains where this was not `u32`. Those went extinct some time ago though. This is not forbidden anywhere and people are experimenting sometimes, there is no reason to punish them. Or is there? We certainly should change it into `u32` (it would be easier and cleaner indeed) if this is planned to be forced, but that has to be documented well in chain developer docs first, which is out of my reach.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,295 @@
+# RFC-0000: Metadata for offline signers
```
joepetrowski, line 1:
```suggestion
# RFC-0046: Metadata for Offline Signers
```

Slesarew:
> I already posted this yesterday internally in some chat, I also want to make it public here. I'm still not convinced that you can implement this Rfc by reading the Rfc. To actually implementing you will need to lookup all the types that are not mentioned here in the scale codec crate. 

Do you suggest we just place whole `frame-metadata` documentation here? We are trying to reuse the types, as was mentioned many times, as this way we reuse old decoding code and speed up adoption.

> I would personally also layout a more customized format for the types that would for example remove the docs field everywhere. Currently the docs are emptied everywhere, but it means that there is everytime a byte being used for each encoded type information.

and this would exactly break that idea for - what? It's typical 20 chunks, at most 50 for heavy XCM calls - so we'll win 50 bytes out of few kb of chunks and proof data, that are not even close to the chain - that are just transferred to cold device and consumed there, with auxiliary data, erasure overheads, and error correction codes (or flash momentarily at build time of runtime to generate digest constant).

Slesarew:
We've also discussed this privately and would change the format slightly to better adhere to metadata V15 soon. This would remain backward-compatible with V14, as well as non-shortened metadata itself. I'll commit the update soon.

text/0046-metadata-for-offline-signers.md
```diff
@@ -0,0 +1,405 @@
+# RFC-0046: Metadata for Offline Signers
+
+|                 |                                                                                              |
+| --------------- | -------------------------------------------------------------------------------------------- |
+| **Start Date**  | 2023-10-31                                                                                   |
+| **Description** | Include merkelized metadata hash in extrinsic signature for trust-less metadata verification.|
+| **Authors**     | Alzymologist Oy, Zondax LLC, Parity Technologies                                             |
```
jleni, line 7:
```suggestion
| **Authors**     | Alzymologist Oy, Zondax AG, Parity Technologies                                             |
```

The company name should be corrected. Zondax is an AG.

jleni (approved):

text/0046-metadata-for-offline-signers.md
jleni, line 7:
```suggestion
| **Authors**     | Alzymologist Oy, Zondax AG, Parity Technologies                                             |
```

The company name should be corrected. Zondax is an AG.

jleni (changes requested):

archtuxfan:
No update since Jan 30?

Slesarew:
> No update since Jan 30?

My fellow archer, I've been informed that the Fellowship intends to re-do this work internally based on this proposal; thus this (which is certainly a finished text as far as I can tell) should be superseded now by another PR. The latter is supposed to be voted on and merged quite soon as far as I understand, so that's good news. I think this PR would just be closed in case of success afterwards.

archtuxfan:
> > No update since Jan 30?
> 
> My fellow archer, I've been informed that the Fellowship intends to re-do this work internally based on this proposal; thus this (which is certainly a finished text as far as I can tell) should be superseded now by another PR. The latter is supposed to be voted on and merged quite soon as far as I understand, so that's good news. I think this PR would just be closed in case of success afterwards.

My understanding is that this needs to pass before the ledger app gets released. Wondering if anyone has a sense on the ETA?

ggwpez:
Isnt this superseded by https://github.com/polkadot-fellows/RFCs/pull/78?

bkchr:
Yes.

PR #47, opened 2023-11-13
Random assignment of availability-chunk indices to validators

alindima:
[Rendered](https://github.com/alindima/RFCs/blob/av-chunk-indices/text/0047-random-assignment-of-availability-chunks.md)

This RFC proposes a way of randomly permuting the availability chunk indices assigned to validators for a given core and relay chain block, in the context of [recovering available data from systematic chunks](https://github.com/paritytech/polkadot-sdk/issues/598).

### DIFF
--- a/text/0047-random-assignment-of-availability-chunks.md
+++ b/text/0047-random-assignment-of-availability-chunks.md
@@ -1,0 +1,179 @@
+# RFC-0047: Random assignment of availability chunks to validators
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 03 November 2023                                                                            |
+| **Description** | An evenly-distributing indirection layer between availability chunks and the validators that hold them.|
+| **Authors**     | Alin Dima                                                                                   |
+
+## Summary
+
+Propose a way of randomly permuting the availability chunk indices assigned to validators for a given core and relay chain block,
+in the context of [recovering available data from systematic chunks](https://github.com/paritytech/polkadot-sdk/issues/598).
+
+## Motivation
+
+The relay chain node must have a deterministic way of evenly distributing the first ~(N_VALIDATORS / 3) systematic availability chunks to different validators,
+based on the session, relay chain block number and core.
+This is needed in order to optimise network load distribution as evenly as possible during availability recovery.
+
+## Stakeholders
+
+Relay chain node core developers.
+
+## Explanation
+
+### Systematic erasure codes
+
+An erasure coding algorithm is considered systematic if it preserves the original unencoded data as part of the resulting code.
+[The implementation of the erasure coding algorithm used for polkadot's availability data](https://github.com/paritytech/reed-solomon-novelpoly) is systematic. Roughly speaking, the first N_VALIDATORS/3
+chunks of data can be cheaply concatenated to retrieve the original data, without running the resource-intensive and time-consuming decoding algorithm.
+
+### Availability recovery from systematic chunks
+
+As part of the effort of [increasing polkadot's resource efficiency, scalability and performance](https://github.com/paritytech/roadmap/issues/26), work is under way to
+modify the Availability Recovery subsystem by leveraging systematic chunks. See [this comment](https://github.com/paritytech/polkadot-sdk/issues/598#issuecomment-1792007099)
+for preliminary performance results.
+
+In this scheme, the relay chain node will first attempt to retrieve the N/3 systematic chunks from the validators that should hold them,
+before falling back to recovering from regular chunks, as before.
+
+The problem that this RFC aims to address is that, currently, the ValidatorIndex is always identical to the ChunkIndex.
+Since the validator array is only shuffled once per session, naively using the ValidatorIndex as the ChunkIndex
+would pose an unreasonable stress on the first N/3 validators during an entire session.
+
+### Chunk assignment function
+
+#### Properties
+The function that decides the chunk index for a validator should be parametrised by at least `(validator_index, session_index, block_number, core_index)`
+and have the following properties:
+1. deterministic
+1. pseudo-random
+1. relatively quick to compute and resource-efficient.
+1. when considering the other params besides `validator_index` as fixed,
+the function should describe a random permutation of the chunk indices
+1. considering `session_index` and `block_number` as fixed arguments, the validators that map to the first N/3 chunk indices should
+have as little overlap as possible for different cores.
+
+#### Proposed function and runtime API
+
+Pseudocode:
+
+```rust
+pub fn get_chunk_index(
+    n_validators: u32,
+    validator_index: ValidatorIndex,
+    session_index: SessionIndex,
+    block_number: BlockNumber,
+    core_index: CoreIndex
+) -> ChunkIndex {
+    let threshold = systematic_threshold(n_validators); // Roughly n_validators/3
+    let seed = derive_seed(session_index, block_number);
+    let mut rng: ChaCha8Rng = SeedableRng::from_seed(seed);
+    let mut chunk_indices: Vec<ChunkIndex> = (0..n_validators).map(Into::into).collect();
+    chunk_indices.shuffle(&mut rng);
+
+    let core_start_pos = threshold * core_index.0;
+    return chunk_indices[(core_start_pos + validator_index) % n_validators]
+}
+```
+
+The function should be implemented as a runtime API, because:
+
+1. it's critical to the consensus protocol that all validators have a common view of the Validator->Chunk mapping.
+1. it enables further atomic changes to the shuffling algorithm.
+1. it enables alternative client implementations (in other languages) to use it
+1. mitigates the problem of third-party libraries changing the implementations of the `ChaCha8Rng` or the `rand::shuffle`
+that could be introduced in further versions, which would stall parachains. This would be quite an "easy" attack.
+
+Additionally, so that client code is able to efficiently get the mapping from the runtime, another API will be added
+for retrieving chunk indices in bulk for all validators at a given block and core.
+
+#### Upgrade path
+
+Considering that the Validator->Chunk mapping is critical to para consensus, the change needs to be enacted atomically via
+governance, only after all validators have upgraded the node to a version that is aware of this mapping.
+It needs to be explicitly stated that after the runtime upgrade and governance enactment, validators that run older client versions that don't
+support this mapping will not be able to participate in parachain consensus.
+
+### Getting access to core_index in different subsystems
+
+Availability-recovery can currently be triggered by three actors:
+1. `ApprovalVoting` subsystem of the relay chain validator.
+2. `pov_recovery` task of collators
+3. `DisputeCoordinator` subsystem of the relay chain validator.
+
+The one parameter of the assignment function that poses problems to some subsystems is the `core_index`.
+The `core_index` refers to the index of the core that the candidate was occupying while it was pending availability (from backing to inclusion).
+
+1. The `ApprovalVoting` subsystem starts the approval process on a candidate as a result to seeing a `CandidateIncluded` event.
+This event also contains the core index that the candidate is leaving, so getting access to the core index is not an issue.
+1. The `pov_recovery` task of the collators starts availability recovery in response to a `CandidateBacked` event, which enables
+easy access to the core index the candidate started occupying.
+1. Disputes may be initiated on a number of occasions:
+  
+    3.a. is initiated by the current node as a result of finding an invalid candidate while doing approval work. In this case,
+  availability-recovery is not needed, since the validator already issued their vote.
+    
+    3.b is initiated by noticing dispute votes recorded as a result of chain scraping. In this case, the subsystem
+  assumes it already has a copy of the `CandidateReceipt` as a result of scraping `CandidateBacked` events. The logic can be modified
+  to also record core indices.
+  
+    3.c is initiated as a result of getting a dispute statement from another validator. It is possible that the dispute is happening on
+  a fork that was not yet imported by this validator, so the subsystem will not always have access to the `CandidateBacked` event
+  recorded by the chain scraper.
+
+As a solution to 3.c, a new version for the disputes request-response networking protocol will be added.
+This message type will include the relay block hash where the candidate was included. This information will be used
+in order to query the runtime API and retrieve the core index that the candidate was occupying.
+
+The usage of the systematic data availability recovery feature will also be subject to all nodes using the V2 disputes networking protocol.
+
+#### Alternatives to using core_index
+
+As an alternative to core_index, the `ParaId` could be used. It has the advantage of being readily available in the
+`CandidateReceipt`, which would enable the dispute communication protocol to not change and would simplify the implementation.
+However, in the context of [CoreJam](https://github.com/polkadot-fellows/RFCs/pull/31), `ParaId`s will no longer exist (at least not in their current form).
+
+## Drawbacks
+
+Has a fair amount of technical complexity involved:
+
+- Introduces another runtime API that is going to be issued by multiple subsystems. With adequate client-side caching, this should be acceptable.
+
+- Requires a networking protocol upgrade on the disputes request-response protocol
+
+## Testing, Security, and Privacy
+
+Extensive testing will be conducted - both automated and manual.
+This proposal doesn't affect security or privacy.
+
+## Performance, Ergonomics, and Compatibility
+
+### Performance
+
+This is a necessary optimisation, as reed-solomon erasure coding has proven to be a top consumer of CPU time in polkadot when scaling the parachain count.
+
+With this optimisation, preliminary performance results show that CPU time used for reed-solomon coding can be halved and total POV recovery time decrease by 80% for large POVs. See more [here](https://github.com/paritytech/polkadot-sdk/issues/598#issuecomment-1792007099).
+
+### Ergonomics
+
+Not applicable.
+
+### Compatibility
+
+This is a breaking change. See "upgrade path" section above.
+All validators need to have upgraded their node versions before the feature will be enabled via a runtime upgrade and governance call.
+
+## Prior Art and References
+
+See comments on the [tracking issue](https://github.com/paritytech/polkadot-sdk/issues/598) and the [in-progress PR](https://github.com/paritytech/polkadot-sdk/pull/1644)
+
+## Unresolved Questions
+
+- Is it a future-proof idea to utilise the core index as a parameter of the chunk index compute function? Is there a better alternative that avoid complicating the implementation?
+- Is there a better upgrade path that would preserve backwards compatibility?
+
+## Future Directions and Related Material
+
+This enables future optimisations for the performance of availability recovery, such as retrieving batched systematic chunks from backers/approval-checkers.


alindima:
CC: @sandreim, @burdges, @ordian, @eskimor, @alexggh

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,179 @@
...
+for retrieving chunk indices in bulk for all validators at a given block and core.
+
+#### Upgrade path
+
+Considering that the Validator->Chunk mapping is critical to para consensus, the change needs to be enacted atomically via
+governance, only after all validators have upgraded the node to a version that is aware of this mapping.
+It needs to be explicitly stated that after the runtime upgrade and governance enactment, validators that run older client versions that don't
+support this mapping will not be able to participate in parachain consensus.
+
+### Getting access to core_index in different subsystems
```
sandreim, line 99: I think this section is all implementation details of the Parity Polkadot node and should not be here.

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,179 @@
...
+## Testing, Security, and Privacy
+
+Extensive testing will be conducted - both automated and manual.
+This proposal doesn't affect security or privacy.
+
+## Performance, Ergonomics, and Compatibility
+
+### Performance
+
+This is a necessary optimisation, as reed-solomon erasure coding has proven to be a top consumer of CPU time in polkadot when scaling the parachain count.
```
sandreim, line 155:
```suggestion
This is a necessary DA optimisation, as reed-solomon chunk reconstruction has proven to be a top consumer of CPU time as we scale up the parachain block size and number of availability cores.
```

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,179 @@
...
+
+With this optimisation, preliminary performance results show that CPU time used for reed-solomon coding can be halved and total POV recovery time decrease by 80% for large POVs. See more [here](https://github.com/paritytech/polkadot-sdk/issues/598#issuecomment-1792007099).
```
sandreim, line 157: Calling into the runtime to get chunk mappings is slower than doing that in the client, especially doing so many calls and building up overhead. I suggest only a bulk API, so it's just one call per candidate.

text/0047-random-assignment-of-availability-chunks.md
sandreim: I mean, just rephrasing to not refer to a particular architecture (subsystems).

text/0047-random-assignment-of-availability-chunks.md
alindima: There are places (like bitfield signing) that don't need to query the chunk indices of all the validators, since we only need to attest that we have our chunk.

That's why I think it's valuable to have both a bulk API and one for retrieving a single index

text/0047-random-assignment-of-availability-chunks.md
alindima: I rewrote the content in this section to refer to the protocol instead of the subsytems of the parity polkadot node

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,212 @@
...
+    3.b is initiated by the validator noticing dispute votes recorded on-chain. In this case, we can safely
+  assume that the backing event for that candidate has been recorded and kept in memory.
+  
+    3.c is initiated as a result of getting a dispute statement from another validator. It is possible that the dispute
+  is happening on a fork that was not yet imported by this validator, so the subsystem may not have seen this candidate
+  being backed.
+
+As a solution to 3.c, a new version for the disputes request-response networking protocol will be added.
+This message type will include the relay block hash where the candidate was included. This information will be used
+in order to query the runtime API and retrieve the core index that the candidate was occupying.
```
alindima, line 148: on a second thought: is this even possible? If the validator is not aware of the fork, how can it call a runtime API on that fork?

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,212 @@
...
+    chunk_indices.shuffle(&mut rng);
+
+    let core_start_pos = threshold * core_index.0;
+    return chunk_indices[(core_start_pos + validator_index) % n_validators]
+}
+```
+
+The function should be implemented as a runtime API, because:
+
+1. it's critical to the consensus protocol that all validators have a common view of the Validator->Chunk mapping.
```
ordian, line 101: so is approval-voting, which is off-chain

and in this case, it's critical for finality + parachain liveness, not safety

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,212 @@
...
+1. it enables further atomic changes to the shuffling algorithm.
+1. it enables alternative client implementations (in other languages) to use it
+1. mitigates the problem of third-party libraries changing the implementations of the `ChaCha8Rng` or the `rand::shuffle`
+that could be introduced in further versions, which would stall parachains. This would be quite an "easy" attack.
```
ordian, line 105: I didn't get this argument. Are you talking about supply chain attacks? How is it specific to this RFC?

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,212 @@
...
+    let mut rng: ChaCha8Rng = SeedableRng::from_seed(seed);
+    let mut chunk_indices: Vec<ChunkIndex> = (0..n_validators).map(Into::into).collect();
+    chunk_indices.shuffle(&mut rng);
+
+    let core_start_pos = threshold * core_index.0;
+    return chunk_indices[(core_start_pos + validator_index) % n_validators]
+}
+```
+
+The function should be implemented as a runtime API, because:
```
ordian, line 99: tbh, only reason 2 seems convincing to me and client features runtime API makes it less convincing

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,212 @@
...
+and have the following properties:
+1. deterministic
+1. pseudo-random
+1. relatively quick to compute and resource-efficient.
+1. when considering the other params besides `validator_index` as fixed,
+the function should describe a random permutation of the chunk indices
+1. considering `session_index` and `block_number` as fixed arguments, the validators that map to the first N/3 chunk indices should
+have as little overlap as possible for different cores.
+
+#### Proposed function and runtime API
```
ordian, line 76: I think we should either propose a runtime API or the (client-side) function implementation and I'm leaning towards the latter

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,212 @@
...
+#### Properties
+The function that decides the chunk index for a validator should be parametrised by at least
+`(validator_index, session_index, block_number, core_index)`
+and have the following properties:
+1. deterministic
+1. pseudo-random
+1. relatively quick to compute and resource-efficient.
+1. when considering the other params besides `validator_index` as fixed,
+the function should describe a random permutation of the chunk indices
+1. considering `session_index` and `block_number` as fixed arguments, the validators that map to the first N/3 chunk indices should
```
ordian, line 73: Thinking more about my argument to use `session_index` [here](https://github.com/paritytech/polkadot-sdk/issues/598#issuecomment-1734075024), I think it's a very niche edge-case that's not worth worrying about. An alternative to (`session_index`, `block_number`) would be block hash which would also sidestep that issue.

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,212 @@
...
+
+In this scheme, the relay chain node will first attempt to retrieve the N/3 systematic chunks from the validators that
+should hold them, before falling back to recovering from regular chunks, as before.
+
+### Chunk assignment function
+
+#### Properties
+The function that decides the chunk index for a validator should be parametrised by at least
+`(validator_index, session_index, block_number, core_index)`
+and have the following properties:
```
ordian, line 67: in other words what want is: create a uniformly distributed but deterministic mapping (shuffle) ValidatorIndex -> ChunkIndex per block per core

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,212 @@
...
+The usage of the systematic data availability recovery feature will also be subject to all nodes using the V2 disputes
+networking protocol.
+
+#### Alternatives to using core_index
+
+As an alternative to core_index, the `ParaId` could be used. It has the advantage of being readily available in the
+`CandidateReceipt`, which would enable the dispute communication protocol to not change and would simplify the
+implementation.
+However, in the context of [CoreJam](https://github.com/polkadot-fellows/RFCs/pull/31), `ParaId`s will no longer exist
+(at least not in their current form).
```
ordian, line 159: IMHO we shouldn't worry too much about CoreJam compatibility. IIUC there might be a drop-in replacement for `ParaId` (`AuthId`), so we should first explore an avenue with `ParaId` perhaps. 

The (bigger) problem with `ParaId` is that it's claimable by users, so in the theory one could create a collision attack where multiple paras have the same systematic chunks. In practice, I believe such an attack would be high cost and low benefit. And, perhaps, it could be mitigated by using a cryptographic hash of `ParaId` (or `AuthId` or `core_index`) in the future.

text/0047-random-assignment-of-availability-chunks.md
ordian: if the validator hasn't seen that fork, it's can't call into runtime API of that

ordian (commented):
I do agree we need such a mapping, but not convinced it needs to be a runtime API (the only benefit to me seems simplifying alternative client impls, but by not much).

Such a mapping doesn't enable systematic chunk recovery - it's possible already - but it spreads the load more evenly across validators (right now all systematic chunks are assigned to the same validators per block regardless of core id).

My suggestion is to use a hash of `ParaId` for now instead of `core_index`, which is readily available in the candidate receipt.

text/0047-random-assignment-of-availability-chunks.md
alindima: fair enough. this bullet-point should be coerced with the one talking about the supply chain attack

text/0047-random-assignment-of-availability-chunks.md
alindima: here's an example:

1. all validators are running normally, using systematic recovery.
2. a contributor on the `rand` crate (malicious or not) makes a perfectly valid change to the shuffling algorithm, that results in a different output for `shuffle()`.
3. there's a polkadot node release that bumps rand to this new version.
4. some of the validators upgrade their node.
5. as a result, some/all parachains are stalled, because some validators have a differing view of the validator->chunk mapping. Also, the validators that upgraded stop receiving rewards

IMO this has a high chance of happening in the future.

I view this mapping in a similar way to the per-session validator shuffle we do in the runtime to choose the validator active set.

text/0047-random-assignment-of-availability-chunks.md
alindima: maybe it wasn't clear. I was only proposing the runtime API here

text/0047-random-assignment-of-availability-chunks.md
alindima: yes. I'll add this

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,212 @@
...
+
+## Explanation
+
+### Systematic erasure codes
+
+An erasure coding algorithm is considered systematic if it preserves the original unencoded data as part of the
+resulting code.
+[The implementation of the erasure coding algorithm used for polkadot's availability data](https://github.com/paritytech/reed-solomon-novelpoly) is systematic. Roughly speaking, the first N_VALIDATORS/3
+chunks of data can be cheaply concatenated to retrieve the original data, without running the resource-intensive and
+time-consuming reconstruction algorithm.
```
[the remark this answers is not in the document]
alindima: I added the concatenation algorithm to the doc

text/0047-random-assignment-of-availability-chunks.md
alindima: Sounds fair. Another issue I just thought of with using block number is that for disputes hapenning on unimported forks, the ChainAPI call for getting the block number would also fail.

alindima:
Considering feedback and latest discoveries, I'll rewrite the proposal using ParaId instead of CoreIndex and using relay_parent_hash instead of (session_index, block_number)

text/0047-random-assignment-of-availability-chunks.md
alindima: now, we could implement our own version of `shuffle`, which would mitigate this issue.
The thing I'm concerned about is implementing our own version of `ChaCha8Rng`. How feasible is it to assume that it won't change in the future? CC: @burdges

alindima:
Rewritten the proposal. Have a look

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,283 @@
...
+			systematic_bytes.push(chunk[i]);
+			systematic_bytes.push(chunk[i + 1]);
+		}
+	}
+
+	Decode::decode(&mut &systematic_bytes[..]).map_err(|err| Error::Decode(err))
+}
+```
+
+In a nutshell, it performs a column-wise concatenation with 2-bit chunks.
```
ordian, line 68:
```suggestion
In a nutshell, it performs a column-wise concatenation with 2-byte chunks.
```

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,283 @@
...
+	chunks: Vec<&[u8]>,
+) -> T {
+	let mut threshold = (n_validators - 1) / 3;
+	if !is_power_of_two(threshold) {
+		threshold = next_lower_power_of_2(threshold);
+	}
+
+	let shard_len = chunks.iter().next().unwrap().len();
+
+	let mut systematic_bytes = Vec::with_capacity(shard_len * kpow2);
```
ordian, line 55:
```suggestion
	let mut systematic_bytes = Vec::with_capacity(shard_len * threshold);
```

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,283 @@
...
+
+	for i in (0..shard_len).step_by(2) {
+		for chunk in chunks.iter().take(kpow2) {
```
ordian, line 58:
```suggestion
		for chunk in chunks.iter().take(threshold) {
```

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,283 @@
...
+
+#### Proposed runtime API
+
+The mapping function should be implemented as a runtime API, because:
+
+1. it enables further atomic changes to the shuffling algorithm.
+1. it enables alternative client implementations (in other languages) to use it
+1. considering how critical it is for parachain consensus that all validators have a common view of the Validator->Chunk
+mapping, this mitigates the problem of third-party libraries changing the implementations of the `ChaCha8Rng` or the `rand::shuffle`
+that could be introduced in further versions. This would stall parachains if only a portion of validators upgraded the node.
```
ordian, line 124: We should obviously not rely in consensus on something that is not specified to be deterministic, like https://polkadot.network/blog/a-polkadot-postmortem-24-05-2021#the-good. And try to reduce third-party dependencies in general.
I would assume that a library that implements `ChaCha8Rng` adheres to the spec, otherwise it's a bug.
To mitigate supply-chain issues including bugs, we should probably use `cargo-vet` or a similar tool, but again, this is out of scope of and not limited to this RFC.

ordian (commented):
Thanks for addressing the comments. 

I am still not convinced we need runtime API for this, but curious what others think. If we put the idea to an extreme, we could put most of the client code into the runtime, so it can be automatically and atomically upgraded. But do we want to?

alindima:
Thanks for the reviews and suggestions Andronik!
Curious to see what others think as well!

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,283 @@
...
+## Drawbacks
+
+- In terms of guaranteeing even load distribution, a simpler function that chooses the per-core start position in the
+shuffle as `threshold * core_index` would likely perform better, but considering that the core_index is not part of the
+CandidateReceipt, the implementation would be too complicated. More details in [Appendix A](#appendix-a).
+- Considering future protocol changes that aim to generalise the work polkadot is doing (like CoreJam), `ParaId`s may be
+removed from the protocol, in favour of more generic primitives. In that case, `ParaId`s in the availability recovery
+process should be replaced with a similar identifier. It's important to note that the implementation is greatly simplified
+if this identifier is part of the `CandidateReceipt` or the future analogous data structure.
+- It's a breaking change that requires most validators to be upgrade their node version.
```
eskimor, line 197: superfluous `be`.

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,283 @@
...
+    .collect()
+}
+```
+
+#### Upgrade path
+
+Considering that the Validator->Chunk mapping is critical to para consensus, the change needs to be enacted atomically
+via governance, only after all validators have upgraded the node to a version that is aware of this mapping.
+It needs to be explicitly stated that after the runtime upgrade and governance enactment, validators that run older
+client versions that don't support this mapping will not be able to participate in parachain consensus.
```
eskimor, line 184: Does this need to be the case? As discussed, I think they should be able to still do the recovery from non systemic chunks.

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,283 @@
...
+This appendix explores alternatives to using the `ParaId` as the factor by which availability chunk indices are
+distributed to validators within the same relay chain block, and why they weren't chosen.
+
+### Core index
+
+Here, `core_index` refers to the index of the core that a candidate was occupying while it was pending availability
+(from backing to inclusion).
+
+Availability-recovery can currently be triggered by the following phases in the polkadot protocol:
+1. During the approval voting process.
```
eskimor, line 253: As long as recovery is possible in disputes, it should be fine if it can not always use systemic recovery. What is missing for me, is to understand why this needs to be an incompatible change. Shouldn't recovery be possible always, even if you don't know what the systemic chunks are?

ordian: Currently, when you request a chunk, you need to know which chunk to request. If you specify a wrong chunk, you'll get an empty response. So everyone should be onboard how chunks are shuffled in order for reconstruction to work, not just from systematic chunks. Unless we rely on backers, which we can't in case of a dispute.

We could add an additional request type to request any chunk you have. That could probably be a reasonable fallback in this case.

text/0047-random-assignment-of-availability-chunks.md
tomaka: This is off-topic for this RFC, but as a heads up we already do this: https://github.com/paritytech/polkadot-sdk/blob/2d09e83d0703ca6bf6aba773e80ea14576887ac7/polkadot/node/network/gossip-support/src/lib.rs#L601-L610

text/0047-random-assignment-of-availability-chunks.md
alindima: > We could add an additional request type to request any chunk you have. That could probably be a reasonable fallback in this case.

I don't think this achieves the purpose of availability-recovery working with an arbitrary mix of upgraded & non-upgraded validators. Adding a new request type (or even changing the meaning of the existing one) would also mean a node upgrade (because responding to the new type of request is only useful for the nodes that haven't yet upgraded to use the new mapping). So validators might as well just upgrade to the version that uses the new shuffle. I'm not sure this has any benefit

ordian: I meant as a fallback in case we want to use `core_id`s which might be not readily available in some cases of disputes.

alindima: > I meant as a fallback in case we want to use core_ids which might be not readily available in some cases of disputes.

I see. In this case, it would be a reasonable fallback. But this would mean giving up systematic recovery in those cases. I'm wondering if it's worth doing all of this just to replace the para-id thingy with the core_id

eskimor: As discussed in chat. If we keep requesting via `ValidatorIndex`, then in recovery you don't need to know the mapping, if you are fine with non systemic recovery.

text/0047-random-assignment-of-availability-chunks.md
burdges: We should use the core index here probably, not the para id.  I'd thought the core indexes could be spaced somewhat evenly, but with the actual sequence being random.

text/0047-random-assignment-of-availability-chunks.md
burdges: It's not a supply chain attack so much as rand having a history of instability.

text/0047-random-assignment-of-availability-chunks.md
burdges: yeah but they need to know what chunks map to where.

text/0047-random-assignment-of-availability-chunks.md
burdges: You need to know the chunk-validator mapping.  we may say the chunk index in the chunk itself, but you'd then think you could still get the same chunk for the guy who just gave you one.  In extreme cases, I bet this breaks reconstruction.

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,283 @@
...
+	for i in (0..shard_len).step_by(2) {
+		for chunk in chunks.iter().take(threshold) {
+			systematic_bytes.push(chunk[i]);
+			systematic_bytes.push(chunk[i + 1]);
+		}
+	}
+
+	Decode::decode(&mut &systematic_bytes[..]).map_err(|err| Error::Decode(err))
+}
+```
```
burdges, line 66: This could be done in the reed-solomn crate itself.

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,283 @@
...
+```rust
+pub fn get_chunk_index(
+  n_validators: u32,
+  validator_index: ValidatorIndex,
+  relay_parent: Hash,
+  para_id: ParaId
+) -> ChunkIndex {
+  let threshold = systematic_threshold(n_validators); // Roughly n_validators/3
+  let seed = derive_seed(relay_parent);
+  let mut rng: ChaCha8Rng = SeedableRng::from_seed(seed);
```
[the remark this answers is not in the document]
burdges: Avoid paraid here imho, not overly keen on relay parent either.

Instead use era/session, slot, and chain spec to define the validator sequence, and then core index to define the start position in the validator sequence.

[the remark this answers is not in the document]
eskimor: Validators are already randomized at the beginning of the session. Core index for start position makes sense. What is wrong with using the block number in addition? 

Reason, I would like to have it dependent on the block (could also be slot, I just don't see the benefit) is that by having a start position by core index, we ensure equal distribution of systemic chunks across a block, but paras are not all equal. Some could be heavier than others, hence it would be beneficial to change the mapping each block. 

In my opinion we could also use the hash instead of the block number, here I think anything is likely better than static.

text/0047-random-assignment-of-availability-chunks.md
eskimor: yep in distribution, for fetching I think the mapping ValidatorIndex -> ChunkIndex can be unknown, without any real issues.

text/0047-random-assignment-of-availability-chunks.md
eskimor: Indeed, we discussed this. The only issue is that you could not know whether you got the chunk you actually asked for. But that is only relevant if we wanted to change reputations. With that aside, we can just check whether we already fetched that chunk - if yes, treat it as invalid/garbage and move on to the next validator.

text/0047-random-assignment-of-availability-chunks.md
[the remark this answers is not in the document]
burdges: We want slot or block I think.  We're not overly adversarial here, but you can manipulate block numbers easily, which slot numbers represent an opportunity for doing something, so you always pay something to chose another one.  I'd say slot not block.

We randomize the validator list based upon the randomness two epochs/sessions ago?  Cool.  Any idea if we similarly randomize the map from paraids to cores per era/session too?  If yes, then maybe we could just progress sequentially through them?  
```
let k = num_validators / num_cores;
let fist_validator_index_for_core = ((core_id - slot) * k % num_validators) as u32;
```
We'd prefer the randomization of core_id for this because otherwise you could still make hot spots.  We could've hot spots even with this scheme, but not so easy to make them.  We'd avoid those if we randomize per slot.

Also this exactly computation does not work due to signed vs unsigned, and the fact that it suggests things progress backwards as time progresses time, which again tried to avoid hot spots.

text/0047-random-assignment-of-availability-chunks.md
burdges: It's worse than reputations.  I think it changes whether the 2/3rds honest assumption works correctly.

alindima: > It's worse than reputations. I think it changes whether the 2/3rds honest assumption works correctly.

Hm, how?

we currently don't modify any reputations within the availability-recovery subsystem. also, what @eskimor is describing above can happen with the current production code. the node does not check that it received the chunk it requested. It only checks that it's a valid chunk and that the proof is valid

burdges: The first 1/3rd of validators refuse to talk to you.  You obtain some chunks from most of the second 1/3rd, but they give you the chunks of the third 1/3rd, so you keep asking the second 1/3rd for the last chunks, which they do not have.  Reconstruction fails.

alindima: > You obtain some chunks from most of the second 1/3rd, but they give you the chunks of the third 1/3rd

Doesn't that mean the second third is also malicious? They don't adhere to the network protocol (which in polkadot's case is still specified by the reference implementation AFAICT).

We assume that (aside from the malicious third), all validators will respond with unique chunks. We don't care that they're not the ones they should/shouldn't hold, as long as they don't overlap. I think this is a reasonable assumption, isn't it? (we are already making this assumption in the current released code).

Leaving systematic recovery aside, the key aspect would be that we wouldn't ask for the chunk by its index. we would ask each validator for their chunk. The responders would map their validator index to the right chunk index. They would then respond with that chunk. Assuming at most a third of them are malicious and they all have the same mapping, we should get at least two thirds correct responses (and we only need a third. And for non-systematic recovery, we don't care if the responses are actually the ones we asked for, as long as we don't get duplicates.

For systematic recovery, we'd assume the validator initiating recovery has the mapping and requests chunks from the validators. Systematic recovery will then check that the chunks received are indeed the ones requested and fall back to regular behaviour if a peer misbehaves.

alindima: > No. We're discussing if it's okay for validators to have different mappings. I'm saying no.

I was discussing the case where validators all have the same mapping, but a validator participating in a dispute does not have access to the mapping at that point.

text/0047-random-assignment-of-availability-chunks.md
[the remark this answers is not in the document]
alindima: > Any idea if we similarly randomize the map from paraids to cores per era/session too?

AFAICT from the scheduler code, we don't (at least for the parachain auction model; for on-demand, I see more complicated logic which takes into account core affinities).

> We're not overly adversarial here, but you can manipulate block numbers easily, while slot numbers represent an opportunity for doing something, so you always pay 19 DOTs or more to chose another one. I'd say slot not block.

I'm having a hard time understanding the advantage of slot number vs block number. This may be simply because I don't know that much about slots. AFAICT, slots are equally useful as block number for the mapping function (monotonically increasing by 1), except that there may be slots that go unoccupied (if the chain is stalled for example) and are therefore skipped. Is that correct?

> so you always pay 19 DOTs or more to chose another one

what is this fee? where can I read more about this?

alindima:
Thanks to everybody for the feedback!
Based on this and especially my discussions with @eskimor , I will modify the RFC so that:

1. The shuffling function does not contain any randomness (or other complex third party libraries). This mitigates the problem of the `rand` crate changing implementations and enables us to not embed the algorithm in the runtime. We want an implementation that can be easily coded and has minimal points of failure in terms of updates/supply chain issues. It'll also make it easier for alternate clients to implement the same algorithm.
2. The shuffling function uses `core_index` instead of `para_id`. This enables us to easily implement an even distribution, because core indices form a monotonically increasing sequence. At the same time, core_index cannot be chosen by a parachain to influence the chunk assignment.
3. The `ChunkFetchingRequest` will not ask for a specific chunk index. Validators will respond with the chunk they should be holding according to the assignment function. This enables availability-recovery in scenarios where we don't know the mapping (for example for collators or disputes happening on unimported forks). As a result, the implementation of this RFC will no longer be a breaking change for collators.
4. Disputes protocol will not always be able to perform systematic recovery. Since not all disputes have easy access to the core_index occupied by a candidate, this would be very complicated to implement. We'll consider this acceptable, as it's a very rare circumstance. Regular systematic recovery will work as before.

Note that we'll keep asserting that enabling this feature needs to happen atomically between validators.

One thing that I've not decided on is the exact assignment function implementation. I think it could be (inspired by other comments):
```rust
pub fn get_chunk_index(
  n_validators: u32,
  validator_index: ValidatorIndex,
  block_number: BlockNumber,
  core_index: CoreIndex
) -> ChunkIndex {
  let threshold = systematic_threshold(n_validators); // Roughly n_validators/3
  let core_start_pos = abs(core_index - block_number) * threshold;

  (core_start_pos + validator_index) % n_validators
}
```

text/0047-random-assignment-of-availability-chunks.md
burdges: Ahh fine.  We'll loose the mappings sometime before the data expires?  We should probably avoid doing that.  We can obviously ask everyone and account for the DLs differently, but I'd prefer to be more like bittorrent trackers here, so that parachains can fetch data this way.

ordian:
> The ChunkFetchingRequest will not ask for a specific chunk index. Validators will respond with the chunk they should be holding according to the assignment function. This enables availability-recovery in scenarios where we don't know the mapping (for example for collators or disputes happening on unimported forks).

Do you mean introducing v3 version of the req-resp protocol?

> As a result, the implementation of this RFC will no longer be a breaking change for collators.

If the answer to the previous question is yes, I don't see how this is not a breaking change. If no, the receiving side is validating the chunk index matches the response, so I don't see how it's not a breaking change.

I think it would be reasonable to introduce req-resp v3 and later after some time enable this assignment via the runtime feature flag.

alindima:
> Do you mean introducing v3 version of the req-resp protocol?

No, I didn't want to add a new version to the protocol. The idea was that we assume all validators have upgraded to have the same validator->index mapping and we also assume that two thirds of them are honest and responded with the right chunk they should be holding according to the mapping (but we don't check that they responded with the right chunk, because we may not know it beforehand).

> And how would that work for availability when we need to get our chunk from backers?

That's a good point. Same as before. The validator would issue the request, and in this specific case also check the index.

The idea is that validators will usually check the indices they get (during availability-distribution or systematic recovery during approval checking), but the regular recovery will also work when the entity (collator or validator participating in a dispute) does not know/check the indices against the canonical mapping.

> If no, the receiving side is validating the chunk index matches the response, so I don't see how it's not a breaking change.

you're right, it's a breaking change. I guess what I meant is that, after the upgrade, regular recovery will still work even if the collator does not have access to the canonical mapping.

I'll reword my comment.

> I think it would be reasonable to introduce req-resp v3 and later after some time enable this assignment via the runtime feature flag.

Experimenting a bit more with what I proposed above, I realised that we either bump the protocol or we do some other arguably hacky way of discovering the chunk index on the receveing side (iterating through keys after reconstructing the partial trie from the proof and checking which value matches the chunk hash).

Considering that you called out the fact that even not bumping the protocol version would still be a breaking change to collators, I think it's best to bump the protocol version and add the `chunk_index` to the response.
This new protocol version would also have the semantic meaning that the requester MAY not always check the chunk index against the canonical mapping

text/0047-random-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,283 @@
...
+
+### Chunk assignment function
+
+#### Properties
+
+The function that decides the chunk index for a validator should be parameterized by at least
+`(validator_index, relay_parent, para_id)`
+and have the following properties:
+1. deterministic
+1. pseudo-random
```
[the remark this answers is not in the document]
alindima: agreed, I removed the randomness part

text/0047-random-assignment-of-availability-chunks.md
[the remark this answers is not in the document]
eskimor: @burdges I don't get your arguments. How can block numbers be manipulated? They are always increasing by 1, you clearly must be talking about forks. So an adversary could create multiple forks, with all the same load distribution and tries to overload validators this way? If we create forks or reversions we already have a performance problem anyway. 

Really not getting how slot numbers are better here. I also don't get your argument about hot spots, my proposal above was precisely to avoid hot spots (by not using randomness). What do you mean by hot spots and how would randomness help here?

A single validator not providing its systemic chunk would be enough to break systemic recovery. I don't see how randomization schemes help here. If we were smart we could somehow track the validators withholding systemic chunks and then make an assignment where they are all pooled together into one candidate. This way, at least only one systemic recovery fails. (We could equally well just remove them entirely.)

But honestly, I would not worry too much about this here. If we ever found that validators try to mess with this on purpose, the threat is low enough that social (calling them out) and governance would be adequate measures to deal with them.

text/0047-random-assignment-of-availability-chunks.md
eskimor: Yes. All validators will know the correct mapping, if they have seen (and still know) the relay parent. There is a good chance that this is the case even in disputes, but it is not guaranteed and recovery must work even in that case (but can be slower).

Using the core index should be fine, what makes that even more acceptable to me is that we actually want (and need to) include the core index in the candidate receipt at some point. So the moment we have this change, then the block availability will cease to matter.

eskimor:
> At the same time, core_index cannot be chosen by a parachain to influence the chunk assignment.

This is actually not true with agile core time: A parachain bids specifically on one particular core.

But it also should not matter much: We are rotating the mapping every block. On top of that, there is only so much a parachain can do here anyway.

> The ChunkFetchingRequest will not ask for a specific chunk index. Validators will respond with the chunk they should be holding according to the assignment function. This enables availability-recovery in scenarios where we don't know the mapping (for example for collators or disputes happening on unimported forks).

This is not what we discussed. I suggested to keep requesting via validator index. Right now we also store in the av-store per validator index - if we don't change that, then a node can provide a chunk as requested by validator index, even if it does not know the mapping itself*). Hence everything would keep working with neither the requester nor the responder still knowing the mapping.

*) Assuming the chunk data itself gets stored together with the chunk index.

Once we have the core index in the receipt, this would strictly speaking no longer be necessary, but I also don't see much harm _and_ if we really wanted to, we could change the network protocol again once we have this. Given that changing the candidate receipt is already a pita ... would not be that much of an additional annoyance.

text/0047-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,313 @@
...
+	pub index: ChunkIndex
+}
+```
+
+An important thing to note is that in version 1, the `ValidatorIndex` value is always equal to the `ChunkIndex`.
+Until the feature is enabled, this will also be true for version 2. However, after the feature is enabled,
+this will generally not be true.
+
+The requester will send the request to validator with index `V`. The responder will map the `V` validator index to the
+`C` chunk index and respond with the `C`-th chunk.
```
eskimor, line 197: The responder does not even need to do that, if we keep storing per [validator index in the av store](https://github.com/paritytech/polkadot-sdk/blob/fc12f435e3796d33018aab93a9e8cd851d4431d5/polkadot/node/subsystem-types/src/messages.rs#L504).

text/0047-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,313 @@
...
+
+The protocol implementation MAY check the returned `ChunkIndex` against the expected mapping to ensure that
+it received the right chunk.
```
eskimor, line 200: In any case the requester has to verify the chunk's correctness using the provided index. Then given that it can not know (if the mapping is not available) whether the received chunk corresponds to the requested index, it has to keep track of received chunk indices and check for duplicates. (Duplicates should be considered the same as an invalid/garbage response (we drop it and move on to the next validator - we can not punish as in reputation changes, because we don't know which validator misbehaved.)

Ideally see above we can make this all more robust with an updated candidate receipt.

text/0047-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,313 @@
...
+for easy access to the validator->chunk mapping. See [Appendix A](#appendix-a) for more details.
+
+
+### Upgrade path
+
+#### Step 1: Enabling new network protocol
+In the beginning, both `/polkadot/req_chunk/1` and `/polkadot/req_chunk/2` will be supported, until all validators and
+collators have upgraded to use the new version. V1 will be considered deprecated.
+Once all nodes are upgraded, a new release will be cut that removes the v1 protocol. Only once all nodes have upgraded
+to this version will step 2 commence.
```
eskimor, line 212: In this step the mapping will still be 1:1, regardless of protocol.

text/0047-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,313 @@
...
+via governance, only after all validators have upgraded the node to a version that is aware of this mapping.
+It needs to be explicitly stated that after the runtime upgrade and governance enactment, validators that run older
+client versions that don't support this mapping will not be able to participate in parachain consensus.
+
+Additionally, an error will be logged when starting a validator with an older version, after the runtime was upgraded
+and the feature enabled.
+
+On the other hand, collators will not be required to upgrade, as regular chunk recovery will work as before, granted
+that version 1 of the networking protocol has been removed.  However, they are encouraged to upgrade in order to take
+advantage of the faster systematic recovery.
```
eskimor, line 225: I think collators have to upgrade as well.

text/0047-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,313 @@
...
+
+#### Proposed function
+
+Pseudocode:
+
+```rust
+pub fn get_chunk_index(
+  n_validators: u32,
+  validator_index: ValidatorIndex,
+  block_number: BlockNumber,
```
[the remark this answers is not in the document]
eskimor: Ok using the relay parent of the candidate is at the very least not ideal because with async backing, the used relay parent could vary from candidate to candidate in the same block, which means that the equal load distribution might end up not being that equal after all.

burdges:
>  >   At the same time, core_index cannot be chosen by a parachain to influence the chunk assignment.
>
>  This is actually not true with agile core time: A parachain bids specifically on one particular core.
>
>  But it also should not matter much: We are rotating the mapping every block. On top of that, there is only so much a parachain can do here anyway.

It's likely fine.  We want the systemic chunks spread evenly enough to minimize other protocol manipulation.  It's possible some parachains chose their cores to make frierndly validators more profitable, which overworks some other validators, and causes reconstructions.  It's probably fine though.

text/0047-assignment-of-availability-chunks.md
alindima: correct, I'll add this detail

text/0047-assignment-of-availability-chunks.md
alindima: Why?

I think they just need to upgrade to the v2 networking protocol. Once that's done, they don't really need to upgrade to use systematic recovery. They'll request regular chunks as before.
That's why I added this bit only on the step 2 (which enables systematic recovery). Upgrade for step 1 will still be needed

text/0047-random-assignment-of-availability-chunks.md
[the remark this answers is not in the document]
alindima: > A single validator not providing its systemic chunk would be enough to break systemic recovery

Slightly offtopic: In my WIP PR I added functionality to request up to 5 chunks from the backing group as a backup solution, so that a couple of validators not returning their systematic chunks would not invalidate the entire procedure

ordian (commented):
Left small remarks, looking good otherwise

text/0047-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,313 @@
...
+
+The request-response `/polkadot/req_chunk` protocol will be bumped to a new version (from v1 to v2).
+For v1, the request and response payloads are:
+```rust
+/// Request an availability chunk.
+pub struct ChunkFetchingRequest {
+	/// Hash of candidate we want a chunk for.
+	pub candidate_hash: CandidateHash,
+	/// The index of the chunk to fetch.
+	pub index: ValidatorIndex,
```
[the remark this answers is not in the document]
alindima: this will stay the same, indeed. The meaning will change, in the sense that we cannot expect any more that the ValidatorIndex will be equal to the returned ChunkIndex. I hope I describe this in sufficient detail in the following paragraph

text/0047-random-assignment-of-availability-chunks.md
alindima: this discussion is obsolete, as the current proposal does not include a runtime API for the mapping function

text/0047-assignment-of-availability-chunks.md
eskimor: >  think they just need to upgrade to the v2 networking protocol.

Ok, to me that sounded like an upgrade. In other words: If you already have the requirement that they upgrade their network protocol, the rest does not sound like an issue anymore.

text/0047-assignment-of-availability-chunks.md
alindima: right. yeah, that makes sense now 👍🏻

text/0047-assignment-of-availability-chunks.md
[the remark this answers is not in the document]
alindima: @burdges WDYT of this?

text/0047-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,337 @@
...
+	let mut systematic_bytes = Vec::with_capacity(shard_len * threshold);
+
+	for i in (0..shard_len).step_by(2) {
+		for chunk in chunks.iter().take(threshold) {
+			systematic_bytes.push(chunk[i]);
+			systematic_bytes.push(chunk[i + 1]);
+		}
+	}
+
+	Decode::decode(&mut &systematic_bytes[..]).unwrap()
```
burdges, line 59: Does this remove the end zero padding if odd length?

Anyways we should be careful about the boundary here, like maybe this should live in the erasure coding crate.

text/0047-assignment-of-availability-chunks.md
```diff
@@ -0,0 +1,337 @@
...
+}
+
+fn systematic_threshold(n_validators: usize) -> usize {
+	let mut threshold = (n_validators - 1) / 3;
+	if !is_power_of_two(threshold) {
+		threshold = next_lower_power_of_2(threshold);
+	}
```
burdges, line 66: Yes, next lower power of two is correct here.

We encode `2^{x+1}` chunks where `2^x < n_validators <= 2^{x+1}` of course.  We need `threshold < (n_validators-epsilon)/3` though, so assuming `epsilon=1` then we reconstruct from `2^{x-1}` chunks where `2^{x-1} <= (n_validators-1) / 3 < 2^x`.

If `n_validators = 2^{x+1}`, so no wasted encoding, then we reconstruct from `n_validators / 4` chunks, and our reconstruction threshold winds up `(n_validators - 4) / 12` smaller than necessary.  If otoh `(n_validators-1)/3 = 2^{x-1}`, so optimal reconstruction threshold, then `n_validators = 3 * 2^{x-1} + 1` unecessarily encodes `2^{x-1} - 1` extra chuinks.

text/0047-assignment-of-availability-chunks.md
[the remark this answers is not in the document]
burdges: I'd expect the opposite:  After rewards exist, you choose your core so that friendly validators have more systemic chunk assignments.  It annoys some validators who become slightly less profitable.   It annoys polkadot because then unprepared validators could recieve more systemic chunk assignments than they could handle.  It's probably fine since those unprepared validatoprs have the backers as systemic chunks backup, not just themselves.

A simple rotation suffices here imho since any adversary must specifically code the rotation into their core purchasing, so anyone not doing this gets into their way, but they can rotate very coarsely I guess.

If we do not rotate then some cores could naturally do this, which could prompt some stupid behavior within the community.  Ain't too big a deal, but maybe easier to rotate by slot, unless we've some technical reason otherwise.

text/0047-assignment-of-availability-chunks.md
alindima: yes, scale decoding ignores the trailing zeros

alindima: In order to know the exact number of zeroed bytes added as padding, we need to know the size of the input data that was encoded.
Unfortunately, we don't have easy access to that in polkadot, unless we add it to the `CandidateReceipt`.

But scale decoding it works

alindima: I ran some roundtrip quickcheck tests on regular chunk reconstruction. The regular reed-solomon code can have extra zeroed padding when reconstructing. So truncation to the expected size was already needed.

text/0047-assignment-of-availability-chunks.md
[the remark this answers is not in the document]
alindima: > Ain't too big a deal since again 1/3 gets pretty coarse, but maybe easier to rotate by slot, unless we've some technical reason otherwise.

rotating by block number/slot will require that the chunk request requester has access to the relay parent of the candidate if they want to perform systematic recovery.
Considering that core index will be added to the candidate receipt, if we were to only use the core index for the rotation, we would only need the `CandidateReceipt` in order to perform systematic recovery or get our assigned chunk. Which would be easier and would also enable systematic recovery for all dispute participants that may not have access to the relay chain blocks for remote disputes.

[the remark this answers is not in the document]
eskimor: I would really appreciate if all information could be self contained. @burdges I don't get your argument. Systemic chunks will be distributed fairly equally among cores in a block. We don't need rotation for this: First core, first x validators are systemic chunk providers, next core , next x validators, .... 

The only difference is, if we assume that some cores are predictably higher loaded than others (PoVs larger in expectation). This can happen, but could be mitigated by incorporating in rewards: You reward people in proportion to the size of the chunks they served. Should be possible to design this in a way, so there is no meaningful advantage in trying to manipulate for which cores your validators (or others) are responsible for.

In general we should strive for (and agile coretime is one important step into that direction), that cores are fully loaded as much as possible: Better allocation efficiency. If we achieve that goal, then having no rotation should not matter at all anymore.

Also in case this ever proves to be a real problem still, we could adjust the scheduler to move parachains between cores all the time, which would then still give us self contained systemic recovery, with evened out load. I don't think we we will ever need that, just pointing out that we could achieve rotation via different means, which don't affect availability recovery at all.

There exists some uneven load, even if all cores are equal, but:

With 100 cores, two thirds of the validators will be systemic chunk providers for 33 cores. One third of them will be systemic chunk providers for 34 cores. That is as bad as it gets, when it comes to unequal load in number of cores directly. It also has the good property that this difference gets smaller with more cores. We can also even it out, by always increasing core count in steps of three, if we wanted to get rid of this entirely.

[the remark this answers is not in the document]
burdges: Isn't the relay parent given in the candidate reciept?  

> I would really appreciate if all information could be self contained.

It's unclear what you mean here.  We'll have the candidate reciept when fetching data, no?

We'd communicate something less than the full candidate reciept for off-chain messaging I guess.

> We don't need rotation for this: First core, first x validators are systemic chunk providers, next core , next x validators, ....

We absolutely require rotation of backing groups.   I'd give systemic chunks a different rotation, but not sure that matters.

We do want cores loaded as equally as possible of course.  We've at least two dimensions here, space and cpu time, so we'll have trouble even defining equally correctly, much less achieving it while we've this secondary market. 

> This can happen, but could be mitigated by incorporating in rewards: You reward people in proportion to the size of the chunks they served.  Should be possible to design this in a way, so there is no meaningful advantage in trying to manipulate for which cores your validators (or others) are responsible for.

Rewards are complex, if not chaotic.  Avoid complicating rewards unecessarily.

We've never discussed availability rewards being proportional to size.  Yes, proportional sounds interesting & nice.  Yet, rewards involve imprecissions or inequalities already, so proportional sounds like over commitment.  I'm unsure how proportional-to-size fits into tit-for-tat, for example. 

Rewards could acquire new bizarre constriants in future, via politics or competition, ala Cardano proves some Nash equilibrium, or craziness like cross-chain MEV, or stupidity like quadratic voting.  Also, rewards have bizarre conditionals because validators have different situations. 

Rewards solve major problems that overarch the whole project.  Rewards do not fine tune small problems. 

> Also in case this ever proves to be a real problem still, 

Would we notice?

> we could adjust the scheduler to move parachains between cores all the time, which would then still give us self contained systemic recovery, with evened out load. 

Yes, but isn't this much more complex than what we're discussing here?

> That is as bad as it gets

Yes, we're maybe polishing rocks here.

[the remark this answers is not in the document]
burdges: Anyways..

Are you worried about slot numbers in off-chain messaging?

Or is there some simpler reason why slot numbers pose a problem?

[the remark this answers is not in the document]
alindima: > It's unclear what you mean here. We'll have the candidate reciept when fetching data, no?

Yes, and we'd like availability-recovery to be possible when having nothing more than the candidate receipt. The problem is that the receipt does not contain any info about the slot or block number.
For the moment, the core index isn't there either, but the plan is to add it.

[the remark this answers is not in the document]
burdges: I see, thank you.  We could only use some historical slot, not the slot where the candidate recipet gets backed.

All these options give some control of course, even the slot where backed, so relay parent hash works better than slots, due to simplicity. 

As I understand it @eskimor suggests our map depend upon only num_cores, num_validators, and core index, so each core has their systemic validators fixed for the session.  It avoids bias except through selecting your core.  I'd wager it worsens user experence, but only slightly. 

We do however rotate backing groups and backers provide systemic chunks too, hence the slightly above.  How do we determin the backers from the candidate recipet?  Just because they signed the candidate recipet? 

It's likely fine either way.  We'll anyways have systemic reconstruction if each systemic chunk can be fetched from some backers or its one availability provider. 

Actually who determines core index?  We'd ideas where this occurs after the candidate reciept.  We'd enable these if the map depends upon only num_cores, num_validators, relay parent, and paraid.

text/0047-random-assignment-of-availability-chunks.md
alindima: done: See https://github.com/paritytech/reed-solomon-novelpoly/pull/36

text/0047-random-assignment-of-availability-chunks.md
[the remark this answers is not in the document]
burdges: We could permit requesting them all from the backing group, but make the availability providers the first choice, meaning maybe:  1st, try all systemic chunk providers.  2nd, try remaining systemic chunks from backers.  3rd, fetch random non-systemic chunks.  The concern is just that we overload the backers.

It'll also make rewards somewhat more fragile, but likely worth the difficulty for the performance.

text/0047-random-assignment-of-availability-chunks.md
burdges: I changed my view here above in https://github.com/polkadot-fellows/RFCs/pull/47#discussion_r1433047970  

It's probably more future proof if we do not require the core index here, so the map uses num_cores, num_validators, relay parent, and paraid.  We've other options but this looks pretty future proof.  CoreJam cannot really work without something like paraid.

text/0047-assignment-of-availability-chunks.md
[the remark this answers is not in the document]
alindima: > All these options give some control of course, even the slot where backed, so relay parent hash works better than any slot choice, due to simplicity.

As @eskimor said, "using the relay parent of the candidate is at the very least not ideal because with async backing, the used relay parent could vary from candidate to candidate in the same block, which means that the equal load distribution might end up not being that equal after all."

> As I understand it @eskimor suggests our map depend upon only num_cores, num_validators, and core index, so each core has their systemic validators fixed for the session. It avoids bias except through selecting your core. I'd wager it worsens user experence, but only slightly.

yes, that's the suggestion AFAIU.

> How do we determin the backers from the candidate recipet? Just because they signed the candidate recipet?

No, we only use the backers currently during approval-voting, when we have access to the relay block and we can query the session's validator groups from the runtime. For complexity reasons, we don't even verify that validators in the group all backed the candidate. We just assume that to be true (in practice, it's mostly true).

> Actually who determines core index? We'd ideas where this occurs after the candidate reciept. We'd enable these if the map depends upon only num_cores, num_validators, relay parent, and paraid.

Currently, for the existing lease parachains, there's a fixed assignment between the paraid and the core index. @eskimor you know more here

text/0047-random-assignment-of-availability-chunks.md
alindima: Using paraid was my first attempt, but it complicates things quite a bit.
As Andronik notes, it's claimable by a user. I don't think it's the biggest problem though, they can't get a different ParaId every day (can they?).

`paraid`'s biggest problem is that it doesn't form a strictly monontonically increasing sequence, so we'd have to do something else with it. My initial attempt was to seed a ChaCha8 RNG with it and generate one random number. Then this number was the offset into the chunk index vector.

But then if we complicate it so much, we may want to embed it into the runtime, so that future updates to the `rand` crate don't break us (which creates all sorts of problems that could be avoided).

We've had this paraid -> core_index back and forth for a while now, it'd be great if we could all (the ones interested in this RFC) hop on a call or somehow reach a conclusion. There are pros and cons to both, AFAICT.

I think this remains the only bit that prevents this RFC from moving on (anybody correct me if I'm wrong)

text/0047-assignment-of-availability-chunks.md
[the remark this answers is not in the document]
eskimor: Let's just go with only the core index, from the discussion I don't see any real problems with that and once we have the core index in the candidate receipt we are golden from the simplicity/robustness perspective.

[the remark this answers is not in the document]
alindima: sounds good, I'll update the document

alindima:
I think that the RFC is in a form that can be proposed and voted on. Please give another review and approve it if you think so.

text/0047-assignment-of-availability-chunks.md
alindima: I thought about this a bit more.

The backing subsystem will issue a request to av-store to store all chunks. For that, we need to know the core index in order to store the ValidatorIndex -> Chunk as you suggest (so that we can compute the mapping).

Since we don't yet have the core index in the receipt, the backing subsystem needs to know the per-relay parent core_index assignment of the local validator.
From my knowledge, this would be just fine. When doing attestation, the availabiliy_cores runtime API already gets the core_index for us (but doesn't store it yet).

The slight caveat is that, when importing a statement, we may also have to call the availability_cores runtime API to see which core our para has been scheduled on. but it's no big deal, we need to have access to the relay parent anyway when doing candidate validation.

@eskimor please weigh in on my analysis. until elastic scaling, a para id couldn't be scheduled on multiple cores and the core assignment could only change on a relay-block boundary. And when we'll get elastic scaling, we'll have the core_index in the receipt anyway. so all good.

eskimor (approved):

ordian (approved):

alindima:
/rfc process 0xe750513eb583410686c094e8e5f00c0b18dc33e0296c1c8572c2e83f021c03a5

PR #48, opened 2023-11-13
Generate ownership proof for `SessionKeys`

bkchr:
This RFC changes the `SessionKeys` runtime api to support generating a proof of ownership of the generated session keys.

### DIFF
--- a/text/0048-session-keys-runtime-api.md
+++ b/text/0048-session-keys-runtime-api.md
@@ -3,23 +3,29 @@
 |                 |                                                                                             |
 | --------------- | ------------------------------------------------------------------------------------------- |
 | **Start Date**  | 13 November 2023                                                                    |
-| **Description** | Change `SessionKeys` runtime api to also create a proof of ownership for on chain registration. |
+| **Description** | Change `SessionKeys` runtime api to support generating an ownership proof for the on chain registration. |
 | **Authors**     | Bastian Köcher                                                                    |
 
 ## Summary
 
-When rotating/generating the `SessionKeys` of a node, the node calls into the runtime using the 
-`SessionKeys::generate_session_keys` runtime api. This runtime api function needs to be changed
-to add an extra parameter `owner` and to change the return value to also include the `proof` of 
-ownership. The `owner` should be the account id of the account setting the `SessionKeys` on chain 
-to allow the on chain logic the verification of the proof. The on chain logic is then able to proof 
-the possession of the private keys of the `SessionKeys` using the `proof`.
+This RFC proposes to changes the `SessionKeys::generate_session_keys` runtime api interface. This runtime api is used by validator operators to
+generate new session keys on a node. The public session keys are then registered manually on chain by the validator operator. 
+Before this RFC it was not possible by the on chain logic to ensure that the account setting the public session keys is also in 
+possession of the private session keys. To solve this the RFC proposes to pass the account id of the account doing the 
+registration on chain to `generate_session_keys`. Further this RFC proposes to change the return value of the `generate_session_keys` 
+function also to not only return the public session keys, but also the proof of ownership for the private session keys. The 
+validator operator will then need to send the public session keys and the proof together when registering new session keys on chain.
 
 ## Motivation
 
-When a user sets new `SessionKeys` on chain the chain can currently not ensure that the user 
-actually has control over the private keys of the `SessionKeys`. With the RFC applied the chain is able 
-to ensure that the user actually is in possession of the private keys.
+When submitting the new public session keys to the on chain logic there doesn't exist any verification of possession of the private session keys.
+This means that users can basically register any kind of public session keys on chain. While the on chain logic ensures that there are 
+no duplicate keys, someone could try to prevent others from registering new session keys by setting them first. While this wouldn't bring
+the "attacker" any kind of advantage, more like disadvantages (potential slashes on their account), it could prevent someone from 
+e.g. changing its session key in the event of a private session key leak.
+
+After this RFC this kind of attack would not be possible anymore, because the on chain logic can verify that the sending account 
+is in ownership of the private session keys.
 
 ## Stakeholders
 
@@ -35,9 +41,10 @@ type Proof = (Signature, Signature, ..);
 ```
 
 The `proof` being a SCALE encoded tuple over all signatures of each private session 
-key signing the `owner`. The actual type of each signature depends on the
+key signing the `account_id`. The actual type of each signature depends on the
 corresponding session key cryptographic algorithm. The order of the signatures in 
-the `proof` is the same as the order of the session keys in the `SessionKeys` type.
+the `proof` is the same as the order of the session keys in the `SessionKeys` type 
+declared in the runtime.
 
 The version of the `SessionKeys` needs to be bumped to `1` to reflect the changes to the 
 signature of `SessionKeys_generate_session_keys`:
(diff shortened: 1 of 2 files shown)


text/0048-session-keys-runtime-api.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Explanation
+
+We are first going to explain the `proof` format being used:
+```rust
+type Proof = (Signature, Signature, ..);
+```
+
+The `proof` being a SCALE encoded tuple over all signatures of each private session 
+key signing the `owner`. The actual type of each signature depends on the
```
tomaka, line 38: So the message being signed is the SCALE-encoded `account_id`?

text/0048-session-keys-runtime-api.md
```diff
@@ -0,0 +1,102 @@
...
+
+The version of the `SessionKeys` needs to be bumped to `1` to reflect the changes to the 
+signature of `SessionKeys_generate_session_keys`:
+```rust
+pub struct OpaqueGeneratedSessionKeys {
+	pub keys: Vec<u8>,
+	pub proof: Vec<u8>,
+}
+
+fn SessionKeys_generate_session_keys(owner: Vec<u8>, seed: Option<Vec<u8>>) -> OpaqueGeneratedSessionKeys;
```
tomaka, line 50: An `AccountId` is a u64 I believe. How does that translate to a `Vec<u8>`?
Since the function signature here is the "high level one", shouldn't it be `u64` here?

text/0048-session-keys-runtime-api.md
```diff
@@ -0,0 +1,102 @@
...
+to add an extra parameter `owner` and to change the return value to also include the `proof` of 
+ownership. The `owner` should be the account id of the account setting the `SessionKeys` on chain 
+to allow the on chain logic the verification of the proof. The on chain logic is then able to proof 
+the possession of the private keys of the `SessionKeys` using the `proof`.
+
+## Motivation
+
+When a user sets new `SessionKeys` on chain the chain can currently not ensure that the user 
+actually has control over the private keys of the `SessionKeys`. With the RFC applied the chain is able 
+to ensure that the user actually is in possession of the private keys.
```
tomaka, line 22: Worded differently, you want to give the possibility to the implementation of `SessionKeys_generate_session_keys` needs to be able to know which `AccountId` the client controls?

text/0048-session-keys-runtime-api.md
burdges: You'd presumably sign the actual controller key here, not some reference.

burdges:
In this RFC, all session keys first acknowledge their controller key by signing it, after which their controller key certifies them.  Yes, we do obtain the desired back certificates this way, but I'd envisioned doing this in part the other way round, plus adding a master session key.  

It'd look like this:
0. Add some master session key, maybe the grandpa key.
1. All other subordinate session keys back certify this master session key 
2. The master session key certifies all other subordinate session keys, along with their certificates of the master session key, probably as pairs (key.
3. Operator certifies the master session key with their controller key.
4. At node start, the master session key certifies the controller key along with both its certificate of the master session key and a random nonce. We call this the active certificate.
5. Pre-sync, the starting node announces its certificate collection, including this new one created upon startup.  
6. An already active synced node posts the active certificate on-chain via some intrinsic, replacing the previous one.
7. The starting node waits until its active certificate get finalized before running protocols with equivocation slashing, like grandpa or babe.  
8. Any node shuts down if its active certificate get replaced on-chain.

In emergencies, a command line option could overrides 7, and maybe this could happen automatically, but usually 7 prevents nodes being slashed for operator error.

In this, 1-2 have same orientation as proposed here, but 3-4 have the reversed orientation.  We must check a two step certificate chain, including two back certificates, before believing a session key, but this should be a rare enough it does not matter.  Our nonce in 4 could be some ephemeral key like maybe the transport layer key.

Anyways..

If there is a pressing need to do something simpler first then fine, and validator operators rarely get slashed for equivocations these days, but it's worth defending against the equivocations footgun like this.  A master session key permits changing session keys without impacting user experience for validator operators, and consolidates the nonce lock under one session key, but this could be done without the master session key.

text/0048-session-keys-runtime-api.md
bkchr: > An `AccountId` is a u64 I believe. How does that translate to a `Vec<u8>`?

`AccountId` for Polkadot is an enum with public keys of sr/ed25519 or ecdsa. Other chain can also have different account ids. Thus, this should be a `Vec<u8>` that represents the SCALE encoded account id.

text/0048-session-keys-runtime-api.md
bkchr: Yes

text/0048-session-keys-runtime-api.md
bkchr: Yes. I mean `generate_session_keys` doesn't care as it just signs whatever you pass it.

eskimor: At least for me, the motivation does not go deep enough. Why should the chain care?

bkchr: > Why should the chain care?

Care that you are in possession of the keys? The chain doesn't really need to care. However, someone could may include a transaction faster than you with your public keys registered. So, someone could theoretically prevent that a validator is rotating its keys.

eskimor: LOL. Ok, makes sense! Thanks!

text/0048-session-keys-runtime-api.md
```diff
@@ -0,0 +1,115 @@
...
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 13 November 2023                                                                    |
+| **Description** | Change `SessionKeys` runtime api to support generating an ownership proof for the on chain registration. |
+| **Authors**     | Bastian Köcher                                                                    |
+
+## Summary
+
+This RFC proposes to changes the `SessionKeys::generate_session_keys` runtime api interface. This runtime api is used by validator operators to
```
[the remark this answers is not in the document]
bkchr: Because the node doesn't know what session keys the runtime is using.

text/0048-session-keys-runtime-api.md
```diff
@@ -0,0 +1,115 @@
...
+The interfaces have been optimized to make it as easy as possible to generate the ownership proof.
+
+### Compatibility
+
+Introduces a new version of the `SessionKeys` runtime api. Thus, nodes should be updated before 
+a runtime is enacted that contains these changes otherwise they will fail to generate session keys. 
+The RPC that exists around this runtime api needs to be updated to support passing the account id 
+and for returning the ownership proof alongside the public session keys.
+
+UIs would need to be updated to support the new RPC and the changed on chain logic.
```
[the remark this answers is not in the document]
bkchr: There is a new RPC for this. So, they will not break. However when the runtime is upgraded it will require the proofs which only the new UI generates. There is also only polkadot-js which supports this AFAIK.

text/0048-session-keys-runtime-api.md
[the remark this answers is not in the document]
davxy: Instead of generating the keys in the runtime, which in fact ends up generating the keys via some host call would be nice to have instead the runtime returning a list of `(CryptoTypeId, KeyTypeId)`.
Then generate the keys directly using the keystore.

The advantage of this approach would be to deprecate the host calls to generate the keys from the runtime which (IMO) is a bit awkward.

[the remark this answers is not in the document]
bkchr: These host calls are also required for benchmarking to generate private keys. `CryptoTypeId` didn't exist at the time this runtime api was introduced. This is also really the topic of this RFC. 

In the end you also still need to call into the runtime to do this. In hindsight I would also design some of the host calls differently to make nodes for example "more generic". This can be done, but yeah isn't part of this RFC and this RFC will also not make any of that more complicated.

burdges:
We'd reduce slashing risks considerably if we handle this like I propsed above:
https://github.com/polkadot-fellows/RFCs/pull/48#issuecomment-1809483250

bkchr:
/rfc process 0x34f2bcbf3ef8b5ef7d7ab74cef6d391c1f6aee227daa5a67108f9e9c8ffd00aa

PR #50, opened 2023-11-17
Fellowship Salaries

joepetrowski:
The Fellowship Manifesto states that members should receive a monthly allowance on par with gross income in OECD countries. This RFC proposes concrete amounts.


xlc (commented):
presumably we want to update the amount yearly. should we have a section describe how the amount get updated?

text/0050-fellowship-salaries.md
```diff
@@ -0,0 +1,114 @@
...
+|    II |  20,000 |      11 |   220,000 |  18,333 |
+|   III |  80,000 |       8 |   640,000 |  53,333 |
+|    IV | 120,000 |       3 |   360,000 |  30,000 |
+|     V | 160,000 |       5 |   800,000 |  66,667 |
+|    VI | 200,000 |       3 |   600,000 |  50,000 |
+|  > VI | 200,000 |       0 |         0 |       0 |
+|       |         |         |           |         |
+| Total |         |         | 2,890,000 | 240,833 |
+
+Note that these are the maximum amounts; members may choose to take a passive (lower) level. On the
```
iammasterbrucewayne, line 78: Are there any specific guidelines for how the amounts will be determined for passive (lower) levels or is it up to the members to propose their salaries as they see fit?

joepetrowski: Right now in the Salary pallet's implementation, passive is just zero, as in you don't claim. But that seems like a pretty simple configuration addition.

text/0050-fellowship-salaries.md
```diff
@@ -0,0 +1,114 @@
...
+
+| Dan   | Salary  | Members | Yearly    | Monthly |
+|:-----:|:-------:|:-------:| ---------:| -------:|
+|     I |  10,000 |      27 |   270,000 |  22,500 |
+|    II |  20,000 |      11 |   220,000 |  18,333 |
+|   III |  80,000 |       8 |   640,000 |  53,333 |
+|    IV | 120,000 |       3 |   360,000 |  30,000 |
+|     V | 160,000 |       5 |   800,000 |  66,667 |
+|    VI | 200,000 |       3 |   600,000 |  50,000 |
+|  > VI | 200,000 |       0 |         0 |       0 |
```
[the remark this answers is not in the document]
joepetrowski: It's "greater than 6"

text/0050-fellowship-salaries.md
```diff
@@ -0,0 +1,118 @@
...
+|    V |    160,000    |
+|   VI |    200,000    |
+|  VII |    200,000    |
+| VIII |    200,000    |
+|   IX |    200,000    |
+
+Note that there is a sizable increase between II Dan (Proficient) and III Dan (Fellow). By the third
+Dan, it is generally expected that one is working on Polkadot as their primary focus in a full-time
+capacity. The salary at this level is commensurate with average salaries in OECD countries (note:
+77,000 USD in the U.S., with an average engineer at 100,000 USD).
```
ggwpez, line 56: Can we make it a bit more technical, to allow for easy evaluation and adjustments?  

It looks like the 100K are in the center between DAN 3 and 4. Then we can use that as baseline and calculate the other ones in relation to that. Eg DAN 5 is DAN 4 +33% and DAN 2 is -75% of DAN 3 and so on.

text/0050-fellowship-salaries.md
```diff
@@ -0,0 +1,118 @@
...
+
+The salary increases for Architects (IV, V, and VI Dan) with levels typical of senior engineers.
+
+Allowances will be managed by the Salary pallet.
+
+### Projections
+
+Based on the current membership, the maximum yearly and monthly costs are shown below:
```
ggwpez, line 64: Maximum as long as there is no change to the members?

text/0050-fellowship-salaries.md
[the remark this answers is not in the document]
eskimor: ha missed the > ;-)

text/0050-fellowship-salaries.md
joepetrowski: I added a note a [few lines down](https://github.com/polkadot-fellows/RFCs/pull/50/files#diff-0b1740c525b5f4d411e0c5f9e2adac79bd34bf5fd9a2f30907aee0ffa5b437e0R78) that membership could (probably will) change.

text/0050-fellowship-salaries.md
joepetrowski: I'm hesitant to make it more technical because I think statistics in this case are a facade for the actual numbers. There are just so many data sources on salaries that you could always cherry pick data or aggregation methods that support the number you want. Then it appears technical because you have lots of math/data and "precise" numbers, but it's really not.

For example, I was not able to find OECD sourced data on _engineering_ salaries, just on general averages. Even for engineering specific data across large geopolitical areas (America, Europe), it varies widely based on field and location. I found a few data sources that indicated US engineering salaries were around 100k on average and in the 80-160k range generally.

I would be in favor of altering the structure such that the levels are all relative to each other. E.g., everything is relative to III Dan (so I Dan is 12.5%, II Dan is 25%, etc.), and then make a proposal for a single number: the III Dan level, which people agree represents a reasonable salary for an IC.

xlc: Makes everything relative to the III Dan level will be good. It simplifies future work when we need to adjust the amount. It is easier to propose say change the base to a new number, than try to figure out the bunch numbers for every rank.

sourabhniyogi:
I believe most, if not all, Polkadot Fellows level 3+ are OGs who hold significant amounts of DOT and salaries are largely ceremonial, like Musk/Ellison/Jobs/Bezos/... saying "see, I only earn $1 in salary!" and pretty much couldn't care less about the salary because staking rewards + DOT value itself dominates their actual compensation.  So the following comments are geared for two purposes:
 (a) addressing new fellows, who are not in the OG category, and thus must take this "living wage" seriously
 (b) a desire to use the Salaries pallet for new collectives on the Collectives chain, and the "Dan 3" value as the "standard"

1.  Polkadot fellows (at least everyone I have seen work at 2 or 3 or higher) are SUPER engineers, far better than the average CS grad from top 10 schools in the US.    I think the $80K/yr for Level 3 would be unheard of for super engineers.   I doubt anyone is doing great work because they are getting a "living wage", they are doing great work because they love doing great work on Polkadot, but you need to increase your Rank 3 to be like a Stanford Master's degree CS grad, which is in the $160K range.     For anyone not in the OG club, this would support getting medical insurance for like a spouse and their family, which is par for the course, I think.        

2. The multiplier mechanism on a Dan 3 baseline is a great idea.  We can use this in like dozens (up to 40!) collectives -- and I presume be able to reach into the Fellowship and fetch that baseline in USDT and not have to make independent salary decisions.  This way, when Dan 3 goes up due to inflation (its a thing, in the US), other collectives can rise in direct tandem programmatically.   Is this possible?
  
3. I doubt there is anything special about there being 9 ranks instead of 90 or 100?   If you increase the number of ranks by a factor of 10 or 11, you can promote people 10 ranks instead of 1 ... and not have these unnecessarily massive jumps of $40K/yr in salaries.   This can be especially valuable for the spread between  rank 2 and 3.. which is a jump of $60K, which is just... too much of a jump!    No HR person would think these jumps as reasonable.   

4. Receiving a % of your compensation in DOT instead of 100% in USDT or 100% in DOT should be part of this Salaries pallet, in the same way that receiving a portion of your compensation in restricted stock or stock options is common to most tech companies.     Defining the Dan 3 baseline in USDT and then getting a "salary" in both USDT _and_ DOT determined by the recipient's chosen percentage would be usable by collectives that have members "Salary" receiving  a % of Dan 3 in USDT _and_ DOT.    All collectives in the Collectives Chain would then manage the USDT+DOT with the Salary pallet.  This is not about Fellowship salaries alone.

5. Being able to take a few months off for parental leave, sick leave, sabbatical, health reasons, etc.  should be a key feature of the Salaries pallet, and having something  proportional to eras seems highly appropriate.   I think you're covering this, but the resolution of "do salaries get paid on 28 day eras" would be highly desirable, not just for fellows in this collective, but for other collectives that also should use the same Salaries pallet.    For larger collectives with people entering in permissionlessly we need this granularity?

joepetrowski:
1. It's impossible to make one number that applies to every geopolitical location. A good salary for an engineer out of university in Europe (even in cities like Paris) is 50-60k EUR. I know in the prelude you said you are approximating a Silicon Valley HR person, and in that context, you're right. With this proposal, you'd have to be a very high rank to live in the Valley, or be willing to relocate if "living off the Fellowship salary" is your aim.

2. That would be possible, yes. Just so people know, there's nothing special about having 10 ranks (0 - 9), it's just what was chosen for the Tech Fellowship. You can make a group with 3 or 9 or 100 ranks.

3. Yeah exactly what I was writing in (2). Jumps in salary like that are not normal for year-over-year inflation adjustments, but they are normal when getting significantly promoted. Although the manifesto declares minimum years of service for each rank, in reality it probably takes more than one year to advance one rank; a lot of the requirements are quite difficult and would require consistent performance and contributions at a very high level.

4. Yes I agree about having a percentage in DOT be possible. Right now the Salary pallet makes you specify one asset, but we could change it to accept multiple. Also USDT is not the right asset for every group. For example, with Ambassador Program, DOT makes sense (at least at lower levels) because we want them to have some DOT for demos like "how to {stake, vote}". (Sidenote: Let's save this argument for an Ambassador Program proposal, just an example.) But when it comes to implementing these things, priorities/resources/etc. Should I ask someone to stop working on CCTP compatibility or Coretime to support USDT+DOT salaries (when we can just take salary in a single asset)? IMO, no.

5. It's all configurable. Collectives should be careful about permissionless entry and salaries though. Collectives are supposed to largely self manage, and if senior members don't pay attention, they will allow abuse. If senior members are paying attention, then dealing with things like a vacation seem very doable.

xlc:
I don't want to scope creep this RFC so https://github.com/polkadot-fellows/RFCs/issues/55 for further discussion after this is accepted.

xlc:
Another related issue, would there be retrospective salary payment? If there is none, I will simply apply for separate treasury proposal.

joepetrowski:
> Another related issue, would there be retrospective salary payment? If there is none, I will simply apply for separate treasury proposal.

No, there is no retrospective payment.

ggwpez (approved):

joepetrowski:
/rfc process 0x92681c170876480c8f6e840f7845083fd899cdfedd1f556ffb1aa9ab38891005

gavofyork:
> A III Dan is someone whose contributions match the expectations of a full-time individual contributor.

Note that this is false. The manifesto defines the expectations at each grade. Even at I Dan, there is an expectation for full-time contributions.

shawntabrizi:
> > A III Dan is someone whose contributions match the expectations of a full-time individual contributor.
> 
> Note that this is false. The manifesto defines the expectations at each grade. Even at I Dan, there is an expectation for full-time contributions.

But clearly a Dan I salary is an insufficient living wage, so the expectations are that fellowship members are allowed to facilitate their salary with treasury proposals or companies within the ecosystem?

xlc:
I just found this in manifesto today under section 8 Allowances

> These are a monthly affordance, paid in DOT. There are two levels: standard and passive. **Standard should be between the 80th-90th percentile of gross income in the OECD group of countries.** Passive is no greater than 50% of standard. The total amount given as passive allowance should be no greater than 10% of the total amount given as standard allowance.

So I guess that means Dan III should be 80% percentile and Dan VI should be 90%?

gavofyork:
Yes, or the old fashioned way of blood, sweat and tears 😇

gavofyork:
> I just found this in manifesto today under section 8 Allowances
> 
> > These are a monthly affordance, paid in DOT. There are two levels: standard and passive. **Standard should be between the 80th-90th percentile of gross income in the OECD group of countries.** Passive is no greater than 50% of standard. The total amount given as passive allowance should be no greater than 10% of the total amount given as standard allowance.
> 
> So I guess that means Dan III should be 80% percentile and Dan VI should be 90%?

Broadly speaking, yes. However, it turns out those numbers are rather difficult to find so the actual numbers are something of an estimate. Furthermore, unless there is outside funding, the Fellowship is limited to whatever is acceptable to the Treasury.

PR #56, opened 2023-11-30
Enforce one transaction per notification in the networking

tomaka:
[Rendered](https://github.com/tomaka/RFCs-1/blob/one-tx-per-notif/text/0056-one-transaction-per-notification.md)

This RFC is pretty small, and I expect it to be pretty uncontroversial. Just because something is currently designed a certain way doesn't mean that it is so for a good reason, and sometimes, like in this RFC, things just need to be cleaned up.

### DIFF
--- a/text/0000-one-transaction-per-notification.md
+++ b/text/0000-one-transaction-per-notification.md
@@ -1,0 +1,73 @@
+# RFC-0000: Enforce only one transaction per notification
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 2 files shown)


tomaka:
I've opened it for voting: https://collectives.polkassembly.io/referenda/55?network=collectives

tomaka:
/rfc process 0x9089a6b054ba06cadfba032dddcce2a0e78dbb010adee213d6a857fc81f082c5

tomaka:
Let's just consider this PR merged.

Issue #53, opened 2023-11-23
RFCs Web Page [mdbook]


mordamax:
@PierreBesson suggested rendering our polkadot-fellows/RFCs as a webpage - similar to https://eips.ethereum.org/

- Could be made by markdown book https://rust-lang.github.io/mdBook/index.html
- Can either integrated into RFC's repo or in https://github.com/paritytech/rfc-book
- Do we want to put this page behind subdomain, like https://rfc.polkadot.network? 

Tracked in https://github.com/paritytech/eng-automation/issues/1

rzadp:
I have created the mdbook and I'm looking for feedback.

The book can be previewed [here](https://paritytech.github.io/RFCs/).

Calling thumb-uppers: @xlc @lrazovic @davxy @bkchr

xlc:
I can't say I enjoy the font and colour scheme but otherwise looks good
<img width="800" alt="Screenshot 2023-12-04 at 1 08 58 PM" src="https://github.com/polkadot-fellows/RFCs/assets/1201310/3edf33a9-1922-4313-b175-961ed23b7cad">

rzadp:
Thanks for the feedback. :)

## Update

- I have toned down the Polkadot branded theme.
- I'm generating TOCs.
- I've separated proposed RFCs into "new", "stale", and the rest
  - Currently "new" means PRs created in the last 7 days, and "stale" means PRs not updated for at least 30 days.

Please take another [look](https://paritytech.github.io/RFCs/).

PR #60, opened 2023-12-19
Introduce RFC mdBook web page

rzadp:
- Closes https://github.com/polkadot-fellows/RFCs/issues/53
- Preview available [here](https://paritytech.github.io/RFCs/).

Screenshot 2023-12-19 at 13 55 05

---

### What it does

- The workflow gathers the source markdown files of RFCs - approved ones, and proposed ones (from open PRs).
- The proposed RFCs are separated into new (<7 days since PR created) and stale (>30 days since PR updated).
- The RFCs are extended with a table of contents, and a link to the source.
- The RFCs are build into a web page using [mdBook](https://github.com/rust-lang/mdBook).
- The built web page gets pushed to `gh-pages` branch - to be published in Github Pages.
- The workflow is triggered on merge to `main`, and periodically once a day.


shawntabrizi (approved):
lgtm, im sure there will be future iteration :)

bkchr (approved):

ggwpez:
Is this the final URL https://paritytech.github.io/RFCs/ ?  
Then we should probably mention it in the README or even link it in the GitHub repo.

rzadp:
The final URL is: https://polkadot-fellows.github.io/RFCs/

The paritytech one has served as a preview, before it was configured in polkadot-fellows.

rzadp:
Hey @wirednkod, I didn't know about this dashboard 👍 

> merge the 2 and build the "cargo book" and then release it with the site

Do you have an idea on how we could do it in details?
I don't think the output of `mdBook` is gonna work properly inside a Vite SPA application, even forgetting about the design incosistency.

I see that the dashboard is fetching the PRs dynamically from Github. What if it fetched the RFC data from the mdBook Github branch instead? The branch is `gh-pages`, which gets published to Github Pages. I'm thinking I could add a json file with the data, so it would be a single fetch.

That way, the dashboard would operate on the same raw data, but present it with its own front-end design.
What do you think about this approach?

Unless we want to simply go with a link and keep only one site with full details - that seems reasonable too.

wirednkod:
@rzadp mdBook is nothing more than generated `html` pages. If an iFrame (*yuk*) is added inside the fellowship site, then this would work. The point here is that the data on the mdbook are really detailed and nice - and I think, its better to add a link (to a new tab) towards the RFCs page that you creaed, instead of re-inventing the wheel. 

This way the `mdbook` would be part of the main page and still keep its own design and the work you have done.

I would like to suggest some changes to what you have if we agreed to go with the aforrementioned direction:
- remove introduction page
- Move `Newly Proposed` and `Proposed` categories above the `Approved` ones

I can add those as issues at [https://github.com/polkadot-fellows/RFCs/](https://github.com/polkadot-fellows/RFCs/)

rzadp:
Sounds good, I agree with you.

> Move Newly Proposed and Proposed categories above the Approved ones

Makes total sense, no need to create an issue, I will do it right away.

> remove introduction page

It doesn't seem right for me, for a user who stumbles upon this link to not have any kind of description of what he's looking at. Are you sure about it? If the problem is that it's too short (a single sentence at the moment), we could add more to it.

wirednkod:
@rzadp I have integrated now the RFC-book inside the dashboard (see sceenshot):
image

PR #73, opened 2024-02-12
Referendum Deposit Track

JelliedOwl:
RFC for adding a new track which can place decision deposits on other tracks.  I'm happy to work on implementation if there's a reasonable prospect of this being accepted.

### DIFF
--- a/text/0073-referedum-deposit-track.md
+++ b/text/0073-referedum-deposit-track.md
@@ -1,0 +1,70 @@
+# RFC-0073: Decision Deposit Referendum Track
+
+|                 |                                                                                |
+| --------------- |--------------------------------------------------------------------------------|
+| **Start Date**  | 12 February 2024                                                               |
+| **Description** | Add a referendum track which can place the decision deposit on any other track |
+| **Authors**     | JelliedOwl                                                                     |
+
+## Summary
+
+The current size of the decision deposit on some tracks is too high for many proposers. As a result, those needing to use it have to find someone else willing to put up the deposit for them - and a number of legitimate attempts to use the root track have timed out. This track would provide a more affordable (though slower) route for these holders to use the root track.
+
+## Motivation
+
+There have been recent attempts to use the Kusama root track which have timed out with no decision deposit placed. Usually, these referenda have been related to parachain registration related issues. 
+
+## Explanation
+
+Propose to address this by adding a new referendum track ***[22] Referendum Deposit*** which can place the decision deposit on another referendum. This would require the following changes:
+- [Referenda Pallet] Modify the `placeDecisionDesposit` function to additionally allow it to be called by root, with root call bypassing the requirements for a deposit payment.
+- [Runtime] Add a new referendum track which can only call `referenda->placeDecisionDeposit` and the utility functions.
+
+### Referendum track parameters - Polkadot
+
+- **Decision deposit**: 1000 DOT
+- **Decision period**: 14 days
+- **Confirmation period**: 12 hours
+- **Enactment period**: 2 hour
+- **Approval & Support curves**: As per the root track, timed to match the decision period
+- **Maximum deciding**: 10
+
+### Referendum track parameters - Kusama
+
+- **Decision deposit**: 33.333333 KSM
+- **Decision period**: 7 days
+- **Confirmation period**: 6 hours
+- **Enactment period**: 1 hour
+- **Approval & Support curves**: As per the root track, timed to match the decision period
+- **Maximum deciding**: 10
+
+## Drawbacks
+
+This track would provide a route to starting a root referendum with a much-reduced slashable deposit. This might be undesirable but, assuming the decision deposit cost for this track is still high enough, slashing would still act as a disincentive.
+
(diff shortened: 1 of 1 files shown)


text/0073-referedum-deposit-track.md
```diff
@@ -0,0 +1,70 @@
...
+
+|                 |                                                                                |
+| --------------- |--------------------------------------------------------------------------------|
+| **Start Date**  | 12 February 2024                                                               |
+| **Description** | Add a referendum track which can place the decision deposit on any other track |
+| **Authors**     | JelliedOwl                                                                     |
+
+## Summary
+
+The current size of the decision deposit on some tracks is too high for many proposers. As a result, those needing to use it have to find someone else willing to put up the deposit for them - and a number of legitimate attempts to use the root track have timed out. This track would provide a more affordable (though slower) route for these holders to use the root track.
```
ggwpez, line 11: >The current size of the decision deposit on some tracks is too high for many proposers.

I think in all cases these can be proposed on the whitelisted caller track instead and be fast-tracked by the fellowship.

text/0073-referedum-deposit-track.md
```diff
@@ -0,0 +1,70 @@
...
+- **Confirmation period**: 6 hours
+- **Enactment period**: 1 hour
+- **Approval & Support curves**: As per the root track, timed to match the decision period
+- **Maximum deciding**: 10
+
+## Drawbacks
+
+This track would provide a route to starting a root referendum with a much-reduced slashable deposit. This might be undesirable but, assuming the decision deposit cost for this track is still high enough, slashing would still act as a disincentive.
+
+An alternative to this might be to reduce the decision deposit size some of the more expensive tracks. However, part of the purpose of the high deposit - at least on the root track - is to prevent spamming the limited queue with junk referenda.
```
ggwpez, line 45: > However, part of the purpose of the high deposit - at least on the root track - is to prevent spamming the limited queue with junk referenda.

Yes and to scare away bad proposals since they can be slashed and having a big deposit slashed is bad.

text/0073-referedum-deposit-track.md
```diff
@@ -0,0 +1,70 @@
...
+
+The current size of the decision deposit on some tracks is too high for many proposers. As a result, those needing to use it have to find someone else willing to put up the deposit for them - and a number of legitimate attempts to use the root track have timed out. This track would provide a more affordable (though slower) route for these holders to use the root track.
+
+## Motivation
+
+There have been recent attempts to use the Kusama root track which have timed out with no decision deposit placed. Usually, these referenda have been related to parachain registration related issues. 
+
+## Explanation
+
+Propose to address this by adding a new referendum track ***[22] Referendum Deposit*** which can place the decision deposit on another referendum. This would require the following changes:
```
ggwpez, line 19: What you are asking in fact is to just reduce the deposit to 1000 DOT.  
The fact that this is done through a new track does not matter i think since the Treasury does not have a counterparty risk here from being slashed.  
What i mean: if 100,000 DOT from the treasury is slashed, then nobody cares.

text/0073-referedum-deposit-track.md
JelliedOwl: However, whitelisted caller has a minimum of 5% final support level compared to 0% on other tracks.  So you need at least a lot of abstentions (I think those count against the support threshold).

text/0073-referedum-deposit-track.md
JelliedOwl: It's not quite the same since the number of deciding referenda on the root track is small (I think), so I definitely wouldn't be advocating for reducing the DD there without also increasing the number which could in deciding at any one time.

text/0073-referedum-deposit-track.md
joepetrowski: I don't think everything that requires Root will go through Fellowship though, for example [457](https://polkadot.polkassembly.io/referenda/457). It requires Root but it's not a "pure, neutral" call.

text/0073-referedum-deposit-track.md
adamsteeber: > The fact that this is done through a new track does not matter

I must respectfully disagree. Placing a decision deposit directly on a referendum is *permissionless* whereas this RFC proposes a *permissioned* way to place a decision deposit via OpenGov.

A referendum that seeks to place a decision deposit is fundamentally different from a referendum that seeks to execute an action. Voting in favor of the former signals, "I want this referendum to be decided on by token holders without the disincentive of slashing" whereas voting in favor of the latter signals, "I want this action to be executed on chain." These are two very different statements. If a majority of token holders agree on the former, why shouldn't that be allowed? Isn't this supposed to be *Open*Gov?

The current status quo implies that no referendum should be decided on unless the decision deposit comes from some account permissionlessly. I think that is quite inflexible, especially since there are well-intentioned individuals that should be given recourse, other than the Fellowship, for placing large decision deposits.

ggwpez: Maybe this is the wrong place to decide it - the Fellowship should be consulted for technical questions - not political ones. I suggest bringing it up here ASAP: https://github.com/polkadot-fellows/runtimes/pull/184

PR #74, opened 2024-02-15
Add Enhanced Multisig Pallet

asoliman92:
This RFC proposes adding an enhanced Multisig pallet for both better footprint on the blockchain and better ergonomics.

### DIFF
--- a/text/0057-stateful-multisig-pallet.md
+++ b/text/0057-stateful-multisig-pallet.md
@@ -179,11 +179,11 @@ We always use latest threshold and don't store each proposal with different thre
 
 ## Drawbacks
 
-Description of recognized drawbacks to the approach given in the RFC. Non-exhaustively, drawbacks relating to performance, ergonomics, user experience, security, or privacy.
+* New pallet to maintain.
 
 ## Testing, Security, and Privacy
 
-Describe the the impact of the proposal on these three high-importance areas - how implementations can be tested for adherence, effects that the proposal has on security and privacy per-se, as well as any possible implementation pitfalls which should be clearly avoided.
+Standard audit/review requirements apply.
 
 ## Performance, Ergonomics, and Compatibility
 
@@ -191,24 +191,96 @@ Describe the impact of the proposal on the exposed functionality of Polkadot.
 
 ### Performance
 
-Is this an optimization or a necessary pessimization? What steps have been taken to minimize additional overhead?
+Doing back of the envelop calculation to proof that the stateful multisig is more efficient than the stateless multisig given it's smaller footprint size on blocks.
+
+Quick review over the extrinsics for both as it affects the block size:
+
+Stateless Multisig:
+Both `as_multi` and `approve_as_multi` has a similar parameters:
+
+```rust
+origin: OriginFor<T>,
+threshold: u16,
+other_signatories: Vec<T::AccountId>,
+maybe_timepoint: Option<Timepoint<BlockNumberFor<T>>>,
+call_hash: [u8; 32],
+max_weight: Weight,
+```
+
+Stateful Multisig:
+We have the following extrinsics:
+
+```rust
+pub fn start_proposal(
+			origin: OriginFor<T>,
+			multisig_account: T::AccountId,
+			call_hash: T::Hash,
+		)
+```
+
+```rust
+pub fn approve(
+			origin: OriginFor<T>,
+			multisig_account: T::AccountId,
+			call_hash: T::Hash,
+		)
+```
+
+```rust
+pub fn execute_proposal(
+			origin: OriginFor<T>,
+			multisig_account: T::AccountId,
+			call: Box<<T as Config>::RuntimeCall>,
+		)
+```
+
+The main takeway is that we don't need to pass the threshold and other signatories in the extrinsics. This is because we already have the threshold and signatories in the state (only once).
+
+So now for the caclulation
+
+Given the following:
+
+* K is the number of multisig accounts.
+* N is number of owners in each multisig account.
+* For each proposal we need to have 2N/3 approvals.
+
+The table calculates if everyday each of the K multisig accounts has one proposal and it gets approved by the 2N/3 and then executed. How much did the total footprint on Blocks and States increased by the end of the day.
+
+Note: We're not calculating the cost of proposal as both in statefull and stateless multisig they're almost the same and gets cleaned up from the state once the proposal is executed or canceled.
+
+Stateless effect on blocksizes = 2/3*K*N^2 (as each user of the 2/3 users will need to call approve_as_multi with all the other signatories(N) in extrinsic body)
+
+Stateful effect on blocksizes = K * N (as each user will need to call approve with the multisig account only in extrinsic body)
+
+Stateless effect on statesizes = Nil (as the multisig account is not stored in the state)
+
+Stateful effect on statesizes = K*N (as each multisig account (K) will be stored with all the owners (K) in the state)
+
+| Pallet         |  Block Size   | State Size |
+|----------------|:-------------:|-----------:|
+| Stateless      |     2/3*K*N^2 |        Nil |
+| Stateful       |          K*N  |       K*N  |
 
 ### Ergonomics
 
-If the proposal alters exposed interfaces to developers or end-users, which types of usage patterns have been optimized for?
+The Stateful Multisig will have better ergonomics for managing multisig accounts for both developers and end-users.
 
 ### Compatibility
 
-Does this proposal break compatibility with existing interfaces, older versions of implementations? Summarize necessary migrations or upgrade strategies, if any.
+This RFC is compatible with the existing implementation and can be handled via upgrades and migration. It's not intended to replace the existing multisig pallet.
 
 ## Prior Art and References
 
-Provide references to either prior art or other relevant research for the submitted design.
+[multisig pallet in polkadot-sdk](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/multisig)
 
 ## Unresolved Questions
 
-Provide specific questions to discuss and address before the RFC is voted on by the Fellowship. This should include, for example, alternatives to aspects of the proposed design where the appropriate trade-off to make is unclear.
+* On account deletion, should we transfer remaining deposits to treasury or remove account creation deposits completely and consider it as fees to start with?
 
 ## Future Directions and Related Material
 
-Describe future work which could be enabled by this RFC, if it were accepted, as well as related RFCs. This is a place to brain-dump and explore possibilities, which themselves may become their own RFCs.
+* [ ] Batch proposals. The ability to batch multiple calls into one proposal.  
+* [ ] Batch addition/removal of owners.
+* [ ] Add expiry to proposals. After a certain time, proposals will not accept any more approvals or executions and will be deleted.  
+* [ ] Add extra identifier other than call_hash to proposals (e.g. nonce). This will allow same call to be proposed multiple times and be in pending state.  
+* [ ] Implement call filters. This will allow multisig accounts to only accept certain calls.
--- a/text/0057-stateful-multisig-pallet.md
+++ b/text/0057-stateful-multisig-pallet.md
@@ -26,13 +26,13 @@ Current native solutions for multisig operations are not efficient and lack func
 
 #### Stateless Multisig
 
-We refer to current [multisig pallet in polkadot-sdk](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/multisig) because the multisig account is only created in the state when a proposal is made and is deleted after the proposal is executed or canceled. Although deriving the account is determinsitc it relies on exact users and thershold to derive it. This does not allow for control over the multisig account. It's also tightly coupled to exact users and threshold. This makes it hard for an organization to manage existing accounts and to change the threshold or add/remove owners.
+We refer to current [multisig pallet in polkadot-sdk](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/multisig) because the multisig account is only derived and not stored in the state. Although deriving the account is determinsitc as it relies on exact users (sorted) and thershold to derive it. This does not allow for control over the multisig account. It's also tightly coupled to exact users and threshold. This makes it hard for an organization to manage existing accounts and to change the threshold or add/remove owners.
 
 We believe as well that the stateless multisig is not efficient in terms of block footprint as we'll show in the performance section.
 
 #### Pure Proxy
 
-Pure proxy can acheive having the determinstic multisig account from different users but it's unneeded complexity as a way around the limitations of the current multisig pallet.
+Pure proxy can achieve having a stored and determinstic multisig account from different users but it's unneeded complexity as a way around the limitations of the current multisig pallet. It doesn't also have the same fine grained control over the multisig account.
 
 ### Requirements
 
@@ -57,6 +57,8 @@ Multisig accounts can be used for joint accounts where multiple individuals need
 * Decentralized Autonomous Organizations (DAOs):
 DAOs can utilize multisig accounts to ensure that decisions are made collectively. Multiple key holders can be required to approve changes to the organization's rules or the allocation of funds.
 
+and much more...
+
 ## Stakeholders
 
 * Polkadot holders
@@ -70,7 +72,6 @@ Let's start with a sequence diagram to illustrate the main operations of the Sta
 
 ![multisig operations](https://github.com/asoliman92/RFCs/assets/2677789/4f2e8972-f3b8-4250-b75f-1e4788b35752)
 
-
 Notes on above diagram:
 
 * It's a 3 step process to execute a proposal. (Start Proposal --> Approvals --> Execute Proposal)
@@ -80,11 +81,9 @@ Notes on above diagram:
 * Any multisig account owner can start a proposal on behalf of the multisig account. (Bob in the diagram)
 * Any multisig account owener can execute proposal if it's approved by enough owners. (Dave in the diagram)
 
-Detail-heavy explanation of the RFC, suitable for explanation to an implementer of the changeset. This should address corner cases in detail and provide justification behind decisions, and provide rationale for how the design meets the solution requirements.
-
 ### State Transition Functions
 
-All functions have rustdoc in the code. Here is a brief overview of the functions:
+All functions have detailed rustdoc in [PR#3300](https://github.com/paritytech/polkadot-sdk/pull/3300). Here is a brief overview of the functions:
 
 * `create_multisig` - Create a multisig account with a given threshold and initial owners. (Needs Deposit)
 * `start_proposal` - Start a multisig proposal. (Needs Deposit)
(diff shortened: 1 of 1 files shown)


text/0074-stateful-multisig-pallet.md
```diff
@@ -0,0 +1,293 @@
...
+* It's a 3 step process to execute a proposal. (Start Proposal --> Approvals --> Execute Proposal)
+* `Execute` is an explicit extrinsic for a simpler API. It can be optimized to be executed automatically after getting enough approvals.
+* Any user can create a multisig account and they don't need to be part of it. (Alice in the diagram)
+* A proposal is any extrinsic including control extrinsics (e.g. add/remove owner, change threshold,..etc).
+* Any multisig account owner can start a proposal on behalf of the multisig account. (Bob in the diagram)
+* Any multisig account owener can execute proposal if it's approved by enough owners. (Dave in the diagram)
+
+### State Transition Functions
+
+All functions have detailed rustdoc in [PR#3300](https://github.com/paritytech/polkadot-sdk/pull/3300). Here is a brief overview of the functions:
```
xlc, line 86: the RPC is the spec, not the code, so we should have all the technical details here
in this case, we need arguments and detailed descriptions such as origin requirement, events, deposits etc

text/0074-stateful-multisig-pallet.md
```diff
@@ -0,0 +1,293 @@
...
+* `create_multisig` - Create a multisig account with a given threshold and initial owners. (Needs Deposit)
+* `start_proposal` - Start a multisig proposal. (Needs Deposit)
+* `approve` - Approve a multisig proposal.
+* `revoke` - Revoke a multisig approval from an existing proposal.
+* `execute_proposal` - Execute a multisig proposal. (Releases Deposit)
+* `cancel_own_proposal` - Cancel a multisig proposal started by the caller in case no other owners approved it yet. (Releases Deposit)
+
+Note: Next functions need to be called from the multisig account itself. Deposits are reserved from the multisig account as well.
+
+* `add_owner` - Add a new owner to a multisig account. (Needs Deposit)
```
xlc, line 97: I prefer `signer` over `owner`

text/0074-stateful-multisig-pallet.md
```diff
@@ -0,0 +1,293 @@
...
+As for the values:
+
+```rust
+pub struct MultisigAccountDetails<T: Config> {
+	/// The owners of the multisig account. This is a BoundedBTreeSet to ensure faster operations (add, remove).
+	/// As well as lookups and faster set operations to ensure approvers is always a subset from owners. (e.g. in case of removal of an owner during an active proposal)
+	pub owners: BoundedBTreeSet<T::AccountId, T::MaxSignatories>,
+	/// The threshold of approvers required for the multisig account to be able to execute a call.
+	pub threshold: u32,
+	pub creator: T::AccountId,
```
xlc, line 133: do we really need this? looks like it is for refund deposit when the multisig is destroyed. the easy solution is to refund it to the signer that triggers the removal

text/0074-stateful-multisig-pallet.md
```diff
@@ -0,0 +1,293 @@
...
+pub struct MultisigProposal<T: Config> {
+    /// Proposal creator.
+    pub creator: T::AccountId,
+    pub creation_deposit: BalanceOf<T>,
+    /// The extrinsic when the multisig operation was opened.
+    pub when: Timepoint<BlockNumberFor<T>>,
+    /// The approvers achieved so far, including the depositor.
+    /// The approvers are stored in a BoundedBTreeSet to ensure faster lookup and operations (approve, revoke).
+    /// It's also bounded to ensure that the size don't go over the required limit by the Runtime.
+    pub approvers: BoundedBTreeSet<T::AccountId, T::MaxSignatories>,
```
xlc, line 148: do we not need to store rejectors? e.g. for 2/4 multisig, the proposal can be removed if 3 signers rejects it

xlc (commented):
I don't see a section explain why to add this pallet to collectives chain (instead of other chains

text/0074-stateful-multisig-pallet.md
asoliman92: Thanks! I'm adding them now 👍

text/0074-stateful-multisig-pallet.md
asoliman92: I like `signer` more as well :).

text/0074-stateful-multisig-pallet.md
asoliman92: Whoever created the multisig might be not in the signers list and even if they're one of the signers, anyone from the signers can dispatch/cancel the final call. That's why it's better to return it to the original creator.

text/0074-stateful-multisig-pallet.md
asoliman92: The initial design had `cancel_proposal` as a separate multisig operation. Meaning that to cancel the proposal they'd need another proposal to do so. This is not optimal and I agree with you suggestion. After reaching rejections threshold, any signer can call `cancel_proposal` like `execute_proposal`. Working on updating the RFC. Thanks!

asoliman92:
I actually had it as an open question to begin with and then moved it to subject as a suggestion. I did it by process of elimination of other system chains. But now that you mention it, I notice that the original multisig is available on all system chains. What are your thoughts?

text/0074-stateful-multisig-pallet.md
```diff
@@ -0,0 +1,585 @@
...
+		///
+		/// * `multisig_account` - The multisig account ID.
+		/// * `call` - The dispatchable call to be executed.
+		///
+		/// # Errors
+		///
+		/// * `MultisigNotFound` - The multisig account does not exist.
+		/// * `UnAuthorizedSigner` - The caller is not an signer of the multisig account.
+		/// * `TooManySignatories` - The number of signatories exceeds the maximum allowed. (shouldn't really happen as it's the first approval)
+		pub fn start_proposal(
```
Tbaut, line 131: A major hurdle for users, is that they need some offline communication to pass around the `call data`, so that they can verify that  the hash is actually matching, and for the last signer to execute the call.
 
While this is great for privacy, as in it's hard to guess as an external watcher what the call is about by knowing only the hash. Experience has shown that most users don't need this additional privacy, the reality is that they don't verify the `call data` and blindly approve multisig calls, trusting the original proposer. As a result, Multix and many other multisig management platforms use `multisig.asMulti` for the proposer call, hence exposing the call in the block and relying on an indexer to later decode the call, and show users what they are signing. 

If we add such a pallet to a system chain, I'd strongly advocate to include a mechanism to optionally store the call in the state, for `start_proposal` and then have it optional as well for the `execute_proposal` and retrieve it from the state is possible.

asoliman92: I don't really have strong opinion here. I initially had it as a `RuntimeCall` and changed it in favor of the hash for both privacy and block footprint. I was thinking using preimage as well but maybe that's just complicating it.

> the reality is that they don't verify the call data and blindly approve multisig calls, trusting the original proposer.

Don't you think we can offload this part to the platforms instead to get the best of both worlds? Multix can get the call and hash it with the correct hash before calling the extrinsic. That will ensure seamless and private experience for correct users without sacrificing block footprint. It might not be the best for people who want to use it directly from polkadot.js for example but still good enough for most people who'll use some tool. What are your thoughts?

text/0074-stateful-multisig-pallet.md
```diff
@@ -0,0 +1,585 @@
...
+
+The main takeway is that we don't need to pass the threshold and other signatories in the extrinsics. This is because we already have the threshold and signatories in the state (only once).
+
+So now for the caclulations, given the following:
+
+* K is the number of multisig accounts.
+* N is number of signers in each multisig account.
+* For each proposal we need to have 2N/3 approvals.
+
+The table calculates if each of the K multisig accounts has one proposal and it gets approved by the 2N/3 and then executed. How much did the total Blocks and States sizes increased by the end of the day.
```
Ank4n, line 538: Would also be great to see more realistic storage benchmark of both implementations of multisig pallet.

text/0074-stateful-multisig-pallet.md
```diff
@@ -0,0 +1,585 @@
...
+| Pallet         |  Block Size   | State Size |
+|----------------|:-------------:|-----------:|
+| Stateless      |           N^2 |        Nil |
+| Stateful       |           N   |         N  |
+
+So even though the stateful multisig has a larger state size, it's still more efficient in terms of block size and total footprint on the blockchain.
+
+### Ergonomics
+
+The Stateful Multisig will have better ergonomics for managing multisig accounts for both developers and end-users.
```
Ank4n, line 565: Would be great addition to have a better functional comparison to the stateless pallet instead of a generic statement. Are there use cases where you believe stateless is better?

text/0074-stateful-multisig-pallet.md
```diff
@@ -0,0 +1,585 @@
...
+
+This RFC is compatible with the existing implementation and can be handled via upgrades and migration. It's not intended to replace the existing multisig pallet.
+
+## Prior Art and References
+
+[multisig pallet in polkadot-sdk](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/multisig)
+
+## Unresolved Questions
+
+* On account deletion, should we transfer remaining deposits to treasury or remove signers' addition deposits completely and consider it as fees to start with?
```
Ank4n, line 577: Then the user has no incentive to delete the account. Having the deposit that should be returned on clearing makes sense to me.

text/0074-stateful-multisig-pallet.md
```diff
@@ -0,0 +1,585 @@
...
+
+#### Stateless Multisig
+
+We refer to current [multisig pallet in polkadot-sdk](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/multisig) because the multisig account is only derived and not stored in the state. Although deriving the account is determinsitc as it relies on exact users (sorted) and thershold to derive it. This does not allow for control over the multisig account. It's also tightly coupled to exact users and threshold. This makes it hard for an organization to manage existing accounts and to change the threshold or add/remove signers.
+
+We believe as well that the stateless multisig is not efficient in terms of block footprint as we'll show in the performance section.
+
+#### Pure Proxy
+
+Pure proxy can achieve having a stored and determinstic multisig account from different users but it's unneeded complexity as a way around the limitations of the current multisig pallet. It doesn't also have the same fine grained control over the multisig account.
```
Ank4n, line 35: As long as the complexity of pure proxy is not leaked to the end user (via a frontend), it should be fine. Stateless implementation should be simpler I am assuming as a counter argument.

text/0074-stateful-multisig-pallet.md
```diff
@@ -0,0 +1,585 @@
...
+
+So even though the stateful multisig has a larger state size, it's still more efficient in terms of block size and total footprint on the blockchain.
+
+### Ergonomics
+
+The Stateful Multisig will have better ergonomics for managing multisig accounts for both developers and end-users.
+
+### Compatibility
+
+This RFC is compatible with the existing implementation and can be handled via upgrades and migration. It's not intended to replace the existing multisig pallet.
```
Ank4n, line 569: What does compatible mean here? Can existing stateless users migrate to stateful without losing the multisig address? How would the process look like?

Also, why (or why not) replace the existing pallet with stateful? Keeping two pallets with same functionality might add more confusion to the end users.

text/0074-stateful-multisig-pallet.md
```diff
@@ -0,0 +1,585 @@
...
+
+[multisig pallet in polkadot-sdk](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/multisig)
+
+## Unresolved Questions
+
+* On account deletion, should we transfer remaining deposits to treasury or remove signers' addition deposits completely and consider it as fees to start with?
+
+## Future Directions and Related Material
+
+* [ ] Batch proposals. The ability to batch multiple calls into one proposal.
```
Ank4n, line 581: Utility pallet can be used to batch a call already.

Ank4n (commented):
afaiu RFC is not about implementation detail (although in some special cases it can be) but high level technical design decisions. I would like to see more examples/evidences of why stateful multisig would be significant improvement over stateless multisig such that it warrants the existing system chains to add a new pallet in its runtime. Introducing this pallet for new runtimes probably needs much less convincing.

text/0074-stateful-multisig-pallet.md
asoliman92: The creator will get back the deposit. This is only for adding new singers to the multisig after it's been created. In this case the `add_signer` is a multisig operation and the deposit is taken from the multisig account itself.

text/0074-stateful-multisig-pallet.md
asoliman92: That means I need to implement it and implement the benchmark and then compare them. I'll do this for sure if I got the green light to proceed with the RFC. As this is an RFC to illustrate the importance and the overall design I think this back of the envelop calculation should be enough unless you see something fundamentally wrong with my calculations. WDYT?

text/0074-stateful-multisig-pallet.md
asoliman92: > Are there use cases where you believe stateless is better?

That's a good question. I don't have any use case in my mind that the stateless is better TBH. Stateful can do all what the stateless do with added functionality, fine tuned control and clearer API.  Do you have a case in mind?

> Would be great addition to have a better functional comparison to the stateless pallet instead of a generic statement.

I thought the RFC itself is an enough comparison to make it clear by reaching this point. Do you suggest explaining it again like in overview and motivation sections or add something else?

text/0074-stateful-multisig-pallet.md
asoliman92: Using pure-proxy still needs stateless multisig pallet as far as I understand. That means even worse block footprint than the quadratic multisig already. [reference](https://www.youtube.com/watch?v=iGRoGstB_pQ&t=92s)

text/0074-stateful-multisig-pallet.md
asoliman92: Compatible in this section means that it's not a breaking change and no other pallets will have issues if this is deployed.

> Can existing stateless users migrate to stateful without losing the multisig address? How would the process look like?

It's possible in theory yes. I'd rather offload this to users though. Once the new pallet is ready they can create a new account and move fund if needed. The incentive is easier/extensible multisig account to deal with.

> Keeping two pallets with same functionality might add more confusion to the end users.

I agree. I think having the stateful will make it easier for users to use after knowing the enhanced functionalities and they'll start using the new one but maybe that's not enough. What do you suggest here? I think first get the sentiment about the stateful multisig, deploy and then we can later see the usage on-chain and decide to migrate and kill the older one or not.

asoliman92:
> afaiu RFC is not about implementation detail (although in some special cases it can be) but high level technical design decisions. I would like to see more examples/evidences of why stateful multisig would be significant improvement over stateless multisig such that it warrants the existing system chains to add a new pallet in its runtime. Introducing this pallet for new runtimes probably needs much less convincing.

So in your opinion the enhanced functionality, ease of use and less footprint (from quadratic to linear) are not enough? Or you think that these evidences are lacking enough backing from the RFC as is?

text/0074-stateful-multisig-pallet.md
asoliman92: @Tbaut may have more insights on this as well. WDYT Tbaut?

Tbaut: What's the question exactly? I don't see the need of pure proxy usage if this pallet is used.

asoliman92: Ankan was asking why use this pallet if we already have pure proxy as a counter argument.

text/0074-stateful-multisig-pallet.md
Tbaut: We're building a decentralized network, yet for average users to be able to use multisigs we have to resort to centralized indexers (in the case of Multix) or IMHO worse, full blown web2 stacks, storing the callData. Letting such important part of the puzzle to centralized entities is certainly not the best of both worlds. If we let users choose, then yes. But today (and in your proposed solution) there's no choice.

asoliman92: I agree 👍. I'm coming from web2 and in general I'm more inclined to make the API clearer and easier to use. I didn't want that to influence me much and trying to embrace optimizing for the blockchain itself as well.  @shawntabrizi  suggested using the hash as an optimization and using preimage pallet when I discussed with him during the PBA earlier. I know it's different schools but I'm interested to hear others thoughts as well.

CC @xlc @Ank4n @bkchr

text/0074-stateful-multisig-pallet.md
Tbaut: There are many answers to this.
- pure proxies aren't (yet) a thing cross chain
- the end user complexity is much much higher with pure proxies, also for new users smart contract multisig are widely known while pure proxies are obscure. 
- you can shoot yourself in the foot by deleting the proxy, and effectively loosing access to funds with pure proxies. 
- the problem that I mentioned in my other comment is fundamental. If ppl can't easily verify, they won't. Hence, having the ability to host the call is paramount. Otherwise you rely on centralized 3rd party, it's not related to pure proxies, but it's a very important addition if this pallet supports it.

text/0074-stateful-multisig-pallet.md
Tbaut: Preimage would work, but in practice it's still cumbersome because only the submitter can delete/free the preimage that was set. This means that the submitter would need an additional call, once the multisig proposal has been executed, to free their preimage. My choice would be to have the same logic as the preimage, and the same deposit, but this preimage would be linked to the multisig proposal, and would be freed when the call is executed.

xlc: We can just have both. Make it something like

```rs
enum CallOrHash {
  Call(Vec<u8>,
  Hash(H256),
}
```

and up to the user to decide how to use it

In fact, this is how the current multisig works

text/0074-stateful-multisig-pallet.md
asoliman92: Is it common practice though for someone to deposit and another to get their deposit back? I understand the incentive part but having someone else take my deposit if I created the account seems a bit unfair. Maybe have part  of the deposit returned to the person executing the deletion process and another part to the original creator?

text/0074-stateful-multisig-pallet.md
asoliman92: I was thinking having the `CallOrHash` will be complicating the solution more when I started working on this. Now I see it might be better to use it with some preimage in the state as well. I will change the RFC accordingly then.

text/0074-stateful-multisig-pallet.md
asoliman92: Thanks @tbaut! I added these points to the RFC under your name as well

text/0074-stateful-multisig-pallet.md
xlc: the original creator have to explicitly transfer the ownership of the multisig so there is nothing unfair. the multisig itself must already worth something more than deposit so the value of deposit won’t change anything

text/0074-stateful-multisig-pallet.md
asoliman92: I've changed the RFC to use `CallOrHash` in all extrinsics.

>  I'd strongly advocate to include a mechanism to optionally store the call in the state, for start_proposal and then have it optional as well for the execute_proposal and retrieve it from the state if possible.

@Tbaut 
I've not written the details of storing the preimage but the `execute_proposal` specs explains it and added a new error if the preimage is not found by the time of execution.

gavofyork:
Some things which should be mentioned in drawbacks:

- adding a new pallet which can control accounts across system chains substantially increases attack surface.
- stateful code is especially hard to analyse fully.
- multisig/pure proxy support is already patchy at best; adding another way of doing the same thing could stoke fragmentation with different UIs supporting different multisig configurations.

PR #76, opened 2024-02-20
Increase the maximum length of identity raw data values from 32 bytes

ltfschoen:


bkchr:
The identity is now generic and configured in the runtime. Generally not sure we need a RFC for such a change. I think it is fine to just do this as a pr and convincing of the maintainers.

CC @joepetrowski

ggwpez:
I think Joe is OOO, but currently we re-use this [Data](https://github.com/paritytech/polkadot-sdk/blob/d1f678c0ec44e9734cf242247dab90c2678774a9/substrate/frame/identity/src/types.rs#L46) type in the new Rococo people config and partly in Polkadot.  
It should be enough to make that 32 there generic as `Get<u32>`.

joepetrowski:
Yeah I agree with @ggwpez and @bkchr , should be a PR to polkadot-sdk to make that value configurable and then we can propose the change in the pallet config.

bkchr:
@ltfschoen are you okay with the proposed steps?

ltfschoen:
> @ltfschoen are you okay with the proposed steps?

yes i agree with @ggwpez, @joepetrowski, and yourself

bkchr:
So this can be closed?

ltfschoen:
Closed and replaced with Polkadot SDK issue https://github.com/paritytech/polkadot-sdk/issues/3977

PR #77, opened 2024-02-20
Increase the maximum length of identity PGP fingerprint values from 20 bytes

ltfschoen:


davxy:
PGP key fingerprints are 20 bytes long.
If you hex encode 20 bytes obviously you get 40 chars (as 1 byte is encoded in 2 chars).

AFAICT the current implementation is correct as it allows you to put there 20 bytes.
In polkadot.js, to insert any data as a raw byte string you need to prefix your pgp fingerprint with `0x` (otherwise is interpreted as a sequence of characters (40 in this case).

Source: I'm using pgp entry for my identity

muelli:
> PGP key fingerprints are 20 bytes long.


note that the upcoming openpgp standard defines "A v6 fingerprint [as] the 256-bit SHA2-256 hash" in section 5.5.4.3. "Version 6 Key ID and Fingerprint" in https://datatracker.ietf.org/doc/draft-ietf-openpgp-crypto-refresh/ which is close to publication.

davxy:
> > PGP key fingerprints are 20 bytes long.
> 
> note that the upcoming openpgp standard defines "A v6 fingerprint [as] the 256-bit SHA2-256 hash" in section 5.5.4.3. "Version 6 Key ID and Fingerprint" in https://datatracker.ietf.org/doc/draft-ietf-openpgp-crypto-refresh/ which is close to publication.

In that case we eventually need to update the max length to 32 bytes. 

BTW, the issue with this RFC is not about the max length of the fingerprint. Here the main issue is the confusion between the length in bytes and the length when the bytes are encoded as a hex string. 

e.g. [here](https://github.com/ltfschoen/RFCs/blob/2968544c6fe488ebf1d434d1edfa266cc530b0b0/text/0077-increase-max-length-of-identity-pgp-fingerprint-value.md?plain=1#L48) and [here](https://github.com/ltfschoen/RFCs/blob/2968544c6fe488ebf1d434d1edfa266cc530b0b0/text/0077-increase-max-length-of-identity-pgp-fingerprint-value.md?plain=1#L27)

PR #78, opened 2024-02-22
Merkleized metadata

bkchr:


xlc:
Maybe we can have a reference test section with some test data? We already have PoC so it shouldn't be hard to make some just by taking from existing tests.

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,344 @@
...
+## Unresolved Questions
+
+None.
+
+## Future Directions and Related Material
+
+- Does it work with all kind of offline wallets?
+- Generic types currently appear multiple times in the metadata with each instantiation. It could be may be useful to have generic type only once in the metadata and declare the generic parameters at their instantiation. 
+- The metadata doesn't contain any kind of semantic information. This means that the offline wallet for example doesn't know what is a balance etc. The current solution for this problem is to match on the type name, but this isn't a sustainable solution.
+- `ExtrinsicMetadata` only provides one `token` and `decimal`. However, chains support a lot of chains support multiple tokens for paying fees etc. Probably more a question of having semantic information as mentioned above.
```
josepot, line 344:
```suggestion
- `MetadataDigest` only provides one `token` and `decimal`. However, chains support a lot of chains support multiple tokens for paying fees etc. Probably more a question of having semantic information as mentioned above.
```

I'm pretty sure that this was a typo.

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,344 @@
...
+
+### Prune unrelated Types
+
+The FRAME metadata contains not only the type information for decoding extrinsics, but it also contains type information about storage types. The scope of the RFC is only about decoding transactions on offline wallets. Thus, a lot of type information can be pruned. To know which type information are required to decode all possible extrinsics, `ExtrinsicMetadata` has been defined. The extrinsic metadata contains all the types that define the layout of an extrinsic. Therefore, all the types that are accessible from the types declared in the extrinsic metadata can be collected. To collect all accessible types, it requires to recursively iterate over all types starting from the types in `ExtrinsicMetadata`. Note that some types are may accessible, but they don't appear in the final type information and thus, can be pruned as well. These are for example inner types of `Compact` or the types referenced by `BitSequence`. The result of collecting these accessible types is a list of all the types that are required to decode each possible extrinsic.
```
josepot, line 220:
```suggestion
The FRAME metadata contains not only the type information for decoding extrinsics, but it also contains type information about storage types. The scope of the RFC is only about decoding transactions on offline wallets. Thus, a lot of type information can be pruned. To know which type information are required to decode all possible extrinsics, `ExtrinsicMetadata` has been defined. The extrinsic metadata contains all the types that define the layout of an extrinsic. Therefore, all the types that are accessible from the types declared in the extrinsic metadata can be collected. To collect all accessible types, it requires to recursively iterate over all types starting from the types in `ExtrinsicMetadata`. Note that some types are accessible, but they don't appear in the final type information and thus, can be pruned as well. These are for example inner types of `Compact` or the types referenced by `BitSequence`. The result of collecting these accessible types is a list of all the types that are required to decode each possible extrinsic.
```

xlc:
A random example: https://eips.ethereum.org/EIPS/eip-145#test-cases

In this case, I am expecting something like

Input: An example metadata JSON
Output: A digest hash

bkchr:
> Input: An example metadata JSON

The metadata is multiple 100KB's big. I don't want to paste this into the document :see_no_evil: I mean I see what you want, but I don't really see on what this brings us here?

josepot:
I have a couple of questions:

1) In the recent metadata versions (at least in versions `14` and `15`), the `BitSequence` definition looks like this:

```rust
struct BitSequence {
    bit_store_type: TypeRef,
    bit_order_type: TypeRef,
}
```

Which I personally find it quite annoying because the `TypeRef` of the `bit_order_type` is always an empty `composite`, and the only way to infer the `least_significant_bit_first` field is to look into the `path` property of said reference, which is quite brittle.

The proposed `BitSequence` definition of this spec is a lot nicer:

```rust
struct BitSequence {
    num_bytes: u8,
    least_significant_bit_first: bool,
}
```

However, unless I'm mistaken, this change won't make it possible to generate the `MetadataDigest` using "old" versions of the metadata like `v14` and `v15`, correct?

2) `MetadataDigest` has references to the fields: `token_symbol` and `decimals`. However, AFAIK those fields are currently not being stored on-chain. So, I'm guessing that this RFC implies that these fields will be stored on-chain moving forward?

If that's the case, wouldn't it be beneficial to also mention in the spec what's the canonical way for accessing those field values?

Furthermore, it would probably be a good idea to state what the canonical way is for accessing the values for the fields `spec_name`, `spec_symbol` and `base58_prefix`, which IIRC they are today present in the constants of the "System" pallet. I'm not sure what the right way to reflect that in the spec is, TBH.

xlc:
No need to put a real metadata. Just a hand crafted minimal one would do. 
The idea is it will be used as a test case that future implementations must ensure the outcome always matches. The same reason other spec/eip have test cases.

bkchr:
There a quite a lot of different types and combinations of them that would be required to make any useful test case. I mean I'm in for ensuring that future implementation are compatible, but we should ensure this by testing these implementation against each.

bkchr:
Current implementation: https://github.com/bkchr/merkleized-metadata

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,390 @@
...
+
+while nodes.len() > 1 {
+    let left = nodes.pop_back();
+    let right = nodes.pop_back();
+
+    nodes.push_front(blake3::hash(SCALE::encode((left, right))));
+}
+
+let merkle_tree_root = if nodes.is_empty() { [0u8; 32] } else { nodes.back() };
+```
```
carlosala, line 308:
```suggestion
```rust
let nodes = leaves;

while nodes.len() > 1 {
    let left = nodes.pop_back();
    let right = nodes.pop_back();

    nodes.push_front(blake3::hash(SCALE::encode((left, right))));
}

let merkle_tree_root = if nodes.is_empty() { [0u8; 32] } else { nodes.back() };
```
```

File: text/0078-merkleized-metadata.md [hunk 4]
```diff
@@ -0,0 +1,390 @@
...
+}
+
+while nodes.len() > 1 {
+    let left = nodes.pop_back();
+    let right = nodes.pop_back();
+
+    nodes.push_front(blake3::hash(SCALE::encode((left, right))));
+}
+
+let merkle_tree_root = if nodes.is_empty() { [0u8; 32] } else { nodes.back() };
```
carlosala, line 307:
```suggestion
let nodes = leaves;

while nodes.len() > 1 {
    let left = nodes.pop_back();
    let right = nodes.pop_back();

    nodes.push_front(blake3::hash(SCALE::encode((left, right))));
}

let merkle_tree_root = if nodes.is_empty() { [0u8; 32] } else { nodes.back() };
```

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,390 @@
...
+
+struct BitSequence {
+    num_bytes: u8,
+    least_significant_bit_first: bool,
+}
+
+struct EnumerationVariant {
+    name: String,
+    fields: Vec<Field>,
+    index: Compact<u32>,
```
carlosala, line 172: This should be a `u8`, right? Or I'm missing something?

text/0078-merkleized-metadata.md
carlosala, line 307:
```suggestion
let nodes = [];

while leaves.len() > 1 {
    let right = leaves.pop_back();
    let left = leaves.pop_back();

    nodes.push_back(blake3::hash(scale::encode((left, right))));
}

if leaves.len() == 1 {
    nodes.push_front(leaves.pop_back());
}

while nodes.len() > 1 {
    let right = nodes.pop_front();
    let left = nodes.pop_front();


    nodes.push_back(blake3::hash(SCALE::encode((left, right))));
}

let merkle_tree_root = if nodes.is_empty() { [0u8; 32] } else { nodes.back() };
```

text/0078-merkleized-metadata.md
bkchr: This was done on purpose. From time to time the idea pops up to change this `index` in an enum to a `Compact`. So, far no one has done this :see_no_evil: 

For most of the enums this will be `u8` any way when using `Compact`.

carlosala: Sounds good👍

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,390 @@
...
+        token_symbol: String,
+    },
+}
+```
+
+The `Hash` is 32 bytes long and `blake3` is used for calculating it. The hash of the `MetadataDigest` is calculated by `blake3(SCALE(MetadataDigest))`. Therefore, `MetadataDigest` is at first `SCALE` encoded, and then those bytes are hashed.
+
+The `MetadataDigest` itself is represented as an `enum`. This is done to make it future proof, because a `SCALE` encoded `enum` is prefixed by the `index` of the variant. This `index` represents the version of the digest. As seen above, there is no `index` zero and it starts directly with one. Version one of the digest contains the following elements:
+
+- `type_information_tree_root`: The root of the [Merkleized type information](#merkleizing-type-information) tree.
```
SBalaguer, line 89: Should the link point to "Building the Merkle Tree Root?"

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,390 @@
...
+}
+
+struct SignedExtensionMetadata {
+    identifier: String,
+    included_in_extrinsic: TypeRef,
+    included_in_signed_data: TypeRef,
+}
+```
+
+To begin with, `TypeRef`. This is a unique identifier for a type as found in the type information. Using this `TypeRef`, it is possible to look up the type in the type information tree. More details on this process can be found in the section [Merkleizing type information](#merkleizing-type-information).
```
SBalaguer, line 117: Same as above about to which section the link refers to.

SBalaguer (commented):
Some text edit suggestions.

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,410 @@
...
+}
+if leaves.len() == 1 {
+    nodes.push_front(leaves.pop_back());
+}
+while nodes.len() > 1 {
+    let right = nodes.pop_front();
+    let left = nodes.pop_front();
+    nodes.push_back(blake3::hash(SCALE::encode((left, right))));
+}
+let merkle_tree_root = if nodes.is_empty() { [0u8; 32] } else { nodes.back() };
```
carlosala, line 296: I'd rather simplify the code to make it easier to read. Just setting nodes as leaves and going `ahead!
```suggestion
let nodes = leaves;
while nodes.len() > 1 {
    let right = nodes.pop_back();
    let left = nodes.pop_back();
    nodes.push_front(blake3::hash(scale::encode((left, right))));
}

let merkle_tree_root = if nodes.is_empty() { [0u8; 32] } else { nodes.back() };
```

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,410 @@
...
+
+leaves: 0
+nodes: [[3, 4] [1, 2]]
+
+leaves:
+nodes: [[0] [3, 4] [1, 2]]
+
+nodes: [[1, 2] [[3, 4] [0]]]
+
+nodes: [[[[3, 4] [0]], [1, 2]]]
```
carlosala, line 317:
```suggestion
nodes: 0 1 2 3 4

nodes: [3, 4] 0 1 2

nodes: [1, 2] [3, 4] 0

nodes: [[3, 4], 0] [1, 2]

nodes: [[[3, 4], 0], [1, 2]]
```

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,410 @@
...
+
+leaves: 0 1
+nodes: [[4, 5] [2, 3]]
+
+leaves:
+nodes: [[4, 5] [2, 3] [0, 1]]
+
+nodes: [[0, 1] [[2, 3], [4, 5]]]
+
+nodes: [[[[2, 3], [4, 5]] [0, 1]]]
```
carlosala, line 347:
```suggestion
nodes: 0 1 2 3 4 5

nodes: [4, 5] 0 1 2 3

nodes: [2, 3] [4, 5] 0 1

nodes: [0, 1] [2, 3] [4, 5]

nodes: [[2, 3], [4, 5]] [0, 1]

nodes: [[[2, 3], [4, 5]], [0, 1]]
```

carlosala (changes requested):

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,393 @@
...
+    I16,
+    I32,
+    I64,
+    I128,
+    I256,
+    CompactU8,
+    CompactU16,
+    CompactU32,
+    CompactU64,
+    CompactU128,
```
carlosala, line 195: Shall we add `CompactU256` to make it 100% future-proof? It might happen at some point.

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,393 @@
...
+
+There should be no measurable impact on performance to Polkadot or any other chain using this feature. The metadata root hash is calculated at compile time and at runtime it is optionally used when checking the signature of a transaction. This means that at runtime no performance heavy operations are done. 
+
+###  Ergonomics & Compatibility
+
+The proposal alters the way a transaction is built, signed, and verified. So, this imposes some required changes to any kind of developer who wants to construct transactions for Polkadot or any chain using this feature. As the developer can pass `0` for disabling the verification of the metadata root hash, it can be easily ignored.
+
+## Prior Art and References
+
+[RFC 46](https://github.com/polkadot-fellows/RFCs/pull/46) produce by the Alzymologist team is a previous work reference that goes in this direction as well.
```
carlosala, line 380:
```suggestion
[RFC 46](https://github.com/polkadot-fellows/RFCs/pull/46) produced by the Alzymologist team is a previous work reference that goes in this direction as well.
```

carlosala (changes requested):

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,393 @@
...
+        base58_prefix: u16,
+        decimals: u8,
+        token_symbol: String,
+    },
+}
+```
+
+The `Hash` is 32 bytes long and `blake3` is used for calculating it. The hash of the `MetadataDigest` is calculated by `blake3(SCALE(MetadataDigest))`. Therefore, `MetadataDigest` is at first `SCALE` encoded, and then those bytes are hashed.
+
+The `MetadataDigest` itself is represented as an `enum`. This is done to make it future proof, because a `SCALE` encoded `enum` is prefixed by the `index` of the variant. This `index` represents the version of the digest. As seen above, there is no `index` zero and it starts directly with one. Version one of the digest contains the following elements:
```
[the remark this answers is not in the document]
bkchr: We can not change SCALE that easily any way. So, I would not mention this here yet.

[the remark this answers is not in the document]
carlosala: Fair enough 👍🏻

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+The `TypeDef` variants have the following meaning:
+
+- `Composite`: A `struct` like type that is composed of multiple different fields. Each `Field` can have its own type. A `Composite` with no fields is expressed as primitive type `Void`.
```
ascjones, line 212: Maybe mention the order of the fields is significant

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+- `Enumeration`: Stores a `EnumerationVariant`. A `EnumerationVariant` is a struct that is described by a name, an index and a vector of `Field`s, each of which can have it's own type. Typically `Enumeration`s have more than just one variant, and in those cases `Enumeration` will appear multiple times, each time with a different variant, in the type information. `Enumeration`s can become quite large, yet usually for decoding a type only one variant is required, therefore this design brings optimizations and helps reduce the size of the proof. An `Enumeration` with no variants is expressed as primitive type `Void`.
```
ascjones, line 213: I think it is worth mentioning, as below that each of the `Enumeration` instances will have the same `TypeRef` lookup id. This detail helped me understand how this worked.

> For `Enumeration`s all variants have the same unique identifier, while they are represented as multiple type information. All variants need to have the same unique identifier as the reference doesn't know which variant will appear in the actual encoded data.

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+On other ecosystems, there are other solutions to the problem of trusted signing. Cosmos for example has a standardized way of transforming a transaction into some textual representation and this textual representation is included in the signed data. Basically achieving the same as what the RFC proposes, but it requires that for every transaction applied in a block, every node in the network always has to generate this textual representation to ensure the transaction signature is valid.
+
+## Unresolved Questions
+
+None.
+
+## Future Directions and Related Material
+
+- Does it work with all kind of offline wallets?
+- Generic types currently appear multiple times in the metadata with each instantiation. It could be may be useful to have generic type only once in the metadata and declare the generic parameters at their instantiation.
```
ascjones, line 392: I once tried to do this at the `scale-info` level, see https://github.com/paritytech/scale-info/pull/12, but abandoned it.

It would be considerably easier to do as a second pass over the already generated type information.

text/0078-merkleized-metadata.md
bkchr: I mean this is explained in `Generating TypeRef`.

text/0078-merkleized-metadata.md
```diff
@@ -0,0 +1,394 @@
...
+
+## Unresolved Questions
+
+None.
+
+## Future Directions and Related Material
+
+- Does it work with all kind of offline wallets?
+- Generic types currently appear multiple times in the metadata with each instantiation. It could be may be useful to have generic type only once in the metadata and declare the generic parameters at their instantiation. 
+- The metadata doesn't contain any kind of semantic information. This means that the offline wallet for example doesn't know what is a balance etc. The current solution for this problem is to match on the type name, but this isn't a sustainable solution.
```
eskimor, line 393: Indeed. My main complaint about cold wallets at this point is that the rendering is sometimes not helpful. If e.g. an account address is rendered as a decimal or hexadecimal array ... that's not helpful in checking the correctness.

lolmcshizz:
/rfc process 0x5945e40d7cc049b8176a4de631ab20d7e2bb8ac9576fa4de2e620ec7660a0272

PR #84, opened 2024-03-12
General transactions in extrinsic format

georgepisaltu:


xlc:
No mention of the 2 additional extrinsic types to be introduced?

Can we at least have a table like this?

| bits | type |
|--------|--------|
| 00 | unsigned |
| 10 | signed |
| 01 | Reserved |
| 11 | Reserved |

joepetrowski:
I agree a table like that could be helpful, but I don't think we need more detail on exactly what `0x01` and `0x11` will represent. Just saying "Reserved" should be OK, the point of the RFC is that we need to encode more than just "Signed" and "Unsigned" and that we probably don't need 7 bits for the extrinsic version.

georgepisaltu:
> Can we at least have a table like this?

@xlc I added the table.

text/0084-general-transaction-extrinsic-format.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Summary
+
+This RFC proposes the expansion of the extrinsic type identifier from the current most significant bit of the first byte of the encoded extrinsic to the two most significant bits of the first byte of the encoded extrinsic.
+
+## Motivation
+
+In order to support transactions that use different authorization schemes than currently exist, a new extrinsic type must be introduced alongside the current signed and unsigned types. Currently, an encoded extrinsic's first byte indicate the type of extrinsic using the most significant bit - `0` for unsigned, `1` for signed - and the 7 following bits indicate the extrinsic format version, which has been equal to `4` for a long time.
+
+By taking one bit from the extrinsic format version encoding, we can support 2 additional extrinsic types while also having a minimal impact on our capability to extend and change the extrinsic format in the future.
```
eskimor, line 17: Going a bit more in the details here. There is a reference below, but what would be good to have right at the RFC (in my opinion):

1. What is the concrete usecase/what do we want to achieve and why? What benefits does it bring? "Different authorization" schemes is pretty generic ... what is wrong with the existing ones?
2. What other ways of adding these different authorization schemes have been considered? E.g. Why are they not covered by "unsigned"?

text/0084-general-transaction-extrinsic-format.md
```diff
@@ -0,0 +1,72 @@
...
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 12 March 2024                                                                               |
+| **Description** | Support more extrinsic types by updating the extrinsic format                               |
+| **Authors**     | George Pisaltu                                                                              |
+
+## Summary
+
+This RFC proposes the expansion of the extrinsic type identifier from the current most significant bit of the first byte of the encoded extrinsic to the two most significant bits of the first byte of the encoded extrinsic.
```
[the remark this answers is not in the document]
georgepisaltu: Done.

text/0084-general-transaction-extrinsic-format.md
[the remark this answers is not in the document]
georgepisaltu: I updated the motivation. Let me know what you think.

text/0084-general-transaction-extrinsic-format.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Motivation
+
+In order to support transactions that use different authorization schemes than currently exist, a new extrinsic type must be introduced alongside the current signed and unsigned types. Currently, an encoded extrinsic's first byte indicate the type of extrinsic using the most significant bit - `0` for unsigned, `1` for signed - and the 7 following bits indicate the extrinsic format version, which has been equal to `4` for a long time.
+
+By taking one bit from the extrinsic format version encoding, we can support 2 additional extrinsic types while also having a minimal impact on our capability to extend and change the extrinsic format in the future.
+
+## Stakeholders
+
+Runtime users.
```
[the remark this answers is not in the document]
georgepisaltu: Done.

text/0084-general-transaction-extrinsic-format.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Explanation
```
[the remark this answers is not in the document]
georgepisaltu: Done.

text/0084-general-transaction-extrinsic-format.md
georgepisaltu: I tried to answer your questions in the new iteration of `Motivation`. Let me know what you think.

text/0084-general-transaction-extrinsic-format.md
```diff
@@ -0,0 +1,78 @@
...
+
+"General" transactions, a new type of transaction that this RFC aims to support, are transactions which obey the runtime's extensions and have according extension data yet do not have hard-coded signatures. They are first described in [Extrinsic Horizon](https://github.com/paritytech/polkadot-sdk/issues/2415) and supported in [3685](https://github.com/paritytech/polkadot-sdk/pull/3685). They enable users to authorize origins in new, more flexible ways (e.g. ZK proofs, mutations over pre-authenticated origins). As of now, all transactions are limited to the account signing model for origin authorization and any additional origin changes happen in extrinsic logic, which cannot leverage the validation process of extensions.
+
+An example of a use case for such an extension would be sponsoring the transaction fee for some other user. A new extension would be put in place to verify that a part of the initial payload was signed by the author under who the extrinsic should run and change the origin, but the payment for the whole transaction should be handled under a sponsor's account. A POC for this can be found in [3712](https://github.com/paritytech/polkadot-sdk/pull/3712).
```
bkchr, line 17:
```suggestion
An example of a use case for such an extension would be sponsoring the transaction fee for some other user. A new extension would be put in place to verify that a part of the initial payload was signed by the author under whom the extrinsic should run and change the origin, but the payment for the whole transaction should be handled under a sponsor's account. A POC for this can be found in [3712](https://github.com/paritytech/polkadot-sdk/pull/3712).
```

Not sure if `whom` here is really correct :D

bkchr (approved):

text/0084-general-transaction-extrinsic-format.md
joepetrowski: Yeah `whom` is correct :)

text/0084-general-transaction-extrinsic-format.md
eskimor: Thank you! Better proxy accounts - got it ;-) Now that's a good motivation!

eskimor (approved):

muharem (approved):

bkchr:
> The clients look up them in metadata, and this change will affect them.

No, this doesn't change the metadata. Generally it didn't requires any changes so far from the metadata side as the handling is basically the same. Also the encoding of the transaction extensions. The metadata basically just contains the types of signed extra and the type inside the extension itself and the name of the extension to know what it is. Not more.

joepetrowski:
/rfc process 0xdfba8ee9626a5a211a4bbd1bd33a56c1459f4cbccfecf7dcb84cc16be6ccfbf9

PR #88, opened 2024-04-25
Slashable deposit, purchaser reputation, and reserved cores for Broker pallet

ltfschoen:


text/0088-broker-pallet-slashable-deposit-purchaser-reputation-reserved-cores.md
```diff
@@ -0,0 +1,96 @@
...
+
+There are sales of Kusama Coretime cores that are scheduled to occur later this month by Coretime Marketplace [Lastic.xyz](https://www.lastic.xyz/kusama/bulkcore1) initially in limited quantities, and potentially also by [RegionX](https://regionx.tech/) in future that is subject to their [Polkadot referendum #582](https://polkadot.polkassembly.io/referenda/582). This poses a risk in that some Kusama Coretime core purchasers may buy Kusama Coretime cores when they have no intention of actually placing a workload on them or leasing them out, which would prevent those that wish to purchase and actually use Kusama Coretime cores from being able to use any at cores at all.
+
+### Problem
+
+The types of purchasers may include:
+
+* Collectors (e.g. purchase a significant core such as the first core that is sold just to increase their likelihood of receiving an NFT airdrop for being one of the first purchasers).
+* Resellers (e.g. purchase a core that may be used at a popular period of time to resell closer to the date to realise a profit)
+* Market makers (e.g. buy cores just to change the floor price or volume).
```
eskimor, line 26: people can do that, but it is expensive. The system is designed so that you need to invest more than you create cost for others. You can also not directly create profit out of that as the sales on the primary market get burnt. You can achieve higher prices on secondary markets, but you yourself also paid more on the primary market. 

The only real small issue is with renewals. If price goes too low, people could start hoarding cores and keep renewing them, because it is cheap enough. Given that you can not do anything with a core that is eligible for renewal except running a particular parachain on it forever, this is still just cost without any benefit, if you don't actually need to run a parachain, as you can not resell such a core.

PR #90, opened 2024-05-17
Merkleized Metadata: Bring some clarification for the extrinsic inclusion

bkchr:
This augments the `RFC78` with some clarification on the front of how to include the data into the extrinsic.


muharem (approved):

rzadp:
The bot is not prepared for edits, only for adding new RFCs.
What should the process be here exactly?

bkchr:
Yeah I also realized that the bot is not prepared :D 

I don't exactly know what the process is. Current assumption is that we will do some `RFC_APPROVE(hash_of_the_file, 90)`

bkchr:
Was approved on chain here: https://collectives.subsquare.io/fellowship/referenda/171

PR #96, opened 2024-06-19
Adding smart contracts functionality to the Coretime chain

poppyseedDev:
This RFC proposes adding smart contracts to the Coretime chain using platforms like Ethereum’s Solidity or Polkadot's Ink!. This will support automated region trading, dynamic pricing, and other advanced functionalities, expanding the chain's capabilities and fostering innovation. This suggestion follows discussions in PR #94.

### DIFF
--- a/text/00xx-smart-contracts-coretime-chain.md
+++ b/text/00xx-smart-contracts-coretime-chain.md
@@ -8,7 +8,7 @@
 
 ## Summary
 
-This RFC proposes the implementation of smart contracts on the Coretime chain to enhance flexibility and enable complex decentralized applications, including secondary market functionalities.
+This RFC proposes the integration of smart contracts on the Coretime chain to enhance flexibility and enable complex decentralized applications, including secondary market functionalities.
 
 ## Motivation
 
--- a/text/00xx-smart-contracts-coretime-chain.md
+++ b/text/00xx-smart-contracts-coretime-chain.md
@@ -1,0 +1,102 @@
+# RFC-0002: Smart Contracts on the Coretime Chain
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 2024-06-09                                                                                  |
+| **Description** | Implement smart contracts on the Coretime chain                                             |
+| **Authors**     | [Aurora Poppyseed](https://github.com/poppyseedDev/), [Phil Lucksok](https://github.com/phillux)  |
+
+## Summary
+
+This RFC proposes the implementation of smart contracts on the Coretime chain to enhance flexibility and enable complex decentralized applications, including secondary market functionalities.
+
+## Motivation
+
+Currently, the Coretime chain lacks the capability to support smart contracts, which limits the range of decentralized applications that can be developed and deployed. By enabling smart contracts, the Coretime chain can facilitate more sophisticated functionalities such as automated region trading, dynamic pricing mechanisms, and other decentralized applications that require programmable logic. This will enhance the utility of the Coretime chain, attract more developers, and create more opportunities for innovation.
+
+Additionally, while there is a proposal (#885) to allow EVM-compatible contracts on Polkadot’s Asset Hub, the implementation of smart contracts directly on the Coretime chain will provide synchronous interactions and avoid the complexities of asynchronous operations via XCM. 
+
+## Stakeholders
+
+Primary stakeholders include:
+- Developers working on the Coretime chain.
+- Users who want to deploy decentralized applications on the Coretime chain.
+- Community members interested in expanding the capabilities of the Coretime chain.
+- Secondary Coretime marketplaces.
+
+## Explanation
+
+This RFC introduces the following key components:
+
+1. **Smart Contract Support**:
+    - Implement support for deploying and executing smart contracts on the Coretime chain.
+    - Use a well-established smart contract platform, such as Ethereum’s Solidity or Polkadot's Ink!, to ensure compatibility and ease of use.
+
+2. **Storage and Execution**:
+    - Define a storage structure for smart contracts and their associated data.
+    - Ensure efficient and secure execution of smart contracts, with proper resource management and gas fee mechanisms.
+
+3. **Integration with Existing Pallets**:
+    - Ensure that smart contracts can interact with existing pallets on the Coretime chain, such as the broker pallet.
+    - Provide APIs and interfaces for seamless integration and interaction.
+
+4. **Security and Auditing**:
+    - Implement robust security measures to prevent vulnerabilities and exploits in smart contracts.
+    - Conduct thorough security audits and testing before deployment.
+
+## Drawbacks
+
+There are several drawbacks to consider:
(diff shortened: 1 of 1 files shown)


chungquantin:
To expand more on this, wonder why don't we just create a new Contract Hub system parachain that is ready to XCM any other system parachains in the ecosystem? I saw there is a proposal for expanding the functionalities of the Asset Hub via contract execution as well: https://polkadot.subsquare.io/referenda/885 and I saw this is a conflict of interests while we keep making the system parachains more complicated instead of simplifying the way other L1 parachains can XCM to these system parachains. 

Let me know if this is not a correct place to start a further discussion.

eskimor:
In my mind the complication this brings to coretime is mostly integrating one more pallet. While if we not do this, we push the complexity to the developers of smart contracts. While we are working on making XCM easier and better integrated/accessible to smart contracts/ .... it will nevertheless always be a bit more complex than just calling a function, it will always stay asynchronous, it will always be much slower and it will always be of higher cost. Therefore it makes sense to stay on a chain where you can and only go cross chain when there is an actual need. E.g. getting funds to the Coretime chain to interact with the market.

athei:
I am in favour of this. I think asking contract developers to deal with XCM is a big ask when there is essentially no need for this. Sharding is a tool for scalability but we are not there, yet.  It is good that we have XCM once sharding is useful, though.

text/00xx-smart-contracts-coretime-chain.md
```diff
@@ -0,0 +1,102 @@
...
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 2024-06-09                                                                                  |
+| **Description** | Implement smart contracts on the Coretime chain                                             |
+| **Authors**     | [Aurora Poppyseed](https://github.com/poppyseedDev/), [Phil Lucksok](https://github.com/phillux)  |
+
+## Summary
+
+This RFC proposes the implementation of smart contracts on the Coretime chain to enhance flexibility and enable complex decentralized applications, including secondary market functionalities.
```
joepetrowski, line 11:
```suggestion
This RFC proposes the integration of smart contracts on the Coretime chain to enhance flexibility and enable complex decentralized applications, including secondary market functionalities.
```

text/00xx-smart-contracts-coretime-chain.md
```diff
@@ -0,0 +1,102 @@
...
+- Community members interested in expanding the capabilities of the Coretime chain.
+- Secondary Coretime marketplaces.
+
+## Explanation
+
+This RFC introduces the following key components:
+
+1. **Smart Contract Support**:
+    - Implement support for deploying and executing smart contracts on the Coretime chain.
+    - Use a well-established smart contract platform, such as Ethereum’s Solidity or Polkadot's Ink!, to ensure compatibility and ease of use.
```
joepetrowski, line 33: Solidity and ink aren't platforms, maybe you mean, "Use a well-supported smart contract virtual machine that supports contracts written in either Solidity or ink."?
```suggestion
    - Integrate support for deploying and executing smart contracts on the Coretime chain.
    - Use a well-established smart contract platform, such as Ethereum’s Solidity or Polkadot's Ink!, to ensure compatibility and ease of use.
```

text/00xx-smart-contracts-coretime-chain.md
```diff
@@ -0,0 +1,102 @@
...
+
+There are several drawbacks to consider:
+- **Complexity**: Adding smart contracts introduces significant complexity to the Coretime chain, which may increase maintenance overhead and the potential for bugs.
+- **Performance**: The execution of smart contracts can be resource-intensive, potentially affecting the performance of the Coretime chain.
+- **Security**: Smart contracts are prone to vulnerabilities and exploits, necessitating rigorous security measures and continuous monitoring.
+
+## Testing, Security, and Privacy
+
+### Testing
+- Comprehensive unit tests and integration tests should be developed to ensure the correct functionality of smart contracts.
```
joepetrowski, line 57: This seems outside the scope of the RFC, since by definition we can't control what contracts people deploy. Could you clarify to state that the tests should cover the functionality of the contract virtual machine and interactions that it can have with the rest of the runtime?

text/00xx-smart-contracts-coretime-chain.md
```diff
@@ -0,0 +1,102 @@
...
+- Continuous monitoring and updates will be necessary to address new security threats.
+
+### Privacy
+- The proposal does not introduce new privacy concerns as it extends existing functionalities with programmable logic.
+
+## Performance, Ergonomics, and Compatibility
+
+### Performance
+- The introduction of smart contracts may impact performance due to the additional computational overhead.
+- Optimization techniques, such as efficient gas fee mechanisms and resource management, should be employed to minimize performance degradation.
```
joepetrowski, line 72: This should be pretty easy to mitigate by giving contracts a custom `DispatchClass` and limiting how much of the block they can consume, so we ensure some amount is always reserved for "core" Coretime functionality.

text/00xx-smart-contracts-coretime-chain.md
poppyseedDev: @joepetrowski thanks for catching that. By this I meant that there need to be tests that ensure that the broker pallet functionality that is for example exposed to the smart contracts has doesn't pose any vulnerabilities.

poppyseedDev:
Seems like the sentiment is overall positive for this proposal, could we maybe proceed with the voting process?

text/00xx-smart-contracts-coretime-chain.md
```diff
@@ -0,0 +1,102 @@
+# RFC-0002: Smart Contracts on the Coretime Chain
```
joepetrowski, line 1:
```suggestion
# RFC-0096: Smart Contracts on the Coretime Chain
```

text/00xx-smart-contracts-coretime-chain.md
```diff
@@ -0,0 +1,102 @@
...
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 2024-06-09                                                                                  |
+| **Description** | Implement smart contracts on the Coretime chain                                             |
```
joepetrowski, line 6:
```suggestion
| **Description** | Integrate smart contracts with the Coretime chain                                             |
```

text/00xx-smart-contracts-coretime-chain.md
joepetrowski, line 1: File name should be `0096`, not `00xx`

athei:
I think the RFC should specify that we want to deploy the same pallet that we deploy on AssetHub for contract execution. While this seems to be implied it should be explicitly stated in my opinion. This is important to make sure that we don't end up with two incompatible solutions and set of toolings.

In this scenario runtime functionality will be exposed via pre-compiles. I assume that there will just be a different set of pre-compiles on AssetHub vs. Coretime. From the point if view of Solidity developers it is just a different set of .sol interface files they can make use of.

> Seems like the sentiment is overall positive for this proposal, could we maybe proceed with the voting process?

Once the open suggestions are addressed I can propose it.

PR #99, opened 2024-07-03
Introduce a transaction extension version

bkchr:
This RFC proposes to introduce a transaction extensions version. It is proposed to piggyback on [RFC84](https://github.com/paritytech/polkadot-sdk/issues/2415) to not require a new extrinsic format version. With this RFC it will be possible to change the transaction extensions without breaking the extrinsic format of a chain and thus, staying backwards compatible.

### DIFF
--- a/text/0099-transaction-extension-version.md
+++ b/text/0099-transaction-extension-version.md
@@ -44,7 +44,7 @@ In the chain runtime the version can be used to determine which set of transacti
 
 ## Drawbacks
 
-This adds a least one byte more to each signed transaction. 
+This adds one byte more to each signed transaction. 
 
 ## Testing, Security, and Privacy
 


muharem (commented):
If there is a new version, the client still has to support it, and the tx with old version and extension set is invalid for the runtime. 
Thus the version only helps a client to prepare in advance for the new extension set.
I think the same can be achieved by a client with a chain's metadata, since it contains the list of extensions identifiers.

text/0099-transaction-extension-version.md
```diff
@@ -0,0 +1,79 @@
...
+
+The extrinsic format supports the following types of transactions:
+- `Bare`: Does not add anything to the extrinsic.
+- `Signed`: `(Address, Signature, Extensions)`
+- `General`: `Extensions`
+
+The `Signed` and `General` transaction would change to:
+
+- `Signed`: `(Address, Signature, Compact, Extensions)`
+- `General`: `(Compact, Extensions)`
```
muharem, line 39: One thing to consider here is that for the General extrinsic we need to pass the extension version to `validate` function of `TransactionExtension` trait (from https://github.com/paritytech/polkadot-sdk/pull/3685). Only this way we can include it to the payload of a signature which most probably (if any) will be placed in the first extension. I think it can be passed as `inherited_implication`, together with a `call`.

bkchr: This RFC is only about changing the extrinsic format and doesn't really care how the implementation is working ;)

bkchr:
> may be worth mentioning this in the RFC.

@muharem added a sentence. Do you think this is enough or should I add more details?

text/0099-transaction-extension-version.md
```diff
@@ -0,0 +1,81 @@
...
```
ggwpez, line 39:
```suggestion
- `Signed`: `(Address, Signature, Version, Extensions)`
- `General`: `(Version, Extensions)`
```
?

text/0099-transaction-extension-version.md
```diff
@@ -0,0 +1,81 @@
...
+
+The `Compact` being a SCALE encoded compact representing the version of the transaction extensions.
```
ggwpez, line 41:
```suggestion
The `Version` being a SCALE encoded compact `u32` representing the version of the transaction extensions.
```

text/0099-transaction-extension-version.md
ggwpez: It does not pass down the version since only the set of TEs is versioned - not the TEs individually.   
So if you want to do a breaking change in a TE, you need to add a new version the transaction format.

ggwpez (approved):
Looks good. Just naming nits.

text/0099-transaction-extension-version.md
```diff
@@ -0,0 +1,81 @@
...
+- `Bare`: Does not add anything to the extrinsic.
+- `Signed`: `(Address, Signature, Extensions)`
+- `General`: `Extensions`
+
+The `Signed` and `General` transaction would change to:
+
+- `Signed`: `(Address, Signature, Version, Extensions)`
+- `General`: `(Version, Extensions)`
+
+The `Version` being a SCALE encoded compact `u32` representing the version of the transaction extensions.
```
xlc, line 41: should we just use u8?
cons:
- only up to 256 versions. but it is very unlikely a runtime needs to support that many versions and we can recycle old versions if necessary
pros:
- compact encoding using 2 bits to indicate encoding type and that means for number >=64, it will require 2 bytes to encode instead of 1. it is likely some chains will have that many versions and have 1 byte additional overhead

muharem (approved):

text/0099-transaction-extension-version.md
muharem: Yes, I understand it. 

But I think we will need it if we want it to be part of the signature payload for General extrinsics with the extensions like - https://github.com/paritytech/polkadot-sdk/blob/31dcabf0f83c74f891dfe1976bbb89b07c6abdf1/substrate/frame/support/src/transaction_extensions.rs#L86

text/0099-transaction-extension-version.md
bkchr: I thought about this as well. I also know the limitation of only having `64` versions until we need two bytes :D 

I mean we can go with `u8` and just bump the tx main version again if we need more :see_no_evil:

text/0099-transaction-extension-version.md
bkchr: I still don't get why you need to pass it there. If you are talking about an internal change to the extension itself. This will not work with this number. Basically we will then need to copy the code and create a new extension that introduces the change. On the runtime level we then decide which type to call based on the version.

But we can discuss in the implementation.

gavofyork (approved):

shawntabrizi (approved):

kianenigma (approved):

shawntabrizi:
/rfc process 0x5cc6611824cd9f2da84db94f15d931476598834f0f073f0618c795f78ee815f8

gui1117:
Should this transaction extension version be part of the inherited implications, and also signed payload?

So that signature inside a transaction extension, and signature for signed transaction can sign for a specific transaction extension.

Otherwise there can be some conflict between multiple version, a transaction could be valid but with different effect for different set of extensions. And user can't be able to sign for a specific version.

PR #100, opened 2024-07-11
Introduce new XCM instruction: `InitiateAssetsTransfer` for complex asset transfers

acatangiu:
This is a continuation of https://github.com/polkadot-fellows/xcm-format/pull/63 following the migration of XCM RFCs to this repo.

# Summary

This RFC proposes a new instruction that provides a way to initiate asset transfers which transfer multiple types (teleports, local-reserve, destination-reserve) of assets, on remote chains using XCM alone.

The currently existing instructions are too opinionated and force each XCM asset transfer to a single transfer type (teleport, local-reserve, destination-reserve). This results in inability to combine different types of transfers in single transfer which results in overall poor UX when trying to move assets across chains.

# Example usage

```rust
    // pay remote fees with ROCs
    let remote_fees = Some(
        AssetTransferFilter::ReserveWithdraw(
            AssetFilter::Wild(AllOf { id: rocs_id.into(), fun: WildFungibility::Fungible })
        )
    );
    // XCM to be executed locally
    let xcm = Xcm::<penpal_runtime::RuntimeCall>(vec![
        // Withdraw both ROCs and PENs from origin account
        WithdrawAsset(assets.clone().into()),
        // Execute the transfers while paying remote fees with ROCs
        InitiateAssetsTransfer {
            dest: local_asset_hub,
            assets: vec![
                // ROCs are reserve-withdrawn on AHR
                ReserveWithdraw(rocs.into()),
                // PENs are teleported to AHR
                Teleport(pens.into()),
            ],
            remote_fees,
            remote_xcm: xcm_on_ahr,
        },
    ]);
```

I also have a [proof-of-concept implementation](https://github.com/acatangiu/polkadot-sdk/blob/1738f8eebd33d87c9c09c7a6bcd7c42fc1a8aa9e/polkadot/xcm/xcm-executor/src/lib.rs#L717C4-L831C6) and a [proof-of-concept asset transfer test](https://github.com/acatangiu/polkadot-sdk/blob/1738f8eebd33d87c9c09c7a6bcd7c42fc1a8aa9e/cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-rococo/src/tests/asset_transfers.rs#L484C1-L489C60), transferring 2 different types of assets across 3 chains.

### DIFF
--- a/text/0100-xcm-multi-type-asset-transfer.md
+++ b/text/0100-xcm-multi-type-asset-transfer.md
@@ -39,13 +39,11 @@ types of transfers and reduce complexity.
 
 Bridge asset transfers greatly benefit from this change by allowing building XCM programs to transfer multiple
 assets across multiple hops in a single pseudo-atomic action.  
-For example, allows single XCM program execution to transfer multiple assets from `ParaK` on Kusama, through Kusama
-Asset Hub, over the bridge through Polkadot Asset Hub with final destination `ParaP` on Polkadot.
+For example, allows single XCM program execution to transfer multiple assets from `ParaK` on Kusama, through
+Kusama Asset Hub, over the bridge through Polkadot Asset Hub with final destination `ParaP` on Polkadot.
 
 With current XCM, we are limited to doing multiple independent transfers for each individual hop in order to
 move both "interesting" assets, but also "supporting" assets (used to pay fees).
-
-**Note:** Transferring assets that require different paths (chains along the way) is _not supported within same XCM_ because of the async nature of cross chain messages. This new instruction, however, enables initiating transfers for multiple assets that take the same path even if they require different transfer types along that path.
 
 ## Stakeholders
 
@@ -55,6 +53,24 @@ move both "interesting" assets, but also "supporting" assets (used to pay fees).
 - dApps devs
 
 ## Explanation
+
+A new instruction `InitiateAssetsTransfer` is introduced that initiates an assets transfer from the
+chain it is executed on, to another chain. The executed transfer is point-to-point (chain-to-chain)
+with all of the transfer properties specified in the instruction parameters. The instruction also
+allows specifying another XCM program to be executed on the remote chain.
+If a transfer requires going through multiple hops, an XCM program can compose this instruction
+to be used at every chain along the path, on each hop describing that specific leg of the transfer.
+
+**Note:** Transferring assets that require different paths (chains along the way) is _not supported
+within same XCM_ because of the async nature of cross chain messages. This new instruction, however,
+enables initiating transfers for multiple assets that take the same path even if they require
+different transfer types along that path.
+
+The usage and composition model of `InitiateAssetsTransfer` is the same as with existing
+`DepositReserveAsset`, `InitiateReserveWithdraw` and `InitiateTeleport` instructions. The main
+difference comes from the ability to handle assets that have different point-to-point transfer type
+between A and B. The other benefit is that it also allows specifying remote fee payment and
+transparently appends the required remote fees logic to the remote XCM.
 
 We can specify the desired transfer type for some asset(s) using:
 
@@ -93,8 +109,11 @@ This RFC proposes 1 new XCM instruction:
 /// The onward XCM also potentially contains a `BuyExecution` instruction based on the presence
 /// of the `remote_fees` parameter (see below).
 ///
+/// If a transfer requires going through multiple hops, an XCM program can compose this instruction
+/// to be used at every chain along the path, describing that specific leg of the transfer.
+///
 /// Parameters:
-/// - `dest`: The location of the transfer destination.
+/// - `dest`: The location of the transfer next hop.
 /// - `remote_fees`: If set to `Some(asset_xfer_filter)`, the single asset matching
 ///   `asset_xfer_filter` in the holding register will be transferred first in the remote XCM
 ///   program, followed by a `BuyExecution(fee)`, then rest of transfers follow.
--- a/text/0100-xcm-multi-type-asset-transfer.md
+++ b/text/0100-xcm-multi-type-asset-transfer.md
@@ -45,6 +45,8 @@ Asset Hub, over the bridge through Polkadot Asset Hub with final destination `Pa
 With current XCM, we are limited to doing multiple independent transfers for each individual hop in order to
 move both "interesting" assets, but also "supporting" assets (used to pay fees).
 
+**Note:** Transferring assets that require different paths (chains along the way) is _not supported within same XCM_ because of the async nature of cross chain messages. This new instruction, however, enables initiating transfers for multiple assets that take the same path even if they require different transfer types along that path.
+
 ## Stakeholders
 
 - Runtime users
@@ -104,7 +106,6 @@ This RFC proposes 1 new XCM instruction:
 ///
 /// Kind: *Command*.
 ///
-/// Errors:
 InitiateAssetsTransfer {
 	destination: Location,
 	assets: Vec<AssetTransferFilter>,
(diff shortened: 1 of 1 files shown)


text/0100-xcm-multi-type-asset-transfer.md
```diff
@@ -0,0 +1,279 @@
...
+///   program, followed by a `BuyExecution(fee)`, then rest of transfers follow.
+///   This guarantees `remote_xcm` will successfully pass a `AllowTopLevelPaidExecutionFrom` barrier.
+/// - `remote_xcm`: Custom instructions that will be executed on the `dest` chain. Note that
+///   these instructions will be executed after a `ClearOrigin` so their origin will be `None`.
+///
+/// Safety: No concerns.
+///
+/// Kind: *Command*.
+///
+/// Errors:
```
franciscoaguirre, line 107: None?

text/0100-xcm-multi-type-asset-transfer.md
```diff
@@ -0,0 +1,279 @@
...
+
+    // XCM to be executed at dest (Westend Asset Hub)
+    let xcm_on_dest =
+        Xcm(vec![DepositAsset { assets: Wild(All), beneficiary: beneficiary.clone() }]);
+
+    // XCM to be executed at Rococo Asset Hub
+    let context = PenpalUniversalLocation::get();
+    let reanchored_assets = assets.clone().reanchored(&local_asset_hub, &context).unwrap();
+    let reanchored_dest = destination.clone().reanchored(&local_asset_hub, &context).unwrap();
+    let reanchored_rocs_id = rocs_id.clone().reanchored(&local_asset_hub, &context).unwrap();
```
franciscoaguirre, line 173: `destination` and `rocs_id` are not defined in this example

franciscoaguirre (approved):

xlc:
not sure if I missed something but what about the order of the asset deposit?
e.g. transfer from Acala to AH, of 50 DOT and 10 USDT. it will result one UMP and one HRMP and on AH, it will receive one DMP and one HRMP. how exactly is this case handled?

acatangiu:
> not sure if I missed something but what about the order of the asset deposit? e.g. transfer from Acala to AH, of 50 DOT and 10 USDT. it will result one UMP and one HRMP at Acala. On AH, it will receive one DMP and one HRMP. how exactly is this case handled?

This new instruction is point-to-point (same as the rest). It can be used by higher layers that build cross chain xcm programs to move assets across multiple hops by calling it at every hop along the way. But the required/used path is outside the scope of concerns of the XCVM. It only executes this instr which explicitly specifies the dest (next hop).

Transferring assets that require different paths is still not supported within same XCM because of the async nature of cross chain messages..

I will include some of this information in the RFC for clarity.

text/0100-xcm-multi-type-asset-transfer.md
acatangiu: the error type is not enforced by the RFC, removed

text/0100-xcm-multi-type-asset-transfer.md
acatangiu: I added definitions for `destination` and `rocs_id` for clarity, but please note this is meant as a conceptual example. To actually compile it, it needs much more boiler-plate code..

text/0100-xcm-multi-type-asset-transfer.md
```diff
@@ -0,0 +1,282 @@
...
+
+Bridge asset transfers greatly benefit from this change by allowing building XCM programs to transfer multiple
+assets across multiple hops in a single pseudo-atomic action.  
+For example, allows single XCM program execution to transfer multiple assets from `ParaK` on Kusama, through Kusama
+Asset Hub, over the bridge through Polkadot Asset Hub with final destination `ParaP` on Polkadot.
+
+With current XCM, we are limited to doing multiple independent transfers for each individual hop in order to
+move both "interesting" assets, but also "supporting" assets (used to pay fees).
+
+**Note:** Transferring assets that require different paths (chains along the way) is _not supported within same XCM_ because of the async nature of cross chain messages. This new instruction, however, enables initiating transfers for multiple assets that take the same path even if they require different transfer types along that path.
```
xlc, line 48: should we define the behaviour when not supported parameters are passed? I want to make sure there is no undefined behaviour.
e.g. if the implementation should check and reject it, or if this is implementation-specific behaviour (i.e. out of scope of RFC but should be documented by the XCM executor)

franciscoaguirre:
> not sure if I missed something but what about the order of the asset deposit? e.g. transfer from Acala to AH, of 50 DOT and 10 USDT. it will result one UMP and one HRMP at Acala. On AH, it will receive one DMP and one HRMP. how exactly is this case handled?

If AH is used as a reserve then this would just be one HRMP and it'd be possible, no?

text/0100-xcm-multi-type-asset-transfer.md
franciscoaguirre: The actual path messages take is out of scope for the XCVM, it only knows how to send them. The executor should document how it handles this

acatangiu: I've added more details and clearly specified that this instruction is used point-to-point, and has to be composed to achieve multi-hop transfers.

For each hop, its parameters unequivocally specify what should happen for that specific hop/leg, so there should be no room for undefined behavior.

Any chain-specific checks or validations of the parameters are implementation-specific behavior.

vgeddes (approved):
The design looks good. I've confirmed in https://hackmd.io/@vgeddes/rkeSGIAYR that `InitiateAssetsTransfer` unlocks new potential for Snowbridge, mainly in improving UX.

text/0100-xcm-multi-type-asset-transfer.md
```diff
@@ -0,0 +1,301 @@
...
+These filters identify the assets to be transferred as well as the transfer type to be used for transferring
+them.
+It shall handle the local side of the transfer, then forward an onward XCM to `dest` for handling
+the remote side of the transfer.
+
+It should do so using same mechanisms as existing `DepositReserveAsset`, `InitiateReserveWithdraw`, `InitiateTeleport`
+instructions but practically combining all required XCM instructions to be remotely executed into a _single_
+remote XCM program to be sent over to `dest`.
+
+Furthermore, through `remote_fees: Option<AssetTransferFilter>`, it shall allow specifying a single asset to be used
```
vgeddes, line 147: Just an observation at the implementation level, when `dest` is a bridged location, for example, Ethereum; this instruction does initiate a multi-hop transfer. There is a hidden XCM fee charged by BridgeHub for committing `remote_xcm` for delivery to `dest`, and consequently the AH executor (SovereignPaidRemoteExporter) needs to estimate and take fees to cover that, with a degree of overestimation as xcm fees are variable).

Given that it likely that AH and BH will be collapsed into a single chain (Plaza) at some point in the future (thanks to CoreTime), I think its ok not to complicate the interface of `InitiateAssetsTransfer` with this complexity. 👍

acatangiu: The last commit is exactly trying to clarify that the “dest” on the instruction is not the final destination, just the “next hop” from the point of view of assets movements (e.g. through a reserve chain). It is the destination for the current step of the transfer. Not the final destination.

----

> when dest is a bridged location like Ethereum; this instruction does end up initiating a multi-hop transfer that is somewhat hidden from the builder.

yes but it is “multi-hop” from purely a transport point of view.

from the actual XCM program point of view it is still “directly” AH->Ethereum. The XCM “origin” as seen by Ethereum is AH and the xcm program logic operates as if message was passed directly from AH.

acatangiu:
/rfc process 0xfdf10d88746da6af4245b60405acbd49fbee059d109185b97c7b5fd8541b17c4

PR #101, opened 2024-07-12
Remove `require_weight_at_most` from XCM Transact

acatangiu:
This is a continuation of https://github.com/polkadot-fellows/xcm-format/pull/55 following the migration of XCM RFCs to this repo.

# Summary

Remove the `require_weight_at_most: Weight` parameter of the Transact instruction.

# Motivation

The UX of using Transact is not great, and one part of the problem is guesstimating this require_weight_at_most.
We've seen multiple Transacts on-chain failures caused by the "incorrect" use or the parameter. In practice, this parameter only adds UX overhead. Use cases fall in one of two categories:

1. Unpaid execution of Transacts - in these cases the require_weight_at_most is not really useful, caller doesn't have to pay for it, and on the call site it either fits the block or not;
2. Paid execution of single Transact - the weight to be spent by the Transact is already covered by the BuyExecution weight limit parameter.

We've had multiple OpenGov root/whitelisted_caller proposals initiated by core-devs, completely or partially fail because of incorrect configuration of `require_weight_at_most` parameter. This is a strong indication that the instruction in its current form is hard to use.

### DIFF
--- a/text/0101-xcm-transact-allow-unlimited-weight.md
+++ b/text/0101-xcm-transact-allow-unlimited-weight.md
@@ -1,0 +1,95 @@
+# RFC-0101: XCM Transact shall also allow Unlimited weight
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 12 July 2024                                                                                |
+| **Description** | Enchance XCM Transact to allow `Unlimited` weight inner call                                |
+| **Authors**     | Adrian Catangiu                                                                             |
+
+## Summary
+
+The `Transact` XCM instruction currently forces the user to set a specific maximum weight allowed to the inner call and then also pay for that much weight regardless of how much the call actually needs in practice.
+
+This RFC proposes improving the usability of `Transact` by having the remote chain (which executes the `Transact`), get and charge the actual weight of the inner call from its dispatch info.
+
+## Motivation
+
+The UX of using `Transact` is poor because of having to guess/estimate the `require_weight_at_most` weight used by the inner call on the target.
+
+We've seen multiple `Transact` on-chain failures caused by guessing wrong values for this `require_weight_at_most` even though the rest of the XCM program would have worked.
+
+In practice, this parameter only adds UX overhead with no real practical value. Use cases fall in one of two categories:
+1. Unpaid execution of Transacts - in these cases the `require_weight_at_most` is not really useful, caller doesn't
+have to pay for it, and on the call site it either fits the block or not;
+2. Paid execution of _single_ Transact - the weight to be spent by the Transact is already covered by the `BuyExecution`
+weight limit parameter.
+
+We've had multiple OpenGov `root/whitelisted_caller` proposals initiated by core-devs completely or partially fail
+because of incorrect configuration of `require_weight_at_most` parameter. This is a strong indication that the
+instruction is hard to use.
+
+## Stakeholders
+
+- Runtime Users,
+- Runtime Devs,
+- Wallets,
+- dApps,
+
+## Explanation
+
+The proposed enhancement is simple: change `Transact` instruction:
+
+```diff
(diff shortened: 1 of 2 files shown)


text/0101-xcm-transact-allow-unlimited-weight.md
```diff
@@ -0,0 +1,95 @@
...
+- Wallets,
+- dApps,
+
+## Explanation
+
+The proposed enhancement is simple: change `Transact` instruction:
+
+```diff
+- Transact { origin_kind: OriginKind, require_weight_at_most: Weight, call: DoubleEncoded<Call> },
++ Transact { origin_kind: OriginKind, weight_limit: WeightLimit, call: DoubleEncoded<Call> },
```
bkchr, line 44: Why do we need the `weight_limit` at all? 

I would just remove it?

text/0101-xcm-transact-allow-unlimited-weight.md
```diff
@@ -0,0 +1,95 @@
...
+
+No performance change.
+
+### Ergonomics
+
+Ergonomics are slightly improved by allowing `Unlimited` max weight for most scenarios.
+
+### Compatibility
+
+Compatible with previous XCM programs.
```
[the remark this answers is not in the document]
acatangiu: Easy (automated) conversion from old -> new. So any old program can be executed with same effects on a chain using new version.

text/0101-xcm-transact-allow-unlimited-weight.md
acatangiu: I am worried about potential use-cases where such an artificial limit might be required.

It's also potentially "breaking" if we directly remove it, as there is no conversion available, any existing program relying on that limit for some obscure reason will "break".

Both of the above are theoretical concerns, and in practice I expect nobody cares, but I would rather play it safe:
1. XCMv5: `Weight` -> `WeightLimit` - anyone who doesn't need artificial limit just uses `WeightLimit::Unlimited`, no hassle no worries
2. XCMv6+ : take a look at on-chain data, if we only `Unlimited`, we remove it

I am also not against removing it directly, looking for more feedback.

multi-step phase-out 👍  or kill it quick 👎 ?

bkchr: > It's also potentially "breaking" if we directly remove it, as there is no conversion available, any existing program relying on that limit for some obscure reason will "break".

The only reason I could see is to limit the amount of fees, but we probably already have some other instruction to limit the fees we are willing to pay?

xlc: the proposed fee register should do exactly that. it shouldn’t be possible to charge more fee than then amount allocated into the fee register

acatangiu: Yes, fees can be limited using dedicated fee mechanism, no need to mess around with weight limits for that.

I was thinking whether anyone needs to limit actual weight of "wild/unpredictable" Transact... but I can't think of a practical use-case, so I become more and more convinced we can go for it and directly remove it.

xlc: yeah in the case we do want to limit the execution, the extrinsic should provide a mechanism anyway otherwise it will be unbenchable or have unreasonable weights. so we can basically assume such case should never happen

acatangiu: done

acatangiu:
@xlc @bkchr @seadanda I think this is good to propose on-chain, wdyt?

seadanda (approved):
Just gave it another pass, I think it's ready to propose

franciscoaguirre (approved):

xlc:
/rfc process 0xff7432136fbaa2c74352e316264820c79078f2b0550d41de6e7e901e4f748833

vgeddes:
Just belatedly confirming that removing `require_weight_at_most` won't conflict with Snowbridge's upcoming Transact bridge. 👍

PR #102, opened 2024-07-13
Offchain runtime upgrades

eskimor:
One step closer to making reduction of PVF storage deposits smaller feasible and also in generally improving performance and reliability for parachains.

### DIFF
--- a/text/0102-offchain-parachain-runtime-upgrades.md
+++ b/text/0102-offchain-parachain-runtime-upgrades.md
@@ -1,0 +1,297 @@
+# RFC-0000: Feature Name Here
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 13 July 2024                                                                                |
+| **Description** | Implement off-chain parachain runtime upgrades                                              |
+| **Authors**     | eskimor                                                                                     |
+
+## Summary
+
+Change the upgrade process of a parachain runtime upgrade to become an off-chain
+process with regards to the relay chain. Upgrades are still contained in
+parachain blocks, but will no longer need to end up in relay chain blocks nor in
+relay chain state.
+
+## Motivation
+
+Having parachain runtime upgrades go through the relay chain has always been
+seen as a scalability concern. Due to optimizations in statement
+distribution and asynchronous backing it became less crucial and got
+de-prioritized, the original issue can be found
+[here](https://github.com/paritytech/polkadot-sdk/issues/971).
+
+With the introduction of Agile Coretime and in general our efforts to reduce
+barrier to entry more for Polkadot more, the issue becomes more relevant again:
+We would like to reduce the required storage deposit for PVF registration, with
+the aim to not only make it cheaper to run a parachain (bulk + on-demand
+coretime), but also reduce the amount of capital required for the deposit. With
+this we would hope for far more parachains to get registered, thousands
+potentially even ten thousands. With so many PVFs registered, updates are
+expected to become more frequent and even attacks on service quality for other
+parachains would become a higher risk.
+
+## Stakeholders
+
+- Parachain Teams
+- Relay Chain Node implementation teams
+- Relay Chain runtime developers
+
+## Explanation
+
+The issues with on-chain runtime upgrades are:
+
+1. Needlessly costly.
+2. A single runtime upgrade more or less occupies an entire relay chain block, thus it
+   might affect also other parachains, especially if their candidates are also
+   not negligible due to messages for example or they want to uprade their
+   runtime at the same time.
+3. The signalling of the parachain to notify the relay chain of an upcoming
+   runtime upgrade already contains the upgrade. Therefore the only way to rate
+   limit upgrades is to drop an already distributed update in the size of
+   megabytes: With the result that the parachain missed a block and more
+   importantly it will try again with the very next block, until it finally
+   succeeds. If we imagine to reduce capacity of runtime upgrades to let's say 1
+   every 100 relay chain blocks, this results in lot's of wasted effort and lost
+   blocks.
+
+We discussed introducing a separate signalling before submitting the actual
+runtime, but I think we should just go one step further and make upgrades fully
+off-chain.
+
+### Introduce a new UMP message type `RequestCodeUpgrade`
+
+As part of elastic scaling we are already planning to increase flexibility of [UMP
+messages](https://github.com/polkadot-fellows/RFCs/issues/92#issuecomment-2144538974), we can now use this to our advantage and introduce another UMP message:
+
+```rust
+enum UMPSignal {
+  // For elastic scaling
+  OnCore(CoreIndex),
+  // For off-chain upgrades
+  RequestCodeUpgrade(Hash),
+}
+```
+
+We could also make that new message a regular XCM, calling an extrinsic on the
+relay chain, but we will want to look into that message right after validation
+on the backers on the node side, making a straight forward semantic message more
+apt for the purpose.
+
+
+### Handle `RequestCodeUpgrade` on backers
+
+We will introduce a new request/response protocol for both collators and
+validators, with the following request/response:
+
+```rust
+struct RequestCode {
+  para_id: ParaId,
+  code_hash: Hash,
+}
+
+struct CodeResponse(Vec<u8>)
+```
+
+This protocol will be used by backers to request the PVF from collators in the
+following conditions:
+
+1. They received a collation sending `RequestCodeUpgrade`.
+2. They received a collation, but they don't yet have the code that was
+   previously registered on the relaychain. (E.g. disk pruned, new validator)
+
+In case they received the collation via PoV distribution instead of from the
+collator itself, they will use the exact same message to fetch from the valiator
+they got the PoV from.
+
+### Get the new code to all validators
+
+Once the candidate issuing `RequestCodeUpgrade` got backed on chain, validators
+will start fetching the code from the backers as part of availability
+distribution.
+
+To mitigate attack vectors we should make sure that serving requests for code
+can be treated as low priority requests. Thus I am suggesting the following
+scheme:
+
+Validators will notice via a runtime API (TODO: Define) that a new code has been requested, the
+API will return the `Hash` and a counter, which starts at some configurable
+value e.g. 10. The validators are now aware of the new hash and start fetching,
+but they don't have to wait for the fetch to succeed to sign their bitfield.
+
+Then on each further candidate from that chain that counter gets decremented.
+Validators which have not yet succeeded fetching will now try again. This game
+continues until the counter reached `0`. Now it is mandatory to have to code in
+order to sign a `1` in the bitfield.
+
+PVF pre-checking will happen after the candidate which brought the counter to
+`0` has been successfully included and thus is also able to assume that 2/3 of
+the validators have the code.
+
+This scheme serves two purposes:
+
+1. Fetching can happen over a longer period of time with low priority. E.g. if
+   we waited for the PVF at the very first avaialbility distribution, this might
+   actually affect liveness of other chains on the same core. Distributing
+   megabytes of data to a thousand validators, might take a bit. Thus this helps
+   isolating parachains from each other.
+2. By configuring the initial counter value we can affect how much an upgrade
+   costs. E.g. forcing the parachain to produce 10 blocks, means 10x the cost
+   for issuing an update. If too frequent upgrades ever become a problem for the
+   system, we have a knob to make them more costly.
+
+### On-chain code upgrade process
+
+First when a candidate is backed we need to make the new hash available
+(together with a counter) via a
+runtime API so validators in availability distribution can check for it and
+fetch it if changed (see previous section). For performance reasons, I think we
+should not do an additional call, but replace the [existing one](https://github.com/paritytech/polkadot-sdk/blob/d2fd53645654d3b8e12cbf735b67b93078d70113/polkadot/node/subsystem-util/src/runtime/mod.rs#L355) with one containing the new additional information (Option<(Hash, Counter)>).
+
+Once the candidate gets included (counter 0), the hash is given to pre-checking
+and only after pre-checking succeeded (and a full session passed) it is finally
+enacted and the parachain can switch to the new code. (Same process as it used
+to be.)
+
+### Handling new validators
+#### Backers
+
+If a backer receives a collation for a parachain it does not yet have the code
+as enacted on chain (see "On-chain code upgrade process"), it will use above
+request/response protocol to fetch it from whom it received the collation.
+
+#### Availablity Distribution
+
+Validators in availability distribution will be changed to only sign a `1` in
+the bitfield of a candidate if they not only have the chunk, but also the
+currently active PVF. They will fetch it from backers in case they don't have it
+yet.
+
+### How do other parties get hold of the PVF?
+
+Two ways:
+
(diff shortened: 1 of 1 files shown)


text/0102-offchain-parachain-runtime-upgrades.md
```diff
@@ -0,0 +1,347 @@
...
+
+We discussed introducing a separate signalling before submitting the actual
+runtime, but I think we should just go one step further and make upgrades fully
+off-chain. Which also helps bringing down deposit costs in a secure way, as we
+are also actually reducing costs for the network.
+
+### Introduce a new UMP message type `RequestCodeUpgrade`
+
+As part of elastic scaling we are already planning to increase flexibility of [UMP
+messages](https://github.com/polkadot-fellows/RFCs/issues/92#issuecomment-2144538974), we can now use this to our advantage and introduce another UMP message:
```
bkchr, line 66: "We just need this hack for one thing and will not use it for anything else" ;)

sandreim: It does feels indeed it is creeping out into other features/change like this one but it offers a lot of advantages in the short term. I would not call it a hack, but more of a generalisation of the UMP queue. The alternative is PVF versioning which I believe is the long term solution that we'll likely to develop in 2025.

bkchr: I mean for `CoreIndex` it is clearly a hack. `RequestCodeUpgrade` is an actual message that is sort of fine to be passed here. However, this brings up the topic on, should we add it to XCM? Should we make UMP messages generic, where one variant is `XCM` and the others are more UMP related?

sandreim: We don't want to add it to XCM, instead we will have a UMP queue separator between regular XCM messages and the possible additional ones for `CoreIndex` and the `RequestCodeUpgrade`. I will soon post the RFC which explains the UMP changes in detail.

sandreim: https://github.com/polkadot-fellows/RFCs/pull/103

bkchr: > We don't want to add it to XCM, instead we will have a UMP queue separator between regular XCM messages and the possible additional ones for `CoreIndex` and the `RequestCodeUpgrade`

I know what the plan was/is. However, this doesn't really invalidate what I said above.

sandreim: Ah, I see. I would prefer to make the UMP messages more generic in this case, having two variants, one wrapping XCM and the other UMPSignal as defined here. Sounds much better than using a separator.

text/0102-offchain-parachain-runtime-upgrades.md
```diff
@@ -0,0 +1,354 @@
...
+| **Start Date**  | 13 July 2024                                                                                |
+| **Description** | Implement off-chain parachain runtime upgrades                                              |
+| **Authors**     | eskimor                                                                                     |
+
+## Summary
+
+Change the upgrade process of a parachain runtime upgrade to become an off-chain
+process with regards to the relay chain. Upgrades are still contained in
+parachain blocks, but will no longer need to end up in relay chain blocks nor in
+relay chain state.
```
burdges, line 14: I mildly pushed for PVF upgrade to live in parablocks early on, but we descided for upgradfes on the relay chain since all validators need the data eventually anyways.  It's true however that (a) validator set churn makes off-chain an optimization, and being on-chain incurs extra costs, like repeated downloads.

burdges, line 14: Yes, off-chain upgrades make sense:  I mildly pushed for PVF upgrade to live in parablocks early on, but we descided for upgradfes on the relay chain since all validators need the data eventually anyways. It's true however that (a) validator set churn makes off-chain an optimization, and being on-chain incurs extra costs, like repeated downloads.

text/0102-offchain-parachain-runtime-upgrades.md
```diff
@@ -0,0 +1,354 @@
...
+This protocol will be used by backers to request the PVF from collators in the
+following conditions:
+
+1. They received a collation sending `RequestCodeUpgrade`.
+2. They received a collation, but they don't yet have the code that was
+   previously registered on the relaychain. (E.g. disk pruned, new validator)
+
+In case they received the collation via PoV distribution instead of from the
+collator itself, they will use the exact same message to fetch from the valiator
+they got the PoV from.
```
burdges, line 107: Why not make the code upgade simply be the parachain block?  Isn't that how substrate worked from the beginning?

text/0102-offchain-parachain-runtime-upgrades.md
```diff
@@ -0,0 +1,354 @@
...
+
+Validators will notice via a runtime API (TODO: Define) that a new code has been requested, the
+API will return the `Hash` and a counter, which starts at some configurable
+value e.g. 10. The validators are now aware of the new hash and start fetching,
+but they don't have to wait for the fetch to succeed to sign their bitfield.
+
+Then on each further candidate from that chain that counter gets decremented.
+Validators which have not yet succeeded fetching will now try again. This game
+continues until the counter reached `0`. Now it is mandatory to have to code in
+order to sign a `1` in the bitfield.
```
burdges, line 127: You've just pushed the availability into the last of these fake blocks here.  I guess this works, but I'm not convinced this is better than doing some big block availability variant:

We'd process the code availability in a single big parachain block, which only provides data but nerver gets executed.  This takes as long as it takes, maybe runnoing at some lower priority.  It occupies the availability code for that whole time, exactly like this scheme does.

After that runs, we have code available on chain so everyone must fetch it and build the artifact.  We must delay the PVF upgrade being usable until that build succeeds, which could be done either by a second fake parablock type, or else by some message of the sort discussed here.

text/0102-offchain-parachain-runtime-upgrades.md
```diff
@@ -0,0 +1,354 @@
...
+If a backer receives a collation for a parachain it does not yet have the code
+as enacted on chain (see "On-chain code upgrade process"), it will use above
+request/response protocol to fetch it from whom it received the collation.
+
+#### Availablity Distribution
+
+Validators in availability distribution will be changed to only sign a `1` in
+the bitfield of a candidate if they not only have the chunk, but also the
+currently active PVF. They will fetch it from backers in case they don't have it
+yet.
```
burdges, line 170: Yeah this makes sense regardless.

text/0102-offchain-parachain-runtime-upgrades.md
```diff
@@ -0,0 +1,354 @@
...
+
+1. We changed it so that a candidate is not deemed available if validators were
+   not able to fetch the PVF.
+2. Backers can always fetch the PVF from collators as part of the collation
+   fetching.
+
+But the majority of validators should always keep the latest code of any
+parachain and only prune the previous one, once the first candidate using the
+new code got finalized. This ensures that disputes will always be able to
+resolve.
```
burdges, line 193: Yeah 1 is an improvement here, previously I'd envisions parachains doing code reuploads once per day, just so the code stays in availability

text/0102-offchain-parachain-runtime-upgrades.md
```diff
@@ -0,0 +1,354 @@
...
+value e.g. 10. The validators are now aware of the new hash and start fetching,
+but they don't have to wait for the fetch to succeed to sign their bitfield.
+
+Then on each further candidate from that chain that counter gets decremented.
+Validators which have not yet succeeded fetching will now try again. This game
+continues until the counter reached `0`. Now it is mandatory to have to code in
+order to sign a `1` in the bitfield.
+
+PVF pre-checking will happen after the candidate which brought the counter to
+`0` has been successfully included and thus is also able to assume that 2/3 of
```
sandreim, line 130: Is there an expiry date for when the parachain needs to reach 0, otherwise the code upgrade is dropped ?

text/0102-offchain-parachain-runtime-upgrades.md
```diff
@@ -0,0 +1,354 @@
...
+the validators have the code.
+
+This scheme serves two purposes:
+
+1. Fetching can happen over a longer period of time with low priority. E.g. if
+   we waited for the PVF at the very first avaialbility distribution, this might
+   actually affect liveness of other chains on the same core. Distributing
```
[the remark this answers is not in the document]
eskimor: That's why we have a configurable amount of parachain blocks to do the fetching. If we ever run into availability problems we can:

1. Increase that amount of blocks we have time to fetch the PVF.
2. Limit the amount of runtime upgrades we are willing to do in a timespan and add priority fees (already planned) to requests, to secure a spot in case of competition.

Note however that right now we do distribute those upgrades within a single relaychain slot twice, once via statement distribution then via the relay chain block. In the new scheme if we set the number of required parachain blocks to 10, we reduced pressure 20 times. Thus I doubt it will be a problem in practice and if it ever were, we have means to fix it.

text/0102-offchain-parachain-runtime-upgrades.md
eskimor: Good point. Will add a a section.

eskimor:
For fees with this proposal, given that storage cost is essentially now limited to validator disk space, we should be able to bring down deposit costs significantly. E.g. if we assume that 1TB costs 100 Euro and such a disk lives for 3 years, we have yearly costs of 33 Euro (all very roughly). Everything we store is stored on thousand validators, thus 1MB needs 1GB of storage. This means roughly storage costs for 1MB PVF of 3.3 Cent. So a full blown PVF of 5 MB roughly 17 cent. With 10% staking rewards, this means we would only need to lock up token worth of roughly 2 Euro. Even if we make it 10x that or even 100 times that, we would still be way lower than the current on chain storage.

Obviously this is just a back-of-the-envelope calculation, but assuming I missed a few cost factors (like electricity, ...), going 10x my calculation would still be cheap (20 Euro worth of token locked).

I love it and I will extend the RFC a bit more to at least have everything prepared for smart contract storage.

bkchr:
> For fees with this proposal, given that storage cost is essentially now limited to validator disk space, we should be able to bring down deposit costs significantly.

Even before this was already taking validators into account. This is a decentralized network and you have no control over how many nodes are running aka how many copies exist. Thus, the previous model could also already not include any kind of storage costs from random nodes in the network.

Your biggest argument last week was also the costs for compiling the code. I don't see how this RFC changes the cost for compiling the PVFs.

eskimor:
> Your biggest argument last week was also the costs for compiling the code. I don't see how this RFC changes the cost for compiling the PVFs.

The biggest concern was actually the blockspace used on the relay chain which is fully solved by this proposal. For preparation, indeed nothing changed. Best solution to this problem would be PolkaVM. Until we have that or some other solution, you indeed brought up a good argument to not go that low with fees. Although we should probably differentiate between storage used for the PVF which needs to be prepared and additional storage offered (e.g. for smart contracts), which don't impose a cost on PVF compilation. :thinking:

burdges:
All my above comments can be sumarized like: 

Why is this availability voting countdown hack better than simply occupying one availability core for longer?

We're not going starve the system of cores of course.  A priori, we do not really care how long an availability core stays occupied since they never delay finality.

Are you worried there are parablocks which must be aswsigned to one particular core?

If this were the concern, then we could solve this in other ways, some of which maybe more "orthogonal."  We could've a "code upgrade" system parachian into which all parachains post their code.  It'd be "virtual" in that it has no state, no collators, and no PVF of its own, but it takes arbitrarily large blocks.  

At first blush this may sound like more work, but

burdges:
I noticed "chunk" twice in this document.  If you envision ever doing reconstruction from erasure coded data, then you need approval checkers who check the erasure coding, otherwise someone could replace some chunks with garbage. 

Instead, you could've some notion of mirroring/code core, or state of an availability core, in which validators only sign the bit once they've fetched the the whole data block.  This saves some nodes reencoding the PVF since everyone wants the PVF eventually anyways.

eskimor:
> Why is this availability voting countdown hack better than simply occupying one availability core for longer?

Because that would affect block times of that parachain, if it was core sharing even block times of other chains on that core.

The counter is just an easy solution to:

1. Give it more time than normal availability - so we can re-use availability for this, without introducing another protocol.
2. Indeed it is also an easy way to do some billing. Not perfect, but kind of apt for the purpose. Updating your runtime affects all validators and is therefore rather costly, hence the worst case of wasting a bit of coretime for empty blocks seems fine (low volume, on-demand chain) and we charge for it anyway.

(1) is more important. My biggest concern is usability issues, but should be fine as well with good documentation and emitting events about counter state.

Virtual cores are an interesting idea, although I think this adds actually more complexity, both to code and to cognitive load. It would be complexity we don't need to expose though, thus maybe good. Will think about it.

In fact I plan on using the coretime chain for the initial upload of the PVF (parachain registration). 

> I noticed "chunk" twice in this document. If you envision ever doing reconstruction from erasure coded data, then you need approval checkers who check the erasure coding, otherwise someone could replace some chunks with garbage.

I don't think it makes sense to chunk the data given that all validators need the full data anyway.

text/0102-offchain-parachain-runtime-upgrades.md
```diff
@@ -0,0 +1,354 @@
...
+API will return the `Hash` and a counter, which starts at some configurable
+value e.g. 10. The validators are now aware of the new hash and start fetching,
+but they don't have to wait for the fetch to succeed to sign their bitfield.
+
+Then on each further candidate from that chain that counter gets decremented.
+Validators which have not yet succeeded fetching will now try again. This game
+continues until the counter reached `0`. Now it is mandatory to have to code in
+order to sign a `1` in the bitfield.
+
+PVF pre-checking will happen after the candidate which brought the counter to
```
eskimor, line 129: Question: Do we need to use availability bitfields here or can we rely on pre-checking only?

eskimor: Bitfields offer the advantage that we have an incentive for backers (at least for the last one) and it avoids having impose the work of pre-checking without the "attacker" having paid their bill (produced enough blocks).

eskimor: Other things to consider:

1. To remove the wart for on-demand chains of having to produce n blocks, we could introduce a backwards compatible "fast-track" fee. With this you either produce `n` blocks (backwards compatible with existing chains) or you pay the fast-track fee, which removes this restriction and also will remove the 2 session delay: We just have pre-checking only succeed if either those two sessions passed _or_ validators have seen the including block finalized and the fast-track fee has been paid. Backers can then be incentivized to provide the code by getting a cut of that fast-track fee _iff_ prechecking succeeds. Which will only succeed if validators were able to fetch the code obviously.

2. We could see a stop-gap solution until we go full off-chain by doing the following:
2.1. Introduce a requirement to be eligible for a runtime upgrade, by having produced `n` blocks since the last one. With `n` being something like 1000. This will hardly be noticeable by existing chains (backwards compatible), but will rate-limit upgrades for on-demand chains + ramp up the cost. The effective rate limit is for `x` cores being available: `x`/`10`. So with 100 cores, this would be 1/10. Meaning by fully utilizing 100 cores someone could trigger a runtime-upgrade every 10 blocks, causing 10% service degradation worst case. We can get even better by either fully implementing this RFC or by increasing `n` further.
2.2. Have that above fast-track fee to cater to legit on-demand chains and also allow for secure fast-tracking of upgrades in general.
2.3. A relay chain block containing a candidate which contains a runtime-upgrade is illegal, if the parachain has not produced `n` blocks and is not paying the fast-track fee. Note: This might be problematic, as depending on `n` this might no longer be that backwards compatible and more importantly a parachain could end up permanently DoSing itself.

Issue #92, opened 2024-05-23
Adding a `CoreIndex` commitment to candidate receipts


sandreim:
Starting this pre-RFC conversation to gather some feedback on some changes of [CandidateReceipt](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/primitives/src/v7/mod.rs#L523) and [CommitedCandidateReceipt](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/primitives/src/v7/mod.rs#L548) primitives. 

The necessary changes are:
- add a commitment to a `CoreIndex` in  `CandidateCommitments` 
- add the a `core_index: CoreIndex` field in `CandidateDescriptor` 

These are needed to remove the limitations of only using a trusted collator set for elastic scaling parachains. Without a `CoreIndex` commitment in the candidate receipt it is possible for malicious collators to spam relay chain by sending the same collation to all backing groups assigned to a parachain at a RCB. In such a scenario elasticity is effectively disabled as all backing groups would back the same candidate in parallel instead of multiple chained candidates.

The candidate receipt primitive is used across networking protocols, Parachains Runtime, node implementation, collators and even tooling. Any breaking change here is very hard to deploy  n practice without upgrading everything at the same time or breaking anyone.  So, the following is an approach without breaking things but which might be considered hacky. 

Please keep in mind that this is very importat for short/medium term in the context of the `Elastic Scaling` feature. As such, a proposal around a more flexible, backwards compatible and future-proof (allowing for more dramatic changes) is out of scope in this proposal, but otherwise something that I am working already on.

## Proposed approach 

Changes in [CandidateDescriptor](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/primitives/src/v7/mod.rs#L482):
- reclaim 32 bytes from `collator: CollatorId` and  64 bytes from `signature: CollatorSignature` fields as `reserved` fields 
- use 4 bytes out of this reclaimed space for a new `core_index: u32` field.

[CandidateCommitments](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/primitives/src/v7/mod.rs#L652) doesn't really need to be changed, but one idea is to define a special XCM instruction like `CommitCoreIndex(u32)` that is appended as the last message in the  `upward_messages` vec. The message wouldn't ever be executed, but would just serve as a commitment to a specific `CoreIndex`.

[...]

PR #103, opened 2024-07-15
Constrain parachain block validity on a specific core

sandreim:
Following the discussion on https://github.com/polkadot-fellows/RFCs/issues/92, this is a proposal to introduce the required core index commitments to make elastic scaling work securely with open collator sets.

### DIFF
--- a/text/0103-introduce-core-index-commitment.md
+++ b/text/0103-introduce-core-index-commitment.md
@@ -1,4 +1,4 @@
-# RFC-0102:  Introduce a `CoreIndex` commitment in candidate receipts
+# RFC-0103:  Introduce a `CoreIndex` commitment in candidate receipts
 
 |                 |                                                                                             |
 | --------------- | ------------------------------------------------------------------------------------------- |
--- a/text/0103-introduce-core-index-commitment.md
+++ b/text/0103-introduce-core-index-commitment.md
@@ -155,7 +155,7 @@ pub struct CandidateDescriptorV2<H = Hash> {
 	version: InternalVersion,
 	/// The core index where the candidate is backed.
 	core_index: u16,
-	/// The session index of the candidate relay parent.
+	/// The session in which the candidate is backed.
 	session_index: SessionIndex,
 	/// Reserved bytes.
 	reserved1: [u8; 25],
@@ -278,9 +278,9 @@ the new candidate receipts format.
 
 ### Validators
 
-To ensure a smooth launch, a new node feature is required. 
+To ensure a smooth launch, a new node feature is required.
 The feature acts as a signal for supporting the new candidate receipts on the node side and can 
-only be safely enabled if at least `2/3 + 1` of the validators are upgraded.
+only be safely enabled if at least `2/3 + 1` of the validators are upgraded. Node implementions need to decode the new candidate descriptor once the feature is enabled otherwise they might raise disputes and get slashed.
 
 Once enabled, the validators will skip checking the collator signature when processing the 
 candidate receipts and verify the `CoreIndex` and `SessionIndex` fields if present in the receipt.
--- a/text/0103-introduce-core-index-commitment.md
+++ b/text/0103-introduce-core-index-commitment.md
@@ -189,9 +189,14 @@ The expectation is that performance impact is negligible for sending and process
 
 ## Ergonomics
 
-Parachain that use elastic scaling must send the separator empty message followed by the `UMPSignal::OnCore` only after sending all of the UMP XCM messages.
+Parachain that use elastic scaling must send the separator empty message followed by the `UMPSignal::SelectCore` only after sending all of the UMP XCM messages.
 
 ## Compatibility
+
+### Versioning
+
+At this point there is a simple way to determine the version of the receipt, by testing for zeroed reserved bytes in the
+descriptor. Supporting future changes will require a `u8` version field to be introduced in the reserved space. We consider the current version to be 0 and the version check implicitly done when checking for reserved space to be zeroed.
 
 ### Runtime
 The first step is to remove collator signature checking logic in the runtime, but keep the node side collator signature 
@@ -207,7 +212,6 @@ The feature acts as a signal for supporting the new candidate receipts on the no
 Once enabled, the validators will skip checking the collator signature when processing the candidate receipts and verify the `CoreIndex` and `SessionIndex` fields if present in the receipit.
 
 No new implementation of networking protocol versions for collation and validation are required.
-
 
 ### Parachains
 
--- a/text/0103-introduce-core-index-commitment.md
+++ b/text/0103-introduce-core-index-commitment.md
@@ -61,13 +61,13 @@ The UMP queue layout is changed to allow the relay chain to receive both the XCM
 This way of representing the new messages has been chosen over introducing an enum wrapper to minimize breaking changes of XCM message decoding in tools like Subscan for example. 
 
 Example: 
-```
+```rust
 [ XCM message1, XCM message2, ..., EMPTY message, UMPSignal::CoreSelector ]
 ```
 
 #### `UMPSignal` messages
 
-```
+```rust
 /// An `u8` wrap around sequence number. Typically this would be the least significant byte of the parachain block number.
 pub struct CoreSelector(pub u8);
 
@@ -92,7 +92,7 @@ statically to some sane value. Parachains should prefer to have a static value t
 Considering `para_assigned_cores` is a sorted vec of core indices assigned to a parachain at the
 specified claim queue offset, validators will determine the committed core index like this:
 
-```
+```rust
 let assigned_core_index = core_selector % para_assigned_cores.len();
 let committed_core_index = para_assigned_cores[assigned_core_index];
 ```
--- a/text/0103-introduce-core-index-commitment.md
+++ b/text/0103-introduce-core-index-commitment.md
@@ -90,8 +90,8 @@ The table below represents a snapshot of the claim queue:
 
 The purpose of `ClaimQueueOffset` is to select the column from the above table. 
 For `cq_offset = 1` we get `[ Para A, Para B, Para A]` and use as input to create 
-a sorted vec with the cores A is assigned to: `[ Core 1, Core 2]` and call it `para_assigned_cores`.
-We use `core_selector` and determine the commited core index is `Core 2` like this:
+a sorted vec with the cores A is assigned to: `[ Core 1, Core 3]` and call it `para_assigned_cores`.
+We use `core_selector` and determine the commited core index is `Core 3` like this:
 
 ```rust
 let committed_core_index = para_assigned_cores[core_selector % para_assigned_cores.len()];
--- a/text/0103-introduce-core-index-commitment.md
+++ b/text/0103-introduce-core-index-commitment.md
@@ -126,10 +126,7 @@ let committed_core_index = para_assigned_cores[core_selector % para_assigned_cor
 ```
 
 Parachains should prefer to have a static `ClaimQueueOffset` value that makes sense for their 
-usecase which can be changed by governance at some future point. Changing the value dynamically 
-can be a friction point. It will work out fine to decrease the value to build more into the 
-present. But if the value is increased to build more into the future, a relay chain block will 
-be skipped.
+usecase which can be changed by governance at some future point.
 
 ### Polkadot Primitive changes
 
@@ -186,7 +183,7 @@ If the candidate descriptor is version 1, there are no changes.
 Backers must check the validity of `core_index` and `session_index` fields. 
 A candidate must not be backed if any of the following are true:
 - the `core_index` in the descriptor does not match the core the backer is assigned to
-- the `session_index` is not equal to the session of the `relay_parent` in the descriptor
+- the `session_index` is not equal to the session index the candidate is backed in
 - the `core_index` in the descriptor does not match the one determined by the 
   `UMPSignal::SelectCore` message
 
@@ -195,37 +192,26 @@ A candidate must not be backed if any of the following are true:
 
 If the candidate descriptor is version 1, there are no changes.
 
-For version 2 descriptors the runtime will determine the `core_index` similarly to backers but 
-will always ignore the committed claim queue offset and use `0`, as it only cares 
-about what candidates can be backed at current block. 
-
-As the chain advances the claims also advance to the top of the queue. The runtime will only back  
-a candidate if the claimed core selected by it's claim queue offset has reached the top of the queue 
-at the current relay chain block: `current_block_num - relay_parent_num - 1 == claim_queue_offset`.
-
-The impact of this change is that candidates built into the future (`claim queue offset > 0`) 
-can no longer be backed earlier even if the core is free and the core is assigned to the 
-parachain. 
-
-Introducing this additional check is required to ensure the runtime computes the core index
-determinstically. For example, some collator has missed his slot and the core is now used 
-to back a candidate with a higher claim queue offset. The number of assigned cores can
-be different at these two queue offsets and the committed core indices would be different. 
+For version 2 descriptors the runtime will determine the `core_index` using the same inputs 
+as backers did off-chain. It currently stores the claim queue at the newest allowed 
+relay parent corresponding to claim queue offset `0`. The runtime needs to be changed to store
+a claim queue snapshot at all allowed relay parents.
 
 
 ### Backwards compatibility
 
 There are two flavors of candidate receipts that are used in network protocols, runtime and node 
 implementation:
-- `CommittedCandidateReceipt` which includes the `CanidateDescriptor` and the `CandidateCommitments` 
-- `CandidateReceipt` which includes the `CanidateDescriptor` and just a hash of the commitments
+- `CommittedCandidateReceipt` which includes the `CandidateDescriptor` and the `CandidateCommitments` 
+- `CandidateReceipt` which includes the `CandidateDescriptor` and just a hash of the commitments
 
-We want to support both the old and new versions in the runtime and node. The implementation must 
+We want to support both the old and new versions in the runtime and node, so the implementation must 
 be able to detect the version of a given candidate receipt.
 
-The version of the descriptor is detected by checking if the `version` field is `0` and the 
-reserved fields are zeroed. If this is true it means the descriptor is version 2, 
-otherwise we consider it is version 1.
+The version of the descriptor is detected by checking the reserved fields.
+If they are not zerored, it means it is a version 1 descriptor. Otherwise the `version` field
+is used further to determine the version. It should be `0` for version 2 descriptors. If it is not 
+the descriptor has an unknown version and should be considered invalid.
 
 
 ## Drawbacks
@@ -244,8 +230,8 @@ There is no impact on privacy.
 
 ## Performance
 
-The expected performance impact is negligible for sending and processing the new UMP 
-message in the runtime as well as on the node side.
+Overall performance will be improved by not checking the collator signatures in runtime and nodes.
+The impact on the UMP queue and candidate receipt processing is negligible.
 
 The `ClaimQueueOffset` along with the relay parent choice allows parachains to optimize their 
 block production for either throughput or latency. A value of `0` with newest relay parent 
@@ -265,25 +251,28 @@ runtime or collator nodes.
 
 The proposed changes are backwards compatible in general, but additional care must be taken by 
 waiting for at least `2/3 + 1` validators to upgrade. Validators that have not upgraded will not 
-back candidates using the new descriptor format and will also initiate disputes against.
+back candidates using the new descriptor format and will also initiate disputes against these 
+candidates.
 
 ### Relay chain runtime
 
 The first step is to remove collator signature checking logic in the runtime, but keep the node 
-side collator signature 
-checks. 
+side collator signature checks. 
 
-The runtime must be upgraded to the new primitives before any collator or node is allowed to use 
-the new candidate receipts format. 
+The runtime must be upgraded to support the new primitives before any collator or node is allowed 
+to use the new candidate receipts format. 
 
 ### Validators
 
 To ensure a smooth launch, a new node feature is required.
 The feature acts as a signal for supporting the new candidate receipts on the node side and can 
-only be safely enabled if at least `2/3 + 1` of the validators are upgraded. Node implementions need to decode the new candidate descriptor once the feature is enabled otherwise they might raise disputes and get slashed.
+only be safely enabled if at least `2/3 + 1` of the validators are upgraded. Node implementions 
+need to decode the new candidate descriptor once the feature is enabled otherwise they might 
+raise disputes and get slashed.
 
-Once enabled, the validators will skip checking the collator signature when processing the 
-candidate receipts and verify the `CoreIndex` and `SessionIndex` fields if present in the receipt.
+Once the feature is enabled, the validators will skip checking the collator signature when 
+processing the candidate receipts and verify the `CoreIndex` and `SessionIndex` fields if 
+present in the receipt.
 
 No new implementation of networking protocol versions for collation and validation is required.
 
--- a/text/0103-introduce-core-index-commitment.md
+++ b/text/0103-introduce-core-index-commitment.md
@@ -69,7 +69,7 @@ UMP queue by outputting them in [upward_messages](https://github.com/paritytech/
 
 The UMP queue layout is changed to allow the relay chain to receive both the XCM messages and
 `UMPSignal` messages. An empty message (empty `Vec<u8>`) is used to mark the end of XCM messages and
-the start of `UMPSignal` messages.The `UMPSignal` is optional and can be omitted by parachains
+the start of `UMPSignal` messages. The `UMPSignal` is optional and can be omitted by parachains
 not using elastic scaling.
 
 This way of representing the new messages has been chosen over introducing an enum wrapper to
--- a/text/0103-introduce-core-index-commitment.md
+++ b/text/0103-introduce-core-index-commitment.md
@@ -233,4 +233,4 @@ N/A
 
 The implementation is extensible and future proof to some extent. With minimal or no breaking changes, additional fields can be added in the candidate descriptor until the reserved space is exhausted
 
-Once the reserved space is exhausted, versioning will be implemented. The candidate receipt format will be versioned. This will exteend to pvf execution which requires versioning for the validation function, inputs and outputs (`CandidateCommitments`).
+Once the reserved space is exhausted, versioning will be implemented. The candidate receipt format will be versioned. Versioning should also be implemented for the validation function, inputs and outputs (`CandidateCommitments`).
--- a/text/0103-introduce-core-index-commitment.md
+++ b/text/0103-introduce-core-index-commitment.md
@@ -195,7 +195,6 @@ It is mandatory for elastic parachains to switch to the new receipt format. It i
 switch to the new receipts for providing the session index for disputes.
 
 The implementation of this RFC itself must not introduce any breaking changes for the parachain runtime or collator nodes.
-Collator must compute the core index from the 
 
 ## Compatibility
 
--- a/text/0103-introduce-core-index-commitment.md
+++ b/text/0103-introduce-core-index-commitment.md
@@ -2,34 +2,33 @@
 
 |                 |                                                                                             |
 | --------------- | ------------------------------------------------------------------------------------------- |
-| **Start Date**  | 15 July 2024                                                                    |
+| **Start Date** | 15 July 2024                                                                    |
 | **Description** | Constrain parachain block validity to a specific core and session                                                                 |
-| **Authors**     | Andrei Sandu                                                                                            |
+| **Authors** | Andrei Sandu                                                                                            |
 
 ## Summary
 
 Elastic scaling is not resilient against griefing attacks without a way for a PoV (Proof of Validity)
-to commit to the particular core index it was intended for. This RFC proposes a way to include 
-core index information in the candidate commitments and the `CandidateDescriptor` data structure 
-in a backwards compatible way. Additionally it proposes the addition of a `SessionIndex` field in 
-the `CandidateDescriptor` to make dispute resolution more secure and robust. 
-
+to commit to the particular core index it was intended for. This RFC proposes a way to include
+core index information in the candidate commitments and the `CandidateDescriptor` data structure
+in a backward compatible way. Additionally, it proposes the addition of a `SessionIndex` field in
+the `CandidateDescriptor` to make dispute resolution more secure and robust.
 
 ## Motivation
 
 This RFC proposes a way to solve two different problems:
 
-1. For Elastic Scaling, it prevents anyone who has acquired a valid collation to DoS the parachain 
-   by providing the same collation to all backing groups assigned to the parachain. This can 
-   happen before the next valid parachain block is authored and will prevent the chain of 
-   candidates to be formed, reducing the throughput of the parachain to a single core.
-2. The dispute protocol relies on validators trusting the session index provided by other 
-   validators when initiating and participating in disputes. It is used to lookup validator keys 
-   and check dispute vote signatures. By adding a `SessionIndex` in the `CandidateDescriptor`, 
-   validators no longer have to trust the `Sessionindex` provided by the validator raising a 
-   dispute. It can happen that the dispute concerns a relay chain block not yet imported by a 
-   validator. In this case validators can safely assume the session index refers to the session 
-   the candidate has appeared in, otherwise the chain would have rejected candidate.
+1. For Elastic Scaling, it prevents anyone who has acquired a valid collation to DoS the parachain
+ by providing the same collation to all backing groups assigned to the parachain. This can
+ happen before the next valid parachain block is authored and will prevent the chain of
+ candidates from being formed, reducing the throughput of the parachain to a single core.
+2. The dispute protocol relies on validators trusting the session index provided by other
+ validators when initiating and participating in disputes. It is used to look up validator keys
+ and check dispute vote signatures. By adding a `SessionIndex` in the `CandidateDescriptor`,
+ validators no longer have to trust the `Sessionindex` provided by the validator raising a
+ dispute. The dispute may concern a relay chain block not yet imported by a
+ validator. In this case, validators can safely assume the session index refers to the session
+ the candidate has appeared in, otherwise, the chain would have rejected the candidate.
 
 ## Stakeholders
 
@@ -41,41 +40,42 @@ This approach and alternatives have been considered and discussed in [this issue
 
 ## Explanation
 
-The approach proposed below was chosen primarily because it minimizes the number of breaking 
-changes, the complexity and takes less implementation and testing time. The proposal is to change 
-the existing primitives while keeping binary compatibility with the older versions. We repurpose 
-unused fields to introduce core index and a session index information in the `CandidateDescriptor` 
-and extend the UMP for transporting non-XCM messages. 
+The approach proposed below was chosen primarily because it minimizes the number of breaking
+changes, the complexity and takes less implementation and testing time. The proposal is to change
+the existing primitives while keeping binary compatibility with the older versions. We repurpose
+unused fields to introduce core index and a session index information in the `CandidateDescriptor`
+and extend the UMP to transport non-XCM messages.
 
 ### Reclaiming unused space in the descriptor
 
-The `CandidateDescriptor` currently includes `collator` and `signature` fields. The collator 
-includes a signature on the following descriptor fields: parachain id, relay parent, validation 
-data hash, validation code hash and the PoV hash.
+The `CandidateDescriptor` includes `collator` and `signature` fields. The collator
+includes a signature on the following descriptor fields: parachain id, relay parent, validation
+data hash, validation code hash, and the PoV hash.
 
-However, in practice, having a collator signature in the receipt on the relay chain does not 
-provide any benefits as there is no mechanism to punish or reward collators that have provided 
+However, in practice, having a collator signature in the receipt on the relay chain does not
+provide any benefits as there is no mechanism to punish or reward collators that have provided
 bad parachain blocks.
 
-This proposal aims to remove the collator signature and all the logic that checks the collator 
-signatures of candidate receipts. We use the first 7 reclaimed bytes to represent version, 
-the core and session index, and fill the rest with zeroes. So, there is no change in the layout 
-and length of the receipt. The new primitive is binary compatible with the old one.
+This proposal aims to remove the collator signature and all the logic that checks the collator
+signatures of candidate receipts. We use the first 7 reclaimed bytes to represent the version,
+the core,  session index, and fill the rest with zeroes. So, there is no change in the layout
+and length of the receipt. The new primitive is binary-compatible with the old one.
 
 ### UMP transport
 
-[CandidateCommitments](https://github.com/paritytech/polkadot-sdk/blob/b5029eb4fd6c7ffd8164b2fe12b71bad0c59c9f2/polkadot/primitives/src/v7/mod.rs#L682) 
-remains unchanged as we will store scale encoded `UMPSignal` messages directly in the parachain 
-UMP queue by outputing them in [upward_messages](https://github.com/paritytech/polkadot-sdk/blob/b5029eb4fd6c7ffd8164b2fe12b71bad0c59c9f2/polkadot/primitives/src/v7/mod.rs#L684). 
+[CandidateCommitments](https://github.com/paritytech/polkadot-sdk/blob/b5029eb4fd6c7ffd8164b2fe12b71bad0c59c9f2/polkadot/primitives/src/v7/mod.rs#L682)
+remains unchanged as we will store scale encoded `UMPSignal` messages directly in the parachain
+UMP queue by outputting them in [upward_messages](https://github.com/paritytech/polkadot-sdk/blob/b5029eb4fd6c7ffd8164b2fe12b71bad0c59c9f2/polkadot/primitives/src/v7/mod.rs#L684).
 
-The UMP queue layout is changed to allow the relay chain to receive both the XCM messages and 
-`UMPSignal` messages. An empty message (empty `Vec<u8>`) is used to mark the end XCM messages and 
+The UMP queue layout is changed to allow the relay chain to receive both the XCM messages and
+`UMPSignal` messages. An empty message (empty `Vec<u8>`) is used to mark the end of XCM messages and
 the start of `UMPSignal` messages.
 
-This way of representing the new messages has been chosen over introducing an enum wrapper to 
-minimize breaking changes of XCM message decoding in tools like Subscan for example. 
+This way of representing the new messages has been chosen over introducing an enum wrapper to
+minimize breaking changes of XCM message decoding in tools like Subscan for example.
 
-Example: 
+Example:
+
 ```rust
 [ XCM message1, XCM message2, ..., EMPTY message, UMPSignal::SelectCore ]
 ```
@@ -83,7 +83,7 @@ Example:
 #### `UMPSignal` messages
 
 ```rust
-/// An `u8` wrap around sequence number. Typically this would be the least significant byte of the 
+/// An `u8` wrap-around sequence number. Typically this would be the least significant byte of the 
 /// parachain block number.
 pub struct CoreSelector(pub u8);
 
@@ -92,17 +92,17 @@ pub struct ClaimQueueOffset(pub u8);
 
 /// Signals sent by a parachain to the relay chain.
 pub enum UMPSignal {
-	/// A message sent by a parachain to select the core the candidate is committed to.
-	/// Relay chain validators, in particular backers, use the `CoreSelector` and `ClaimQueueOffset`
-	/// to compute the index of the core the candidate has committed to.
-	///
-	SelectCore(CoreSelector, ClaimQueueOffset),
+    /// A message sent by a parachain to select the core the candidate is committed to.
+    /// Relay chain validators, in particular backers, use the `CoreSelector` and `ClaimQueueOffset`
+    /// to compute the index of the core the candidate has committed to.
+    ///
+    SelectCore(CoreSelector, ClaimQueueOffset),
 }
 ```
 
-The parachain runtime is not concerned with the actual `CoreIndex` the candidate is intended for, 
-but must provide enough information for the validators and collators to compute it. `CoreSelector` 
-and `ClaimQueueOffset` use only 2 bytes and fulfill the requirement.
+The parachain runtime is not concerned with the actual `CoreIndex` the candidate is intended for,
+but must provide enough information for the validators and collators to compute it. `CoreSelector`
+and `ClaimQueueOffset` use only 2 bytes and fulfills the requirement.
 
 **Example:**
 
@@ -111,13 +111,13 @@ and `ClaimQueueOffset` use only 2 bytes and fulfill the requirement.
 The table below represents a snapshot of the claim queue:
 
 |  | offset = 0 | offset = 1 | offset = 2 |
-| :--:	 | :--:   | :--:   | :--:   |
-| Core 1    | **Para A**     | **Para A**     | **Para A**     |
-| Core 2   | **Para A**  | Para B  | **Para A**  |
-| Core 3   | Para B  | **Para A**  | **Para A** |
+| :--:   | :--:   | :--:   | :--:   |
+| Core 1    | **Para A** | **Para A** | **Para A** |
+| Core 2   | **Para A** | Para B  | **Para A** |
+| Core 3   | Para B  | **Para A** | **Para A** |
 
-The purpose of `ClaimQueueOffset` is to select the column from the above table. 
-For `cq_offset = 1` we get `[ Para A, Para B, Para A]` and use as input to create 
+The purpose of `ClaimQueueOffset` is to select the column from the above table.
+For `cq_offset = 1` we get `[ Para A, Para B, Para A]` and use as input to create
 a sorted vec with the cores A is assigned to: `[ Core 1, Core 3]` and call it `para_assigned_cores`.
 We use `core_selector` and determine the committed core index is `Core 3` like this:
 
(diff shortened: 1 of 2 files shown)


text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,192 @@
...
+	pub reserved2: [u8; 64],
+	/// Hash of the para header that is being generated by this candidate.
+	pub para_head: Hash,
+	/// The blake2-256 hash of the validation code bytes.
+	pub validation_code_hash: ValidationCodeHash,
+}
+
+```
+
+In the future, parts of the `reserved1` and `reserved2` bytes can be used to include additional information in the descriptor.
```
eskimor, line 91: E.g. a version of the format.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,192 @@
...
+
+Thew new primitive will look like this:
+```
+pub struct CandidateDescriptor<H = Hash> {
+	/// The ID of the para this is a candidate for.
+	pub para_id: Id,
+	/// The hash of the relay-chain block this is executed in the context of.
+	pub relay_parent: H,
+	/// The core index where the candidate is backed.
+	pub core_index: CoreIndex,
```
eskimor, line 70: This should be private. We must now use accessor functions and should enforce their usage. Those accessor functions will read the version part and either return the data here or `None` in case the version is not `0`.

sandreim: I will add a section to describe the new interface.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+| **Authors**     | Andrei Sandu                                                                                            |
+
+## Summary
+The only requirement for collator nodes is to provide valid parachain blocks to the validators of a backing group and by definition the collator set is trustless. However, in the case of elastic scaling, for security reason, collators must be trusted - non-malicious. `CoreIndex` commitments are required to remove this limitation. Additionally we are introducing a `SessionIndex` field in the `CandidateReceipt` to make dispute resolution more secure and robust. 
+
+## Motivation
+
+At present time misbehaving collator nodes can prevent their parachain from effecitvely using elastic scaling by providing the same valid block to all backing groups assigned to the parachain. This happens before the next parachain block is authored and will prevent the chain of candidates to be formed, reducing the throughput of the parachain to a single core.
+There are no special requirements from collators to do it, just being a full node is sufficient and there are no methods of punishing or rewarding good behaviour.
```
eskimor, line 15: They don't even need to be a full node. Any computer that got ahold of the collation can do it.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Reclaiming unused space in the descriptor
+The `CandidateDescriptor` currently includes `collator` and `signature` fields. The collator includes a signature on the following descriptor fields: parachain id, relay parent, validation data hash, validation code hash and the PoV hash.
+
+However, in practice, having a collator signature in the receipt on the relay chain does not provide any benefits as there is no mechanism to punish or reward collators that have provided bad parachain blocks.
+
+This proposal aims to remove the collator signature and all the logic that checks the collator signatures of candidate receupts. We use the first 6 bytes to represent the core and session index, and fill the rest with zeroes. So, there is no change in the layout and length of the receipt. The new primitive is binary compatible with the old one.
```
eskimor, line 38: Typo: receupts.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+- the UMP queue contains the core and session index commitments and the commitments value matching the ones in the descriptor.
+
+### Polkadot Primitive changes
+
+#### New [CandidateDescriptor](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/primitives/src/v7/mod.rs#L482)
+
+- reclaim 32 bytes from `collator: CollatorId` and 64 bytes from `signature: CollatorSignature` and rename to `reserved1`  and `reserved2` fields.
+- take 2 bytes from `reserved1` for a new `core_index: u16` field.  
+- take 4 bytes from `reserved` for a new `session_index: u32` field.
+- the unused reclaimed space will be filled with zeroes
```
eskimor, line 60: A bit inconsistent. Above you are saying reserved fields are zeroed, here it is only the unused reclaimed space.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+	/// - the metadata for the egress HRMP channels
+	pub relay_chain_state: sp_trie::StorageProof,
+	/// Downward messages in the order they were sent.
+	pub downward_messages: Vec<InboundDownwardMessage>,
+	/// HRMP messages grouped by channels. The messages in the inner vec must be in order they
+	/// were sent. In combination with the rule of no more than one message in a channel per block,
+	/// this means `sent_at` is **strictly** greater than the previous one (if any).
+	pub horizontal_messages: BTreeMap<ParaId, Vec<InboundHrmpMessage>>,
+	/// The core index on which the parachain block must be backed
+	pub core_index: CoreIndex,
```
eskimor, line 152: I don't think this fully resolves the issue. With this any valid collator could still choose the core index freely. Much smaller issue than before, but still. Also on top of that how do we upgrade the ParachainInherentData?

Ideally we would have a compact proof in above `relay_chain_state` telling us which cores we have available at that relay parent and then the parachain logic just iterates through them. First block goes on core 1, child goes on core 2, ....

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+Backers will make use of the core index information to validate the blocks during backing and reject blocks if:
+- the `core_index` in descriptor does not match the one in the `UMPSignal`.
+- the `core_index` in the descriptor does not match the core the backing group is assigned to
+- the `session_index` is equal to the session of the `relay_parent` in the descriptor
+
+If core index (and session index) information is not available (backers got an old candidate receipt), there will be no changes compared to current behaviour.
+
+## Drawbacks
+
+The only drawback is that further additions to the descriptor are limited to the amount of remaining unused space.
```
eskimor, line 180: One field we should introduce at some later point, is a proper version. The currently introduced one is version 0.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+checks. 
+
+The runtime must be upgraded to the new primitives before any collator or node are allowed to use the new candidate receipts format. 
+
+### Validators
+
+To ensure a smooth launch, a new node feature is required. 
+The feature acts as a signal for supporting the new candidate receipts on the node side and can only be safely enabled if at least 2/3 of the validators are upgraded.
+
+Once enabled, the validators will skip checking the collator signature when processing the new candidate receipts.
```
eskimor, line 211: and instead verify the new fields if available (reserved fields all 0).

text/0103-introduce-core-index-commitment.md
eskimor: In any case, I believe we should think about this part (how does the runtime know its core index for the current block?) a bit more.

eskimor: Ok. Just had a late night idea, doing away with both inherent and proofs. We change the UMPSignal name to `CoreSelector` or something similar. What the parachain then does is simply putting the block number of the parachain block in there. Now let's assume for a moment that we know what position in the claim queue the collation is supposed to fill. What we can do then is find all cores the parachain is scheduled and sort them by core number, e.g. 3 cores:

```
[7, 20, 48]
```
Now we take `CoreSelector` % 3 to determine the desired core.

This has a couple of nice properties:

1. No additional proofs required.
2. No additional inherent data required.
3. It autoscales. No matter how many cores there are, we have a proper assignment for each core.

The open question to resolve: How does the parachain tell us (and how would the runtime even know), what position in the claim queue it intents to occupy. Simple and working solution: Don't only pass `CoreSelector`, but also `ClaimQueuePos`.  Then the validators know. Remaining: How should the parachain runtime know? In the most simple case: It is just static, e.g. always position 1. That works and is probably good enough for almost every parachain, but if it is constant, why even pass it? Also most/all parachains will likely pick 1.... 

Reasons to pick other values: 0 ... synchronous backing, if you don't fill the block much and want low latency. 2 ... actually I don't know, maybe because you have slow networking and need the additional time to make sure the collation will make it.

TL;DR: adding `ClaimQueuePos` to the message works. How the parachain would determine anything else than a static value is an open question, but having it in the commitments leaves maximum flexibility.

sandreim: Yeah, good point. Your idea would work, but feels a bit complicated. 

A couple of things I would adjust to make it more simple:
-  Pass the CoreIndex as commitment. It is also a much cleaner interface.
- Use state proofs for the claimqueue for simplicity. They should be made compact.
- The parachain runtime stores a `BTreeSet` of what cores have been used so far at current relay parent. If relay parent advances we clear the used indices.

text/0103-introduce-core-index-commitment.md
sandreim: Yes, do you think I should detail a bit more or even include a version field already ?

text/0103-introduce-core-index-commitment.md
eskimor: I don't think this is simpler. Also the claim queue, if we consider 100 to 200 cores will have a significant size. Putting that in into each and every block seems to be a non-starter and can easily be avoided by the above. In the end what the para runtime would do is very similar to the algorithm I described for backers, but instead of doing it cost efficient on the backers we would need to do it in the runtime.

text/0103-introduce-core-index-commitment.md
eskimor: Just a hint that after we should introduce a version field once we successfully sunsetted collators sending actual signatures. The version implemented by this RFC will then be version 0.

text/0103-introduce-core-index-commitment.md
sandreim: I have spent some time to think about this and I don't have any better idea. I'll adjust the RFC and use `CoreSelector` and `ClaimQueuePos` commitments.

text/0103-introduce-core-index-commitment.md
sandreim: Added a new `Versioning` section under `Compatibility` that describes how versioning will be done with future changes.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+```
+pub struct CandidateDescriptorV2<H = Hash> {
+	/// The ID of the para this is a candidate for.
+	para_id: Id,
+	/// The hash of the relay-chain block this is executed in the context of.
+	relay_parent: H,
+	/// The core index where the candidate is backed.
+	core_index: CoreIndex,
+	/// The session index in which the candidate is backed.
+	session_index: SessionIndex,
```
eskimor, line 118: uh ok. They are actually next to each other. Then I am assuming above is a typo. Swapping the two might be a good idea: core index is 2 bytes, session index 4. This way we should end up with better alignment.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+In future format versions, parts of the `reserved1` and `reserved2` bytes can be used to include additional information in the descriptor.
+
+#### Versioned `CandidateReceipt` and `CommittedCandidateReceipt` primitives:
+
+We want to decouple the actual representation of the `CandidateReceipt` from the higher level code. This should make it easier to implement future format versions of this primitive. To hide the logic of versioning the descriptor fields will be private and accessor methods are provided.
+
+``` 
+pub enum VersionedCandidateReceipt {
```
eskimor, line 145: An alternative proposal is to only make above data structures with all fields private. Then you implement the accessor functions like this (pseudo code):

```rust
impl CandidateDescriptor {
  pub fn core_index(&self) -> Option<CoreIndex> {
    if self.reserved2 == 0 {
      Some(self.core_index)
    } else {
      None
    }
 }
 // ... same for SessionIndex
 
  pub fn signature(&self) -> Option<&CollatorSignature> {
    if self.reserved2 == 0 {
      None
    } else {
      Some(&self.reserved2)
    }
  } 
  // similar for CollatorId, but need to be reconstructed from SessionIndex, CoreIndex and reserved1 fields.
}
```

This way even internal representations stay binary compatible and we also don't need a manual `Decode` instance. Not clear what is better: manual decode or leaving the "decode" to accessor functions ... so just food for thought really.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+A manual decode `Decode`  implementation is required to account for version detection and constructing the appropriate variant.
+
+### Parachain block validation
+
+Backers will make use of the core index information to validate the blocks during backing and reject blocks if:
+- the `core_index` in descriptor does not match the one determined by the `UMPSignal::SelectCore` message
+- the `core_index` in the descriptor does not match the core the backing group is assigned to
+- the `session_index` is not equal to the session of the `relay_parent` in the descriptor
```
eskimor, line 170: In addition, all of these should also be verified by the runtime.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+checks. 
+
+The runtime must be upgraded to the new primitives before any collator or node are allowed to use the new candidate receipts format. 
+
+### Validators
+
+To ensure a smooth launch, a new node feature is required. 
+The feature acts as a signal for supporting the new candidate receipts on the node side and can only be safely enabled if at least 2/3 of the validators are upgraded.
+
+Once enabled, the validators will skip checking the collator signature when processing the candidate receipts and verify the `CoreIndex` and `SessionIndex` fields if present in the receipit.
```
eskimor, line 212: typo: receipt.

eskimor (approved):
Nice work!

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+
+### Polkadot Primitive changes
+
+#### New [CandidateDescriptor](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/primitives/src/v7/mod.rs#L482)
+
+- reclaim 32 bytes from `collator: CollatorId` and 64 bytes from `signature: CollatorSignature` and rename to `reserved1`  and `reserved2` fields.
+- take 2 bytes from `reserved1` for a new `core_index: u16` field.  
+- take 4 bytes from `reserved2` for a new `session_index: u32` field.
```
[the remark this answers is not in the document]
sandreim: this is a typo

text/0103-introduce-core-index-commitment.md
sandreim: This is pretty much similar to what I am currently implementing, except we still wrap the receipt versions in this enum which exposes accessors and handles the versioning logic in the `Decode` implementation.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+There are two flavors of candidate receipts which are used in network protocols, runtime and node implementation:
+- `CommittedCandidateReceipt` which includes the `CanidateDescriptor` and the `CandidateCommitments` 
+- `CandidateReceipt` which includes the `CanidateDescriptor` and just a hash of the commitments
+
+We want to support both the old and new versions in the runtime and node. The implementation must be able to detect the version of a given candidate receipt.
+
+`CandidateDescriptor` is at version 2 if:
+- the reserved fields are zeroed
+- the session index matches the session index of the relay parent
+- the UMP queue contains a core index commitment and it the one in the descriptor.
```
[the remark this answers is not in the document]
sandreim: For simplicity we should have it mandatory. I believe it is possible to implement this without requiring any changes from the parachain devs on top of updating the parachain system crate.

text/0103-introduce-core-index-commitment.md
sandreim: The encoded layout is different than the Rust memory layout which Rust might not obey the ordering in the type declaration.

text/0103-introduce-core-index-commitment.md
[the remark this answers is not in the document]
eskimor: Why is it simpler? Having a check, if present then must be equal, if not present, don't check .... sounds easy enough.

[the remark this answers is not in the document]
sandreim: > Reasoning: There is no benefit in enforcing the existence of the commitment, 

I agree from relay chain perspective it doesn't really matter for single core parachains.

>but there are downsides: E.g. A parachain where developers forgot about the commitment would be completely stalled, while if we are not enforcing the existence, it would always be able to operate on at least one core.

Making the core index commitment optional adds an additional knob for elastic scaling parachains, as they will have to enable the commitments in the runtime and collator node logic to parse the `SelectCore` commitment into a core index to put in the descriptor instead of being the default. There is nothing to forget as a parachain dev because all of this logic should be handled by the parachain system pallet in the runtime and `collation-generation` subsystem on the node side. Just upgrading polkadot-sdk and omninode should do the trick.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+
+Example: 
+```
+[ XCM message1, XCM message2, ..., EMPTY message, UMPSignal::CoreSelector ]
+```
+
+#### `UMPSignal` messages
+
+```
+/// A strictly increasing sequence number, tipically this would be the parachain block number.
```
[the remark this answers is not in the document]
sandreim: Yeah, sounds reasonable, why not `u8` ?

text/0103-introduce-core-index-commitment.md
sandreim: We could use the derived `Decode` and always decode it as latest version. At least for the runtime, the collator singature and id are not even used, so decoding in accessor functions is only needed on the node. It's not really decoding, more like concatenating the bytes from `core_index`, `session_index`, `reserved1` and `reserved2` can be easily used to construct the signature.

This might cause way less changes than using the `enum`.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,236 @@
...
+
+There is no impact on privacy.
+
+## Performance
+
+The expectation is that performance impact is negligible for sending and processing the UMP message has negligible performance impact in the runtime as well as on the node side.
+
+## Ergonomics
+
+Parachain that use elastic scaling must send the separator empty message followed by the `UMPSignal::SelectCore` only after sending all of the UMP XCM messages.
```
alindima, line 192: So this message is only required for elastic scaling? I thought I read that all parachains will use it

alindima: what happens if an elastic scaling parachain doesn't do this? It will fail to back anything?

text/0103-introduce-core-index-commitment.md
alindima: I like the idea of not having an Enum and having accessor functions

alindima (commented):
Looking good!

text/0103-introduce-core-index-commitment.md
sandreim: > So this message is only required for elastic scaling? I thought I read that all parachains will use it

Yes, it is mandatory to provide it if v2 receipts are used. I will rephrase.

sandreim: > what happens if an elastic scaling parachain doesn't do this? It will fail to back anything?

Good question, missing answer in the RFC. It should fallback to how it currently works: injected core index. Will add this.

text/0103-introduce-core-index-commitment.md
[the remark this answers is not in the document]
sandreim: Yes, it should wrap around even with u16, assuming the parachain lives long enough :) I will take note of this.

text/0103-introduce-core-index-commitment.md
sandreim: Seems the only downsides are that decoding these receipts will show junk in v2 fields if the descriptor is v1.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 15 July 2024                                                                    |
+| **Description** | Constrain parachain block validity to a specific core and session                                                                 |
+| **Authors**     | Andrei Sandu                                                                                            |
+
+## Summary
+
+Elastic scaling is not resiliet against griefing attacks without a way for a PoV (Proof of Validity) to commit to the particular core index it was intedened for. This RFC proposes a way to include core index information in the candidate commitments and the `CandidateDescriptor` data strcuture in a backwards compatible way. Additionally it proposes the addition of a `SessionIndex` field in the `CandidateDescriptor` to make dispute resolution more secure and robust.
```
alindima, line 11:
```suggestion
Elastic scaling is not resilient against griefing attacks without a way for a PoV (Proof of Validity) to commit to the particular core index it was intended for. This RFC proposes a way to include core index information in the candidate commitments and the `CandidateDescriptor` data strcuture in a backwards compatible way. Additionally it proposes the addition of a `SessionIndex` field in the `CandidateDescriptor` to make dispute resolution more secure and robust. 
```

Please also keep lines under 100 characters

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+
+
+## Motivation
+
+At present time misbehaving collator nodes, or anyone who has acquired a valid collation can prevent a parachain from effecitvely using elastic scaling by providing the same collation to all backing groups assigned to the parachain. This happens before the next parachain block is authored and will prevent the chain of candidates to be formed, reducing the throughput of the parachain to a single core.
+
+The session index of candidates is important for the disputes protocol as it is used to lookup validator keys and check dispute vote signatures. By adding a `SessionIndex` in the `CandidateDescriptor`, validators no longer have to trust the `Sessionindex` provided by the validator raising a dispute. It can happen that the dispute concerns a relay chain block not yet imported by a validator. In this case validators can safely assume the session index refers to the session the candidate has appeared in, otherwise the chain would have rejected candidate.
```
alindima, line 18: Considering that we now changed this RFC to also add the session index I would change the title of the RFC (which only mentions the core index commitment).
Either this or I would mention here that this change is not needed for elastic scaling but we are taking advantage and bundling these unrelated changes into one

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+
+As we dont want to have a claim queue snapshot in the parachain runtime, we need to set `ClaimQueueOffset` 
+statically to some sane value. Parachains should prefer to have a static value that makes sense for their usecase which can be changed by governance at some future point. Changing the value dynamically can be a friction point. It will work out fine to decrease the value to build more into the present. But if the value is increased to build more into the future, a relay chain block will be skipped.
+
+Considering `para_assigned_cores` is a sorted vec of core indices assigned to a parachain at the
+specified claim queue offset, validators will determine the committed core index like this:
+
+```rust
+let assigned_core_index = core_selector % para_assigned_cores.len();
+let committed_core_index = para_assigned_cores[assigned_core_index];
```
alindima, line 97: these two variables have roughly the same name and meaning

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+### Runtime
+
+The first step is to remove collator signature checking logic in the runtime, but keep the node side collator signature 
+checks. 
+
+The runtime must be upgraded to the new primitives before any collator or node are allowed to use the new candidate receipts format. 
+
+### Validators
+
+To ensure a smooth launch, a new node feature is required.
```
alindima, line 224: the runtime will also reject candidates that use v2 if the node feature is not set

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+To ensure a smooth launch, a new node feature is required. 
+The feature acts as a signal for supporting the new candidate receipts on the node side and can only be safely enabled if at least 2/3 of the validators are upgraded.
+
+Once enabled, the validators will skip checking the collator signature when processing the candidate receipts and verify the `CoreIndex` and `SessionIndex` fields if present in the receipt.
+
+No new implementation of networking protocol versions for collation and validation are required.
+
+### Parachains
+
+The implementation of this RFC will supersede the Elastic MVP feature that relies on injecting a core index in the `validator_indices` fields of the `BackedCandidate` primitive. Elastic parachains must upgrade to use the new receipt format.
```
alindima, line 233: Once all parachains upgraded, we may remove the core index injection

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+	pub fn session_index(&self) -> Option<SessionIndex>;
+
+	/// ...
+}
+
+```
+
+A manual decode `Decode`  implementation is required to account for version detection and constructing the appropriate variant.
+
+### Parachain block validation
```
alindima, line 173: should we have a similar section about how collators are going to use it?

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+| **Authors**     | Andrei Sandu                                                                                            |
+
+## Summary
+
+Elastic scaling is not resiliet against griefing attacks without a way for a PoV (Proof of Validity) to commit to the particular core index it was intedened for. This RFC proposes a way to include core index information in the candidate commitments and the `CandidateDescriptor` data strcuture in a backwards compatible way. Additionally it proposes the addition of a `SessionIndex` field in the `CandidateDescriptor` to make dispute resolution more secure and robust. 
+
+
+## Motivation
+
+At present time misbehaving collator nodes, or anyone who has acquired a valid collation can prevent a parachain from effecitvely using elastic scaling by providing the same collation to all backing groups assigned to the parachain. This happens before the next parachain block is authored and will prevent the chain of candidates to be formed, reducing the throughput of the parachain to a single core.
```
skunert, line 16:
```suggestion
At present time misbehaving collator nodes, or anyone who has acquired a valid collation can prevent a parachain from effectively using elastic scaling by providing the same collation to all backing groups assigned to the parachain. This happens before the next parachain block is authored and will prevent the chain of candidates to be formed, reducing the throughput of the parachain to a single core.
```

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+
+[CandidateCommitments](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/primitives/src/v7/mod.rs#L652) remains unchanged as we will store scale encoded `UMPSignal` messages directly in the parachain UMP queue by outputing them in the [upward_messages](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/primitives/src/v7/mod.rs#L682). 
+
+The UMP queue layout is changed to allow the relay chain to receive both the XCM messages and `UMPSignal` messages. An empty message (empty `Vec<u8>`) is used to mark the end XCM messages and the start of `UMPSignal` messages.
+
+This way of representing the new messages has been chosen over introducing an enum wrapper to minimize breaking changes of XCM message decoding in tools like Subscan for example. 
+
+Example: 
+```rust
+[ XCM message1, XCM message2, ..., EMPTY message, UMPSignal::CoreSelector ]
```
skunert, line 65: This example needs to be adjusted to `UMPSignal::SelectCore(_, _)` as defined below

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+
+```rust
+let assigned_core_index = core_selector % para_assigned_cores.len();
+let committed_core_index = para_assigned_cores[assigned_core_index];
+```
+
+
+### Polkadot Primitive changes
+
+#### New [CandidateDescriptor](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/primitives/src/v7/mod.rs#L482)
```
skunert, line 103: I think its better to use a permalink here, this one already points to a wrong code snippet.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+
+- reclaim 32 bytes from `collator: CollatorId` and 64 bytes from `signature: CollatorSignature` and rename to `reserved1`  and `reserved2` fields.
+- take 1 bytes from `reserved1` for a new `version: u8` field.
```
skunert, line 106: Are these `reserved1` and `reserved2` maybe leftovers from an older version of this doc? They are not visible anywhere except in this description.

text/0103-introduce-core-index-commitment.md
bkchr: Yeah and the motivations reads itself a little bit bumpy. Just a little introduction on that you try to solve two issues whatever.

bkchr: > dispute concerns a relay chain block not yet imported by a validator

If this is the case, doesn't this also means that the `CandidateDescriptor` is unknown and thus, we need to trust the validator giving us a valid descriptor? And the attack here basically is that the validator would maybe check against the wrong set?

bkchr: Can you not just slash the validator for it asking you to check an invalid session index? It signs the dispute and if the session index is wrong, we should be able to proof this to the runtime or?

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+## Ergonomics
+
+It is mandatory for elastic parachains to switch to the new receipt format. It is optional but desired that all parachains
+switch to the new receipts for providing the session index for disputes.
+
+Once this RFC is implemented the parachain runtime and node must not require any manual changes to use it, except if the parachain wants to change the `ClaimQueueOffset` that is used to determine the core index. 
+
+## Compatibility
+
+The proposed changes are backwards compatible in general, but additional care must be taken by waiting for enough validators to upgrade before the validators and runtime start accepting the new candidate receipts.
```
skunert, line 213: I get that we need to wait until the majority of validators have upgraded to a version that includes the implementation of this RFC. Maybe we should still mention here what exactly happens on validators who will encounter the new formats but have not upgraded. I assume they will encounter some decoding error and will just not back?

text/0103-introduce-core-index-commitment.md
bkchr: I don't get this? This is not a live feature? Why is it mentioned here?

text/0103-introduce-core-index-commitment.md
sandreim: Typo: `assigned_core_index` -> `assigned_cores_index`. Does this help ? I could also rewrite it as a one liner to avoid having to indices.

text/0103-introduce-core-index-commitment.md
sandreim: Yes, but that is temporary implementation detail which is not worth mentioning.

text/0103-introduce-core-index-commitment.md
sandreim: No, it's not live in production, but the relay chain support is there.

text/0103-introduce-core-index-commitment.md
sandreim: Not in this section, I will add something more explicit in the `Ergonomics` section.

text/0103-introduce-core-index-commitment.md
sandreim: Yeah, I will add a note. The old validators will consider the candidates invalid because they still check the collator signature which is not valid with the new descriptors.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+	/// A message sent by a parachain to select the core the candidate is commited to.
+	/// Relay chain validators, in particular backers, use the `CoreSelector` and `ClaimQueueOffset`
+	/// to compute the index of the core the candidate has commited to.
+	///
+	SelectCore(CoreSelector, ClaimQueueOffset),
+}
+```
+
+As we dont want to have a claim queue snapshot in the parachain runtime, we need to set `ClaimQueueOffset` 
+statically to some sane value. Parachains should prefer to have a static value that makes sense for their usecase which can be changed by governance at some future point. Changing the value dynamically can be a friction point. It will work out fine to decrease the value to build more into the present. But if the value is increased to build more into the future, a relay chain block will be skipped.
```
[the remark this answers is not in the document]
sandreim: It is explained below the `UMPSignal` definition.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,252 @@
...
+checks. 
+
+The runtime must be upgraded to the new primitives before any collator or node are allowed to use the new candidate receipts format. 
+
+### Validators
+
+To ensure a smooth launch, a new node feature is required. 
+The feature acts as a signal for supporting the new candidate receipts on the node side and can only be safely enabled if at least 2/3 of the validators are upgraded.
+
+Once enabled, the validators will skip checking the collator signature when processing the candidate receipts and verify the `CoreIndex` and `SessionIndex` fields if present in the receipt.
```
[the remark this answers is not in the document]
sandreim: It is mandatory for the new format to provide both `CoreIndex` and `SessionIndex`. So we just look to see if the version is 0 and if the reserved fields are both zeroed. If so, we consider the descriptor to be v2, and expect fields have correct values.

text/0103-introduce-core-index-commitment.md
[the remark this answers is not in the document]
sandreim: If the reserved fields are not zeroed, then we consider that the descriptor is v1 and expect valid signature. Nodes should check the signature in that case.

text/0103-introduce-core-index-commitment.md
sandreim: I've added an example along with the `UMPSignal::SelectCore` definition

text/0103-introduce-core-index-commitment.md
alindima: yeah I get that they are different but they had very similar names so it was a bit confusing. A one-liner would be better IMO

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,240 @@
...
+
+The parachain runtime is not concerned with the actual `CoreIndex` the candidate is intended for, 
+but must provide enough information for the validators and collators to compute it. `CoreSelector` 
+and `ClaimQueueOffset` use only 2 bytes and fullfil the requirement.
+
+**Example:**
+
+`cq_offset = 1` and `core_selector = 3`
+
+The table below represents a snapshot of the claim queue:
```
alindima, line 83: I think we discussed this offline but I wanted to make sure this is clear.
The relay chain runtime cannot take into account the claim queue offset when backing a candidate. It can only back something if the claim reaches the top of the claim queue. By which time the claim at that offset could be something different. So the runtime must not use the claim queue offset provided in the inherent data. right?

A side effect is that that even if the claim queue offset used by the parachain is 1, a candidate can be backed sooner than that if it has a claim at offset 0. So this offset is really a max offset. But this sounds fine

text/0103-introduce-core-index-commitment.md
eskimor: By having the SessionIndex in the descriptor, it is either valid or it does not exist on any chain (as the chain checks on import). If it does not exist on any chain, then the dispute is just spam and won't resolve anyways, hence there is no risk.

> Can you not just slash the validator for it asking you to check an invalid session index? It signs the dispute and if the session index is wrong, we should be able to proof this to the runtime or?

Not easily. You would need to be able to prove the session of a block of another fork.

> And the attack here basically is that the validator would maybe check against the wrong set?

yes.

It is also just "nice" as it makes candidates more self-contained: With the SessionIndex provided in the descriptor you can validate the state transition without any knowledge of the fork it appeared in.

It is true that the `SessionIndex` could be made up, but so could be the persisted validation data. Essentially what a checker checks when validating is: This is valid, assuming you find a chain that would accept the candidate/has accepted the candidate: Both persisted validation data and SessionIndex are verified on chain, thus it is not necessary to prove their validity off-chain: "Assuming this actually exists, I can confirm it is valid."

I am mostly sanity checking myself here. Any more concerns, please bring them up. @burdges @rphmeier thoughts?

text/0103-introduce-core-index-commitment.md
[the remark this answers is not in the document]
eskimor: It could, the `version` field is future stuff. We can only do this once we were able to fully deprecate/remove support for the legacy version having signatures and and ids. For launch of this we have to compare the entire signature field with `0`.

text/0103-introduce-core-index-commitment.md
eskimor: The claim queue offset is with regards to the used relay parent, hence it is fully deterministic. We also don't accept statements in the runtime other than for index 0, so we can just check the cores at position 0, if they match, we are good. The offset specification is only required for the backers to be able to deterministically check,

text/0103-introduce-core-index-commitment.md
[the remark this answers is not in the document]
eskimor: The former signature field has to be all zeros, otherwise we assume the legacy format.

text/0103-introduce-core-index-commitment.md
alindima: why is this temporary?
Personally I think it's worth mentioning

alindima (approved):
Nice job!

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,313 @@
...
+
+|  | offset = 0 | offset = 1 | offset = 2 |
+| :--:	 | :--:   | :--:   | :--:   |
+| Core 1    | **Para A**     | **Para A**     | **Para A**     |
+| Core 2   | **Para A**  | Para B  | **Para A**  |
+| Core 3   | Para B  | **Para A**  | **Para A** |
+
+The purpose of `ClaimQueueOffset` is to select the column from the above table.
```
bkchr, line 119: Does Polkadot right now allows to use a claim ahead in the queue? I still don't really get why we need it. The parachain runtime should ensure that always a new core is being used, for that using selector is enough.
Do we use right now the relay chain block of the pov was build on to determine the claim queue and then to determine the core? Or what are we doing right now?

sandreim: Right now, the core index is determined by the backers who have voted. We map validator index to group index to core index. 

The claim queue offset is required because of the `CoreSelector % number_of_assignments` operation that selects the core. Because parachains can have a varying number of cores assigned to them (on-demand) at a given RCB, the core index you get might be different depending on which claim is used.

text/0103-introduce-core-index-commitment.md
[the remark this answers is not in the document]
sandreim: Added a new section **On-chain backing** which should provide more context.

text/0103-introduce-core-index-commitment.md
bkchr: I mean I get this. However, you want to us all possible claims? So, you probably don't want to waste any cores. Which means you will build on all the cores. With prospective parachains you also can calculate which should be the next claim to use. Thus, we should not require the offset?

sandreim: > I mean I get this. However, you want to us all possible claims? So, you probably don't want to waste any cores. Which means you will build on all the cores.

Yes

> With prospective parachains you also can calculate which should be the next claim to use. Thus, we should not require the offset?

That would not be enough. In backing we need to determine if the candidate is valid on the core the backers are assigned to. We take the core selector and determine whatever core comes out of that. However, if one of these blocks doesn't make it in the next relay chain block, the relay chain cannot validate the candidate core index anymore. At that point the top of the claim queue might have a different number of assignments and you'd get a different core index for the modulo operation.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,313 @@
...
+to back a candidate with a higher claim queue offset. The number of assigned cores can
+be different at these two queue offsets and the committed core indices would be different. 
+
+
+### Backwards compatibility
+
+There are two flavors of candidate receipts that are used in network protocols, runtime and node 
+implementation:
+- `CommittedCandidateReceipt` which includes the `CanidateDescriptor` and the `CandidateCommitments` 
+- `CandidateReceipt` which includes the `CanidateDescriptor` and just a hash of the commitments
```
bkchr, line 221:
```suggestion
- `CommittedCandidateReceipt` which includes the `CandidateDescriptor` and the `CandidateCommitments` 
- `CandidateReceipt` which includes the `CandidateDescriptor` and just a hash of the commitments
```

text/0103-introduce-core-index-commitment.md
bkchr: I thought about this a little bit more and I think I found a way to not require the `offset`. My main problem with this `offset` is that it is static. You also write in the RFC itself that it isn't that great. 

So, to my proposal. I would propose we store the latest `core_selector` of each parachain in the relay chain state. Then we can use this `latest_core_selector` (fetched at the candidate context relay chain block) together with the `claim_queue` to determine if the `core_index` in the candidate is correct. First we would calculate the `core_selector_diff = core_selector.wrapping_sub(latest_core_selector)`. Then we would fetch the `claim_queue` at the candidate context relay chain block. We would count all cores assigned to our parachain until the count reaches `core_selector_diff` and then we would have our `core_index`. 

I think we can then also get rid off `max_candidate_len` from the async backing parameters as the maximum length would be determined by the "claim queue lookahead".

eskimor: For `max_candidate_len` see: https://github.com/paritytech/polkadot-sdk/issues/5079 -> is already planned.

For the rest. I am not following: 

> We would count all cores assigned to our parachain

This is not deterministic, if you don't know the position the parachain is aiming for.

> My main problem with this offset is that it is static.

It does not have to be. It just will be for most parachains as I don't see much reason to change it all the time. But the value could be adjustable by parachain governance for example.

I would expect most parachains to use 1, some maybe 0 to get the lowest latency and then maybe some to use 2 (with an older relay parent) to get more stable/performant block production at the cost of latency when it comes to message processing.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,313 @@
...
+
+```rust
+let committed_core_index = para_assigned_cores[core_selector % para_assigned_cores.len()];
+```
+
+Parachains should prefer to have a static `ClaimQueueOffset` value that makes sense for their 
+usecase which can be changed by governance at some future point. Changing the value dynamically 
+can be a friction point. It will work out fine to decrease the value to build more into the 
+present. But if the value is increased to build more into the future, a relay chain block will 
+be skipped.
```
eskimor, line 132: Maybe we should just remove this paragraph as it is mostly causing confusion.

Note also that this "friction" point has nothing to do with using a claim queue offset or some other solution - the friction of switching is fundamental and can not be mitigated by replacing `ClaimQueueOffset` with something else.

Let's iterate the two situations:

1. You decide to provide collations a bit more ahead of time, e.g. for getting backed in 2 blocks from now, instead of just one. You can do that by producing candidates faster at switch over time, sequence:

a) old settings, you produced a candidate because you saw your assignment upcoming in position 1.
b) settings switch, you are now aiming for (2).
c) you immediately produce another candidate (assuming core is assigned to you at position (2)). Basically if the core is assigned to your para at postion (1) and (2) you would produce two blocks in a row, without any delay at switch over time.

Situation two is the reverse:

(a) You build your candidate when it appeared at position (2) (old settings).
(b) setting switch, we are now aiming for (1). Now assuming we had another assignment at (3) earlier, we will now wait 12 seconds until it moved to (1) until we produced another block.


In fact there is not even too much friction. The only thing that happens is that parachain block times will have a "jump" at switch over time. In case (1), this might lead to one block being less filled for example. 


The real friction @sandreim was referring to has actually nothing to do with the claim queue itself, but rather with the relay parent advancement rule: If you move claim queue position from (1) to (2) because you want to build on the parent of the most current leaf instead of the leaf itself, in order to be more resilient to relay chain forks you will need to skip a relay chain block for the adjustment, because the relay parent block number most not be less than the previous one.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,313 @@
...
+
+```rust
+pub struct CandidateDescriptorV2<H = Hash> {
+	/// The ID of the para this is a candidate for.
+	para_id: ParaId,
+	/// The hash of the relay-chain block this is executed in the context of.
+	relay_parent: H,
+	/// Version field. The raw value here is not exposed, instead it is used
+	/// to determine the `CandidateDescriptorVersion`
+	version: InternalVersion,
```
eskimor, line 155: Maybe we should call this `version_reserved` and make the comment read:

```
Version field to be used for future updates.
```

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,313 @@
...
+
+There are two flavors of candidate receipts that are used in network protocols, runtime and node 
+implementation:
+- `CommittedCandidateReceipt` which includes the `CanidateDescriptor` and the `CandidateCommitments` 
+- `CandidateReceipt` which includes the `CanidateDescriptor` and just a hash of the commitments
+
+We want to support both the old and new versions in the runtime and node. The implementation must 
+be able to detect the version of a given candidate receipt.
+
+The version of the descriptor is detected by checking if the `version` field is `0` and the
```
eskimor, line 226: Uh ok. If we really want to do this, then disregard my comment about the version field above.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,313 @@
...
+to ensure the implementation does not introduce any new security issues.
+
+Backwards compatibility of the implementation will be tested on testnets (Versi and Westend).
+
+There is no impact on privacy.
+
+## Performance
+
+The expected performance impact is negligible for sending and processing the new UMP 
+message in the runtime as well as on the node side.
```
eskimor, line 248: In fact, we dropped a signature: Things should become faster.

eskimor (approved):

text/0103-introduce-core-index-commitment.md
sandreim: Yeah, it is not reserved, but actually used already.

text/0103-introduce-core-index-commitment.md
bkchr: > It does not have to be. It just will be for most parachains as I don't see much reason to change it all the time. But the value could be adjustable by parachain governance for example.

IMO exposing this parameter to parachain builders will be more confusing to them. They already don't know what the unincluded segment is. The unincluded segment used by the lookahead collator would not work anymore with the static offset, because it builds multiple blocks on the same relay chain block.

> This is not deterministic, if you don't know the position the parachain is aiming for.

Why is this not deterministic? The claim queue doesn't change in the past and could only change for future blocks, because of some ondemand core for example. When the collator is sending the collation, it is aware of the claim queue and where it sends the PoV. 


I also think that we will have problems with on demand? Let's assume we are able to buy a core and assign it in the same block (I know there exists somewhere an issue for this). How should the collator be able to build a block on time? Maybe making the UMP optional could be a solution for this? 

So, maybe let me propose the following:

1. We re-arrange the claim queue storage, to be able to include the claim queue in the storage proof used by the parachain. This way the parachain can report the `core` directly it wants to be scheduled on. We don't need any offset nor any selector.

2. The UMP message is made optional. This way the runtime gives up the "control" to the PoV submitter to decide on the core. This gives us the freedom we have right now and will work for all chains that don't use elastic scaling without a problem.

sandreim: > > It does not have to be. It just will be for most parachains as I don't see much reason to change it all the time. But the value could be adjustable by parachain governance for example.
> 
> IMO exposing this parameter to parachain builders will be more confusing to them. They already don't know what the unincluded segment is. 

The way this is being implemented right now doesn't require the parachain builders to care about this parameter - for most usecases `1` is a good value.
> The unincluded segment used by the lookahead collator would not work anymore with the static offset, because it builds multiple blocks on the same relay chain block.
> 

I don't understand why it wouldn't work if it builds on the same relay parent. Do you mean it can't build on all of the cores assigned by claim queue ahead of time ?

> > This is not deterministic, if you don't know the position the parachain is aiming for.
> 
> Why is this not deterministic? The claim queue doesn't change in the past and could only change for future blocks, because of some ondemand core for example. When the collator is sending the collation, it is aware of the claim queue and where it sends the PoV.
> 
> I also think that we will have problems with on demand? Let's assume we are able to buy a core and assign it in the same block (I know there exists somewhere an issue for this). How should the collator be able to build a block on time? Maybe making the UMP optional could be a solution for this?
> 

I don't think that making UMP optional helps, because you might want to get `2` cores in same relay chain block and you need the UMP commitment. For the current proposal, you would want a CQ offset of `0`. However, it would likely be a good idea to always get on-demand cores assigned with at least 1 RCB delay to make sure there is plenty of time for the collation to be built ahead of time.

> So, maybe let me propose the following:
> 
> 1. We re-arrange the claim queue storage, to be able to include the claim queue in the storage proof used by the parachain. This way the parachain can report the `core` directly it wants to be scheduled on. We don't need any offset nor any selector.

I like this idea, I wanted at some point to do it like this, However there are 2 downsides that come to mind:
- PoV overhead from storing these proofs, even if they are small
- parachain needs to keep some state about what claims are used over a window of valid relay parents

But! there are some benefits:
- Without a qc offset, parachain blocks can be built ahead of time and can get in earlier
- Handles the case when availability stalls for a few blocks for a candidate on a core. Current proposal would not work anymore if the claim queue of the stalled core does not advance and claim queue of other cores assigned to the para are popped leading to different number of assignments at cq offset 0 that is used by runtime to compute the core index.
- The core index is fully deterministic, it doesn't depend on any relay chain variables anymore.

> 2. The UMP message is made optional. This way the runtime gives up the "control" to the PoV submitter to decide on the core. This gives us the freedom we have right now and will work for all chains that don't use elastic scaling without a problem.

 I don't really think we should give control to the PoV submitter to decide the core, this RFC is actually about removing that control for improved security, but I agree the RFC is a `no-op` for single core parachains. 
In any case if the parachain itself commits to a `CoreIndex` as discussed above you get all the freedom you have right now and for simplicity lets keep UMP commitments mandatory.

sandreim: A couple of things to do to fix in current proposal so that it fully works with availability timeouts:
- keep claim queue snapshots for all relay valid parents in the runtime
- check the core index using the snapshots rather than assuming cq offset 0

This would allow us to back things earlier and still determine a valid core index in the runtime if availability is slow.

sandreim: Actually a better idea is to pop the claim queue for the stalled cores and push the claim to the back of the queue: https://github.com/paritytech/polkadot-sdk/issues/4776#issuecomment-2309949169.

sandreim: > I also think that we will have problems with on demand? Let's assume we are able to buy a core and assign it in the same block (I know there exists somewhere an issue for this). How should the collator be able to build a block on time? Maybe making the UMP optional could be a solution for this?

One easy fix is to duplicate the claim for a order if the claim queue is empty. The parachain gets two chances if it uses cq offset `0` and has enough time to do well with cq offset `1`.

alindima: > > I also think that we will have problems with on demand? Let's assume we are able to buy a core and assign it in the same block (I know there exists somewhere an issue for this). How should the collator be able to build a block on time? Maybe making the UMP optional could be a solution for this?
> 
> One easy fix is to duplicate the claim for a order if the claim queue is empty. The parachain gets two chances if it uses cq offset `0` and has enough time to do well with cq offset `1`.

Yes, this is planned to be done while solving https://github.com/paritytech/polkadot-sdk/issues/4776 (@eskimor also mentioned this)

eskimor: There is another issue with on-demand as @skunert pointed out today: Assignments can come up after the fact.

E.g. at relay parent X, the claim queue looks like this:

```
[
[A,B,C],
[A,A,C],
[],
]
```

Then the next relay parent something get scheduled for on-demand and above might become:

```
[
[B,C],
[A,C],
[E,C],
]
```

Messing up the position 0 check for parachain `C`.  Not sure yet how to solve this best. The issue is not unique to on-demand, can also happen with incomplete bulk assignments.

sandreim: Keeping a claim queue snapshot at valid relayparents in the runtime allows us to compute the core index correctly.

eskimor: Snapshots are fine. Possible future optimization (if necessary): Accompany each entry in the claim queue with the block number it came into existence. Then by only knowing the block number of the relay parent, we can filter to only the relevant entries.

bkchr: > One easy fix is to duplicate the claim for a order if the claim queue is empty. The parachain gets two chances if it uses cq offset `0` and has enough time to do well with cq offset `1`.

And if the claim queue is not empty? They loose their claim?

alindima: > > One easy fix is to duplicate the claim for a order if the claim queue is empty. The parachain gets two chances if it uses cq offset `0` and has enough time to do well with cq offset `1`.
> 
> And if the claim queue is not empty? They loose their claim?

If the parachain is using offset 1, no. They have enough time to build the candidate.
If the parachain is using offset 0, then it can only do synchronous backing if my understanding is right.

eskimor: > And if the claim queue is not empty? They loose their claim?

No. The claim would not even be lost without the duplication on empty queues, but if the queue was empty before your claim immediately goes to position 0, which means only synchronous backing. Position 0 is therefore kind of "special".

There is no such problem if the queue was not empty. If at least one item was present before, you start at position 1, which already provides capability for full blocks.

text/0103-introduce-core-index-commitment.md
sandreim: Going to keep just first part about preferring a static offset. It's up to the implementation to figure out a good default value and documentation to give details.

skunert (approved):

bkchr (commented):
I'm fine with the general structure of the RFC. I'm only requesting some more clarifications, especially around the offset/selector.

I also want to see that the `UMPSignal` is optional. Maybe we will never need, but there is also no harm in having it, especially as for single core chains there is no downside in omitting it.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,300 @@
...
+[CandidateCommitments](https://github.com/paritytech/polkadot-sdk/blob/b5029eb4fd6c7ffd8164b2fe12b71bad0c59c9f2/polkadot/primitives/src/v7/mod.rs#L682)
+remains unchanged as we will store scale encoded `UMPSignal` messages directly in the parachain
+UMP queue by outputting them in [upward_messages](https://github.com/paritytech/polkadot-sdk/blob/b5029eb4fd6c7ffd8164b2fe12b71bad0c59c9f2/polkadot/primitives/src/v7/mod.rs#L684).
+
+The UMP queue layout is changed to allow the relay chain to receive both the XCM messages and
+`UMPSignal` messages. An empty message (empty `Vec<u8>`) is used to mark the end of XCM messages and
+the start of `UMPSignal` messages.
+
+This way of representing the new messages has been chosen over introducing an enum wrapper to
+minimize breaking changes of XCM message decoding in tools like Subscan for example.
```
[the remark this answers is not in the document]
sandreim: This doesn't say things don't break, only that the impact of breakage is smaller compared to the alternative.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,300 @@
...
+    validation_code_hash: ValidationCodeHash,
+}
+```
+
+In future format versions, parts of the `reserved1` and `reserved2` bytes can be used to include
+additional information in the descriptor.
+
+### Parachain block validation
+
+If the candidate descriptor is version 1, there are no changes.
```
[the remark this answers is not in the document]
sandreim: Before this I was using `old` and `new` which actually might better consider that now we also have an internal version field. I will try to make this more clear.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,300 @@
...
+be able to detect the version of a given candidate receipt.
+
+The version of the descriptor is detected by checking the reserved fields.
+If they are not zeroed, it means it is a version 1 descriptor. Otherwise the `version` field
+is used further to determine the version. It should be `0` for version 2 descriptors. If it is not
+the descriptor has an unknown version and should be considered invalid.
+
+## Drawbacks
+
+The only drawback is that further additions to the descriptor are limited to the amount of
```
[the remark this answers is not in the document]
sandreim: Yes, this is true, but I this assumes all future changes we make to the descriptor are backward compatible to not break other parts of the stack.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,300 @@
...
+There is no impact on privacy.
+
+## Performance
+
+Overall performance will be improved by not checking the collator signatures in runtime and nodes.
+The impact on the UMP queue and candidate receipt processing is negligible.
+
+The `ClaimQueueOffset` along with the relay parent choice allows parachains to optimize their
+block production for either throughput or latency. A value of `0` with the newest relay parent
+provides the best latency while picking older relay parents avoids re-orgs.
```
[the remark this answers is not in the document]
sandreim: Yes, it's about XCM latency which is important. I would just mention this clearly and not remove the paragraph.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,300 @@
...
+It is mandatory for elastic parachains to switch to the new receipt format. It is optional but
+desired that all parachains switch to the new receipts for providing the session index for
+disputes.
+
+The implementation of this RFC itself must not introduce any breaking changes for the parachain
+runtime or collator nodes.
+
+## Compatibility
+
+The proposed changes are backward compatible in general, but additional care must be taken by
```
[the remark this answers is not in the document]
sandreim: I think that's the only case when we are not actually backward compatible. I will update this to reflect it clearly.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,300 @@
...
+
+### UMP transport
+
+[CandidateCommitments](https://github.com/paritytech/polkadot-sdk/blob/b5029eb4fd6c7ffd8164b2fe12b71bad0c59c9f2/polkadot/primitives/src/v7/mod.rs#L682)
+remains unchanged as we will store scale encoded `UMPSignal` messages directly in the parachain
+UMP queue by outputting them in [upward_messages](https://github.com/paritytech/polkadot-sdk/blob/b5029eb4fd6c7ffd8164b2fe12b71bad0c59c9f2/polkadot/primitives/src/v7/mod.rs#L684).
+
+The UMP queue layout is changed to allow the relay chain to receive both the XCM messages and
+`UMPSignal` messages. An empty message (empty `Vec<u8>`) is used to mark the end of XCM messages and
+the start of `UMPSignal` messages.
```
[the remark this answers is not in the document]
sandreim: Making it optional adds some complexity to the implementation while I don't see what are the benefits. This RFC should be a no-op in terms of development for parachain teams. They get it for free when their runtime and node are upgraded but at their own pace as we still support the old descriptors.

text/0103-introduce-core-index-commitment.md
```diff
@@ -0,0 +1,302 @@
...
+
+### UMP transport
+
+[CandidateCommitments](https://github.com/paritytech/polkadot-sdk/blob/b5029eb4fd6c7ffd8164b2fe12b71bad0c59c9f2/polkadot/primitives/src/v7/mod.rs#L682)
+remains unchanged as we will store scale encoded `UMPSignal` messages directly in the parachain
+UMP queue by outputting them in [upward_messages](https://github.com/paritytech/polkadot-sdk/blob/b5029eb4fd6c7ffd8164b2fe12b71bad0c59c9f2/polkadot/primitives/src/v7/mod.rs#L684).
+
+The UMP queue layout is changed to allow the relay chain to receive both the XCM messages and
+`UMPSignal` messages. An empty message (empty `Vec<u8>`) is used to mark the end of XCM messages and
+the start of `UMPSignal` messages.The `UMPSignal` is optional and can be omitted by parachains
```
bkchr, line 72:
```suggestion
the start of `UMPSignal` messages. The `UMPSignal` is optional and can be omitted by parachains
```

bkchr (approved):

sandreim:
/rfc process 0xdcbb1a70e58737edfbfdb0b866cf977bebafcea08479808340ae03e492922b3e

PR #105, opened 2024-07-23
[XCM]: Improved fee mechanism

franciscoaguirre:
Moved from https://github.com/polkadot-fellows/xcm-format/pull/53.

The idea is to extend the current XCM fee mechanism to support in principle delivery fees, but potentially other types of fees as well.
This is accomplished by having two additions:
- `PayFees` instruction
- A `fees` register, where fees go and stay there until the end of execution. Crucially, separate from the `holding` register

### DIFF
--- a/text/0105-xcm-improved-fee-mechanism.md
+++ b/text/0105-xcm-improved-fee-mechanism.md
@@ -104,3 +104,9 @@ None
 ## Future Directions and Related Material
 
 This proposal would greatly benefit from an improved asset trapping system.
+
+[CustomAssetClaimer](https://github.com/polkadot-fellows/xcm-format/blob/master/proposals/0037-custom-asset-claimer.md) is also related, as it directly improves the ergonomics of this proposal.
+
+[LeftoverAssetsDestination](https://github.com/polkadot-fellows/RFCs/pull/107) execution hint would also similarly improve the ergonomics.
+
+[Removal of JIT fees](https://github.com/polkadot-fellows/RFCs/pull/106/files) is also related, they are useless with this proposal.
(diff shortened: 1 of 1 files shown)


text/0105-xcm-improved-fee-mechanism.md
```diff
@@ -0,0 +1,106 @@
...
+The new proposed instruction, `PayFees`, doesn't return the leftover immediately, it keeps it in the `fees` register.
+In order for the proposal to be backwards compatible, `BuyExecution` needs to be slowly rolled out in favour of `PayFees`.
+
+## Prior Art and References
+
+The closed RFC PR on the xcm-format repository, before XCM RFCs got moved to fellowship RFCs: https://github.com/polkadot-fellows/xcm-format/pull/53.
+
+## Unresolved Questions
+
+How to improve the asset trapping mechanism?
```
acatangiu, line 102: I would suggest removing this, it's a more general XCM problem that needs to be tackled independently.

```suggestion
None
```

text/0105-xcm-improved-fee-mechanism.md
```diff
@@ -0,0 +1,112 @@
...
+
+```rust
+PayFees { asset: Asset }
+```
+
+This new instruction will reserve **the entirety** of `asset` for fee payment.
+The asset can't be used for anything else during the entirety of the program.
+This is different from the current semantics of `BuyExecution`.
+
+If not all `Asset` in the `fees` register is used when the execution ends, then we trap them alongside any possible leftover assets from the holding register.
```
xlc, line 49: do we need to mention that `RefundSurplus` can be used to move asset from the fees register to holding register? and programs are expected to use RefundSurplus to avoid trapped assets

acatangiu:
/rfc process 0x39efd93063aeaef13afdb2135a775fec4d87fe3ed51e9bf122bc73a336acd509

PR #107, opened 2024-07-23
[XCM]: Execution hints mechanism

franciscoaguirre:
Moved from https://github.com/polkadot-fellows/xcm-format/pull/59.

This RFC proposes a framework for specifying "execution hints", much like [window hints in GLFW](https://www.glfw.org/docs/3.3/window_guide.html#window_hints).
The first one is `AssetClaimer`, that allows the user to specify who can claim trapped assets resulting from that program execution. That was an instruction already approved and targeted at XCMv5. This RFC proposes to move it to this hints mechanism.

Another potential hint is `LeftoverAssetsDestination`, which, when set, would send all leftover assets to a particular location instead of trapping. This additional hint is not covered by the RFC. It could be proposed in a new one.


acatangiu (approved):

text/0107-xcm-execution-hints.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Stakeholders
+
+- Runtime devs
+- Wallets
+- dApps
+
+## Explanation
+
+A new instruction, `SetExecutionHints`, will be added.
```
vgeddes, line 31: This may be subjective, though I think `SetHints` sounds a bit more concise, without sacrificing meaning.

franciscoaguirre: Sounds good

franciscoaguirre:
Jumped the gun before. Have to generate hash again.

bkchr (approved):

acatangiu:
/rfc process 0x3327c42aa87e7ab3812ffd18a043ae26627bab2d6a9be7e62cff472568d103ef

PR #114, opened 2024-08-17
Adjust Tipper Track Confirmation Periods

Leemo94:
This RFC proposes to change the confirmation periods of the tipper tracks:

- Small Tipper: 10 Minutes -> 12 Hours
-  Big Tipper: 1 Hour -> 1 Day

### DIFF
--- a/text/RFC-114 Adjust Tipper Track Confirmation Periods.md
+++ b/text/RFC-114 Adjust Tipper Track Confirmation Periods.md
@@ -1,4 +1,4 @@
-# RFC-0000: Adjust Tipper Track Confirmation Periods
+# RFC-114: Adjust Tipper Track Confirmation Periods
 
 |                 |                                                                                             |
 | --------------- | ------------------------------------------------------------------------------------------- |
--- a/text/RFC-114 Adjust Tipper Track Confirmation Periods.md
+++ b/text/RFC-114 Adjust Tipper Track Confirmation Periods.md
@@ -3,8 +3,8 @@
 |                 |                                                                                             |
 | --------------- | ------------------------------------------------------------------------------------------- |
 | **Start Date**  | 17-Aug-24                                                                                   |
-| **Description** | Big and Small Tipper Track Conformation Period Modification                                                                                                    |
-| **Authors**     |           Leemo / ChaosDAO                                                                                                        |
+| **Description** | Big and Small Tipper Track Conformation Period Modification                                 |
+| **Authors**     | Leemo / ChaosDAO                                                                            |
 
 ## Summary
 
--- a/text/RFC-XXXX Adjust Tipper Track Confirmation Periods.md
+++ b/text/RFC-XXXX Adjust Tipper Track Confirmation Periods.md
@@ -1,0 +1,87 @@
+# RFC-0000: Adjust Tipper Track Confirmation Periods
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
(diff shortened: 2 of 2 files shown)


ltfschoen:
Increasing the confirmation time on these tipper tracks would allow solo whales and whales in collectives where if they have already exhausted all the conviction of their voting power against a proposal, they'd now have more time to get a temporary on-chain loan (possibly shorting the market to offset losses due to price volatility) against their whale collateral that they probably still have, so they can use the borrowed funds, and temporarily staking those borrowed tokens in staking pools (if rules change so tokens in staking pools are allowed to vote in future) with low conviction (so they can unstake quickly and pay it back quickly after sniping to minimise the risk of losses against it due to price volatility) to negatively snipe referenda. 

If someone who is available during the confirmation period didn't have enough to negatively snipe it enough, they'd now have more time to ask around to find someone in their inner circle who does.

this wouldn't give new and less affluent proposers (little guys) in the ecosystem a chance against malicious and highly opinionated whales who may do whatever it takes to shut down a proposal that they don't like for whatever reason, unless they managed to get significant delegated voting power or somehow got lucky from rubbing shoulders with a whale who sympathise and backed them at a whim. 

Those looking for tips, especially small tips are most likely not whales (just filling out the form asking for the tip might cost more than they'd gain from the tip) unless they have an ulterior motive like asking for a tip in ded tokens just to make the community think they should fomo into buying some, so they wouldn't have the collateral to borrow to compete and defend themselves against whales that have more borrowing power.

I am opposed to the change as it stands, particularly for the small tipper track.

In fact, i think that the duration of the confirmation period for these tracks should be randomised, unknown at any moment, and unpredictable, to minimise the likelihood of whales being able to know and plan precisely how long they need to borrow with high leverage with minimal risk so they can destroy the hopes of small players during the confirmation period that are asking for reasonable tips but that the whales oppose.

ltfschoen:
In fact there shouldn't be any predictable confirmation periods since they just encourage potential:
- malicious use of prediction markets coupled with malicious use of OpenGov
- high leverage trading to get large temporary voting power at the last moment of confirmation periods to swing votes
- collusion to trigger black swan-like events that wipe out highly leveraged traders at the last predictable moment of confirmation periods
- insider traders that encourage use of private chat rooms or forum groups just so they're first to get intel about the feasibility of realising deliverables achieving deadlines from core devs and can front-run public announcements that may be price sensitive and affect prediction market outcomes

ltfschoen:
Could the source of the randomness that dictates the random confirmation period duration that i mentioned earlier be that which is mentioned in this proposal https://kusama.polkassembly.io/referenda/442

text/RFC-114 Adjust Tipper Track Confirmation Periods.md
```diff
@@ -0,0 +1,87 @@
...
+- Polkassembly - directly uses on-chain data via rpc to fetch trackInfo so the change will be automatically reflected.
+- SubSquare - scan script will update their app to the latest parameters and it will be automatically reflected in their app.
+
+## Prior Art and References
+
+N/A
+
+## Unresolved Questions
+
+Some token holders may want these confirmation periods to remain as they are currently and for them not to increase. If this is something that the Polkadot Technical Fellowship considers to be an issue to implement into a runtime upgrade then I can create a Wish For Change to obtain token holder approval.
```
ggwpez, line 83: Yea could be reasonable. Otherwise this puts the Fellowship in a spot where they have to make a decision on a not purely technical matter that may be better decided by OpenGov.

PR #117, opened 2024-08-24
The Unbrick Collective

pandres95:
Related to https://github.com/polkadot-fellows/runtimes/issues/376, this RFC defines the process to define and implement the collective, and the mechanisms needed to do so.

### DIFF
--- a/text/0000-unbrick-collective.md
+++ b/text/0000-unbrick-collective.md
@@ -1,0 +1,172 @@
+# RFC-0000: The Unbrick Collective
+
+|                 |                                                                                          |
+| --------------- | ---------------------------------------------------------------------------------------- |
+| **Start Date**  | 22 August 2024                                                                           |
+| **Description** | The Unbrick Collective aims to help teams rescuing a para once it stops producing blocks |
+| **Authors**     | Bryan Chen, Pablo Dorado                                                                 |
+
+## Summary
+
+A followup of the [RFC-0014]. This RFC proposes adding a new collective to the Polkadot Collectives
+Chain: The Unbrick Collective, as well as improvements in the mechanisms that will allow teams
+operating paras that had stopped producing blocks to be assisted, in order to restore the production
+of blocks of these paras.
+
+## Motivation
+
+Since the initial launch of Polkadot parachains, there has been many incidients causing parachains
+to stop producing new blocks (therefore, being _bricked_) and many occurrences that requires
+Polkadot governance to update the parachain head state/wasm. This can be due to many reasons range
+from incorrectly registering the initial head state, inability to use sudo key, bad runtime
+migration, bad weight configuration, and bugs in the development of the Polkadot SDK.
+
+Currently, when the para is not unlocked in the _paras registrar_[^1], the `Root` origin is required to
+perform such actions, involving the governance process to invoke this origin, which can be very
+resource expensive for the teams. The long voting and enactment times also could result significant
+damage to the parachain and users.
+
+Finally, other instances of governance that might enact a call using the `Root` origin (like the
+Polkadot Fellowship), due to the nature of their mission, are not fit to carry these kind of tasks.
+
+In consequence, the idea of a Unbrick Collective that can provide assistance to para teams when
+they brick and further protection against future halts is reasonable enough.
+
+## Stakeholders
+
+- Parachain teams
+- Parachain users
+- OpenGov users
+- Polkadot Fellowship
+
+## Explanation
+
+### The Collective
+
+The Unbrick Collective is defined as an unranked collective of members, not paid by the Polkadot
+Treasury. Its main goal is to serve as a point of contact and assistance for enacting the actions
+needed to unbrick a para. Such actions are:
+
+- Updating the Parachain Verification Function (a.k.a. a new WASM) of a para.
+- Updating the head state of a para.
+- A combination of the above.
+
+In order to ensure these changes are safe enough for the network, actions enacted by the Unbrick
+Collective must be whitelisted via similar mechanisms followed by collectives like the Polkadot
+Fellowship. This will prevent unintended, not overseen changes on other paras to occur.
+
+Also, teams might opt-in to delegate handling their para in the registry to the Collective. This
+allows to perform similar actions using the _paras registrar_, allowing for a shorter path to unbrick a
+para.
+
+### The Unbrick Process
+
+```mermaid
+flowchart TD
+    A[Start] 
+    
+    A -- Bricked --> C[Request Unbrick «via governance»]
+    C --> D[unbrick call proposal on WhitelistedUnbrickCaller]
+    C --> E[whitelist call proposal on the Unbrick governance]
+    E -- call whitelisted --> F[unbrick call enacted]
+    D -- unbrick called --> F
+    F --> Y
+
+    A -- Not bricked --> O[Opt-in to the Collective]
+    O -- Bricked --> P[Collective calls registrar]
+    P --> Y
+    
+    Y[update PVF / head state] -- Unbricked --> Z[End]
+```
+
+Initially, a para team has two paths to handle a potential unbrick of their para in the case it
+stops producing blocks.
+
+1. **Opt-in to the Unbrick Collective**: This is done by delegating the handling of the para
+  in the _paras registrar_ to the Collective. This doesn't require unlocking the para. This way,
+  the collective is enabled to perform changes in the _paras registrar_ without the need for
+  whitelisting.
+2. **Request an Unbrick Process**: In case the para hasn't delegated its handling in the _paras
+  registrar_, it'll be still possible for the para team to submit a proposal to unbrick the para,
+  assisted by the Collective. This process is expected to be more expedite (and less expensive)
+  for a team to perform than submitting a proposal on the `Root` governance track.
+
+### Belonging to the Collective
+
+The collective will be initially created without members (no seeding). There will be additional
+governance proposals to setup the seed members.
+
+The origins able to modify the members of the collective are:
+
+- The `Fellows` track in the Polkadot Fellowship.
+- `Root` track in the Relay.
+- More than two thrids of the existing Unbrick Collective.
+
+The members are responsible to verify the technical details of the unbrick requests (i.e. the hash
+of the new PVF being set). Therefore, they must have the technical capacity to perform such tasks.
+
+Suggested requirements to become a member are the following:
+
+- Rank 3 or above in the Polkadot Fellowship.
+- Being a CTO or Technical Lead in a para team that has opted-in to delegate the Unbrick Collective
+  to manage the PVF/head state of the para.
+
+## Drawbacks
+
+The ability to modify the Head State and/or the PVF of a para means a possibility to perform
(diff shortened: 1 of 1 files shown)


burdges:
An unbricking call could also issue assets of course.  A parachain could turstile foreign assets, so that unbricking could only issue native assets, and steal wrapped foreign assets, but not issue foreign assets.

Yet, we've long envisioned synchronous privliged library calls, initially the SPREE idea by Gav or whoever, seemingly now folded into JAM.  All these proposals exist firstly so that assets can be native on multiple parachains, so then unbricking could issue foreign assets in theory.

We've no alternatives of course, since staying bricked maybe worse, but unbricking could demand audits based upon usage of SPREE-like or JAM functionality.  It's not necessarily assets either, so in general pallets or whatever should document their security concerns, and unbricking should review these.  You'd expect mostly they say nothing, but in general you would not be permitted to say roll back your own DOT SPREE state to a known good state.

A parachain could've a maximum block gap configuration so the parachain itself could respond to the unbricking calls.  If a parachains wants one block every six seconds, then maybe they want to unbrick within minutes.  If a parachain makes one block every 24 hours, then maybe they require days.

xlc:
Maybe worthwhile to mention that:

Before this change, a parachain will be one of two state: unlocked (para manager essentially have sudo permission), or locked (para manager don't have any permission).

After this change, a parachain will be one of three state: unlocked, locked, and managed by Unbrick Collective. The new state sets between fully centralized (controlled by one account or wasm), fully decentralized (only controlled by the wasm). The chain is controlled by the wasm or the Unbrick Collective + OpenGov. The goal is trying to find the right balance between fully centralized (bad) and fully decentralized (inefficient).

Another thing is that on how to determine if a parachain is bricked. With the old parachain slot model, it is relatively easy that we can assume if no block is produced for sufficiently long time, there must be something wrong. However this heuristic no long works with core time model. So maybe we need a mechanism like @burdges suggested that when a parachain opt-in, it should also set how many blocks are required to consider the parachain is bricked.

burdges:
We'd maybe want declared estimates for how often parachains make blocks elsewhere, like in some flavours of off-chain messaging.  If so, we might make this number mandatory even for parachains who do not join the unbrick colelctive, or maybe that's a slightly different number, but regardless that's future work.

---

I'll clarify my comments above:

In future, we'll have functionality that's more sensitive than a parachains own code, so an unlocked/sudo parachain could use such functionalities, but could not change the code or state of that functionality.  The unbrick collective might be similarly limited, or maybe not, depending upon how its ellections work in practice.

Initially, I'd suggest the unbrick collective be limited to what sudo does, meaning it cannot touch SPREE-like code or state.  We'd consider strengthening the unbrick colelctive if the ellections are polkadot wide and seems strong like our original faster pre-opengov governance (unlikely).  We've no difference anyways right now, since we do not yet have SPREE, JAM, etc, so this could be discussed once those come into being.

In other words, my SPREE comments do not propose changing this RFC per se, except maybe notes in the seeding section and/or Unresolved Questions.

Auditing requirements provides another option, which slows everything down, but nothing like the opengov slowdown.  At the extreme, auditors could be selected at random after the new code & state were proposed, almost like running the [machine elves protocol](https://eprint.iacr.org/2024/961.pdf) or the sampling beefy off-chain among humans, but that's likely too complex.  It's also possible parachains themselves would want auditing before unbricking, beyond merely the delay.

Auditing could be defined by some future RFCs which would maybe take things in other directions, like maybe parachains want auditing for regular upgrades, etc.  In other words, my auditing comments do not propose changing this RFC per se, except maybe notes in Unresolved Questions.

text/0117-unbrick-collective.md
```diff
@@ -0,0 +1,172 @@
...
+from incorrectly registering the initial head state, inability to use sudo key, bad runtime
+migration, bad weight configuration, and bugs in the development of the Polkadot SDK.
+
+Currently, when the para is not unlocked in the _paras registrar_[^1], the `Root` origin is required to
+perform such actions, involving the governance process to invoke this origin, which can be very
+resource expensive for the teams. The long voting and enactment times also could result significant
+damage to the parachain and users.
+
+Finally, other instances of governance that might enact a call using the `Root` origin (like the
+Polkadot Fellowship), due to the nature of their mission, are not fit to carry these kind of tasks.
```
burdges, line 30: This sentence is too hard to parse.  It's they're too busy, right?

text/0117-unbrick-collective.md
```diff
@@ -0,0 +1,172 @@
...
+
+In consequence, the idea of a Unbrick Collective that can provide assistance to para teams when
+they brick and further protection against future halts is reasonable enough.
```
burdges, line 33:
```
Initially, the unbrick collective has powers similar to a parachains own sudo, but permits more decentralized control.  In future, polkadot shall provide functionality like SPREE or JAM that exceeds sudo permissions, so the unbrick collective cannot modify their core or state roots.
```

text/0117-unbrick-collective.md
```diff
@@ -0,0 +1,172 @@
...
+## Unresolved Questions
+
+- What are the parameters for the `WhitelistedUnbrickCaller` track?
+- Any other methods that shall be updated to accept `Unbrick` origin?
+- Any other requirements to become a member?
+- We would like to keep this simple, so no funding support from the Polkadot treasury. But do we
+  want to compensate the members somehow? i.e. Allow parachain teams to donate to the collective
+- Do we want to have this collective offer additional technical support to help bricked parachains?
+  i.e. help debug the code, create the rescue plan, create postmortem report, provide resources on
+  how to avoid getting bricked
```
burdges, line 161:
```
- We hope SPREE/JAM would be carefully audited for miss-use risks before being
   provided to parachain teams, but could the unbrick collective have an elections
   that warranted trust beyond sudo powers? 
- An auditing framework/collective makes sense parachain code upgrades, but
   could also strengthen the unbrick collective.
```

text/0117-unbrick-collective.md
pandres95: Not necessarily. It's just a matter of mission and scope. 😅

Same case with the Unbrick Collective: it's scope would be providing assistance to para teams which need help unbricking their para, not helping teams design their newest runtime version, or auditing code (in which case, your suggestion of an auditing collective sounds great).

text/0117-unbrick-collective.md
```diff
@@ -0,0 +1,184 @@
...
+### Belonging to the Collective
+
+The collective will be initially created without members (no seeding). There will be additional
+governance proposals to setup the seed members.
+
+The origins able to modify the members of the collective are:
+
+- The `Fellows` track in the Polkadot Fellowship.
+- `Root` track in the Relay.
+- More than two thrids of the existing Unbrick Collective.
```
bkchr, line 110:
```suggestion
- More than two thirds of the existing Unbrick Collective.
```

text/0117-unbrick-collective.md
```diff
@@ -0,0 +1,184 @@
...
+- [Unbrick Collective, Polkadot Forum][forum:6931]
+
+## Unresolved Questions
+
+- What are the parameters for the `WhitelistedUnbrickCaller` track?
+- Any other methods that shall be updated to accept `Unbrick` origin?
+- Any other requirements to become a member?
+- We would like to keep this simple, so no funding support from the Polkadot treasury. But do we
+  want to compensate the members somehow? i.e. Allow parachain teams to donate to the collective.
+- We hope SPREE/JAM would be carefully audited for miss-use risks before being
```
bkchr, line 166: I don't get what you want to say with this point.

text/0117-unbrick-collective.md
```diff
@@ -0,0 +1,184 @@
...
+- [RFC-0014: Improve Locking Mechanisms for Parachains][RFC-0014]
+- [How to Recover a Parachain, Polkadot Forum][forum:673]
+- [Unbrick Collective, Polkadot Forum][forum:6931]
+
+## Unresolved Questions
+
+- What are the parameters for the `WhitelistedUnbrickCaller` track?
+- Any other methods that shall be updated to accept `Unbrick` origin?
+- Any other requirements to become a member?
+- We would like to keep this simple, so no funding support from the Polkadot treasury. But do we
```
bkchr, line 164: IMO most of the members should be from different parachain teams. This way it is a quid-pro-quo in longterm.

bkchr: But depending on their involvement they could maybe get retroactive payment for helping with a certain fix.

text/0117-unbrick-collective.md
```diff
@@ -0,0 +1,184 @@
...
+In order to ensure these changes are safe enough for the network, actions enacted by the Unbrick
+Collective must be whitelisted via similar mechanisms followed by collectives like the Polkadot
+Fellowship. This will prevent unintended, not overseen changes on other paras to occur.
+
+Also, teams might opt-in to delegate handling their para in the registry to the Collective. This
+allows to perform similar actions using the _paras registrar_, allowing for a shorter path to unbrick a
+para.
+
+Initially, the unbrick collective has powers similar to a parachains own sudo, but permits more
+decentralized control. In the future, Polkadot shall provide functionality like SPREE or JAM that
```
bkchr, line 63: I don't get the JAM part. Even in JAM this could be build into the parachain service to be have the unbrick collective.

text/0117-unbrick-collective.md
```diff
@@ -0,0 +1,184 @@
...
+exceeds sudo permissions, so the unbrick collective cannot modify those state roots or code.  
+
+### The Unbrick Process
```
bkchr, line 66: IMO this should still require that the chain is actually bricked, aka not producing blocks. When the chain enables the unbrick collective it could for example pass a time out after which the chain is seen as "bricked".

bkchr (commented):
Generally looking good. Left some ideas around improvements/questions.

burdges:
Indirectly related but..  https://github.com/paritytech/polkadot-sdk/issues/5588

bkontur:
> Indirectly related but.. [paritytech/polkadot-sdk#5588](https://github.com/paritytech/polkadot-sdk/issues/5588)

Yes, exactly - this is closely related to the D-Day governance issue, which is essentially addressing a similar problem.
I wrote a draft design document for D-Day [here](https://hackmd.io/@bkontur/SJ07ySIlgx), but after last week’s AHM retreat, many questions have been clarified. I’ll now prepare a new, more "final" version of the D-Day proposal.

---

But for sure, there are some common features needed, for example:

> Updating the Parachain Verification Function (a.k.a. a new WASM) of a para.

IIUC, this proposes being whitelisted/called from the Collectives chains to the Relay Chain.
I saw the [PoC](https://github.com/paritytech/polkadot-sdk/pull/5429) for `fn unbrick` -
```
pub fn unbrick(para: ParaId, maybe_new_code: Option<ValidationCode>, maybe_new_head: Option<HeadData>)
```
And I am sure that this won't work, because we cannot send `maybe_new_code: Option<ValidationCode>` the whole WASM over `UMP` from para to relay.

Exactly for this case, I've implemented this feature: https://github.com/paritytech/polkadot-sdk/pull/7592

---

Overall, I’ll prepare a new, more "final" version of the D-Day governance proposal.
After that, I’ll return to this RFC and can help move it forward.

PR #120, opened 2024-09-17
Referenda Confirmation by Candle Mechanism

pandres95:
[Rendered document](https://github.com/polkadot-fellows/RFCs/blob/35f24b060f6fcd7c1aec010f068c8a017ae9a915/text/0120-referenda-confirmation-by-candle-mechanism.md)

### DIFF
--- a/text/00xx-referenda-confirmation-by-candle-auction.md
+++ b/text/00xx-referenda-confirmation-by-candle-auction.md
@@ -1,0 +1,98 @@
+# RFC-00xx: Referenda Confirmation by Candle Auction
+
+|                 |                                                                  |
+| --------------- | ---------------------------------------------------------------- |
+| **Start Date**  | 22 March 2024                                                    |
+| **Description** | Proposal to decide polls after confirm period via candle auction |
+| **Authors**     | Pablo Dorado                                                     |
+
+## Summary
+
+In an attempt to mitigate risks derived from unwanted behaviours around long ongoing periods on
+referenda, this proposal describes how to finalize and decide a result of a poll via a
+mechanism similar to candle auctions.
+
+## Motivation
+
+Referenda protocol provide permissionless and efficient mechanisms to enable governance actors to
+decide the future of the blockchains around Polkadot network. However, they pose a series of risks
+derived from the game theory perspective around these mechanisms. One of them being where an actor
+uses the the public nature of the tally of a poll as a way of determining the best point in time to
+alter a poll in a meaningful way.
+
+While this behaviour is expected based on the current design of the referenda logic, given the
+recent extension of ongoing times (up to 1 month), the incentives for a bad actor to cause losses
+on a proposer, reflected as wasted cost of opportunity increase, and thus, this otherwise
+reasonable outcome becomes an attack vector, a potential risk to mitigate, especially when such
+attack can compromise critical guarantees of the protocol (such as its upgradeability).
+
+To mitigate this, the referenda underlying mechanisms should incentive actors to cast their votes
+on a poll as early as possible. This proposal's approach suggests using a Candle Auction that will
+be determined right after the confirm period finishes, thus decreasing the chances of actors to
+alter the results of a poll on confirming state, and instead incentivizing them to cast their votes
+earlier, on deciding state.
+
+## Stakeholders
+
+- **Governance actors**: Tokenholders and Collectives that vote on polls that have this mechanism
+  enabled should be aware this change affects the outcome of failing a poll on its confirm period.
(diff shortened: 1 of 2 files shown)


laboon:
So this is how things work - please correct me if I'm wrong.

1. A Referendum enters the Confirmation period. It must stay above a certain threshold (support and approval) for a certain period of time. If at any point, it drops below, it is kicked out of Confirming state. This is exactly how things work now.
2. However, there is a second stage _iff_ the Referendum finishes the confirming period. Here, a random block from the Confirmation period is decided upon as the "true" ending point.

What I don't understand is what additional criteria will decide this. If a Ref is < support or approval, it will have already failed. So every block in the confirmation period will, by definition, already have met the criteria to stay in the Confirming period and thus pass. Will there be a higher threshold(s) necessary for this Finalization phase?

shawntabrizi:
The way I view this proposal is as follows:

- this new mechanism only aids in passing proposals that would otherwise fail
- we should not really be creating mechanisms to make things "pass" more easily
- if there are proposals which are on the border (51/49) which this affects, probably they should be nayed, and re-discussed to get more overall agreement from the community
- there is also a non-negligible amount of computational and storage complexity introduced for EACH proposal, which will hopefully be on the order of tens of thousands over just the next couple of years.

There is nothing fundamentally wrong with the idea, but I don't think this is the kind of direction I would want to see OpenGov go toward.

Just my opinion. Thanks for taking the time to write down and present the idea.

burdges:
I agree with @shawntabrizi that too many proposals suck, like advertising, so you might count late no votes, but cut off late yes votes.

I'll also caution that referenda are not auctions, so the voting theory should works out somewhat different.  In particular, there maybe a psycological concerns too if people's votes go uncounted, but which do not occur when a company buys blockspace.

@AlistairStewart saw some issue with this specific proposal, but might've some thoughts on the benefits of candle auctions, so let's see what he says.

AlistairStewart:
So firstly I should explain why we need a change from the status quo.

It seems that many people are prepared to vote and increase their conviction only if their preferred side is losing a referendum. So sometimes if a refendum always has over 50% for yes, these people don't turnout and turnout remains low. Since turnout is often low, referenda don't confirm until near the end of the decision period. AFter the decision period annd in this confirmation period, a big enough no vote can snipe the refendum, making it fail immediately.

Other referenda tend to swing between over 50% for yes and over 50% for no, with turnout increasing theough the decision period. Accounts that don't normally vote with high convication get into a conviction auction, where those who oppose the currently winning side want to get just enough votes for their side to be winning. These referenda have multiple confirmation periods that don't last the full time. If the switching for yes to no happens after the decision period then the referendum just stops.

If we did something like this RFC or alternatives that would take longer to get a decision, then more accounts would vote with a higher conviction than right now and the result would be fairer. I would argue that this is a good thing. That people using accounts that don't normally vote or who are voting with a higher than normal conviction, locking up their tokens for a long time, are probably people that actually care about the issue voted on. 

Using a candle auction is fairer and never gives a hard cutoff, when voting enough right now ensures victory. It need not extend the auction for longer than is possible now. It should result in people voting at about the end of the decision period and give a higher turnout with conviction than we have right now.

AlistairStewart:
Here's how I think this RFC should work:

If at the end of the decision period, if an auction is confirming or has been confirming recebtly (say in the last half a confirming period length), then the auction should enter an *ending period*, which always lasts for length equal to a confirmation period. After that time, voting stops, and the auction waits for the randomness to decide when in the ending period the auction stopped, which determines the winner.

I'm not sure right now how this code actually works or RFC proposes.

burdges:
I've not looked closely at the cirrent scheme, but anecdotally it extends the conformation period when votes change, but avoids allowing unending extensions, so maybe it prevents aye sniping but permits nay snipping.  If it favors nay then that's already reasonable.

At present, there is a psycological strategy where you initially vote opposite your preference, which motivates voters who favor your position, and then you change your vote.  It's possible a candle auction improves this, but probably not, and maybe just removing or limiting conviction helps more.

It's really unclear why the candle mechanism would increase turnout though?  A priori, complex voting schemes might reduce turnout, until the voters pick up more knowlege, not necessarily about the scheme itself, but that turnout was harmless and benefited their interests.

As above, there maybe risks that not counting some late votes harms turnout.  I suspect our current rules for extending conformation period provides much of the benefits of candle auctions, but without having voters votes go uncounted.

We could've a knowledge gathering phase where votes are not final, and conviction is allowed, and then a late voting phase where votes cannot be changed once cast, or can only be changed from aye to nay, and new votes cannot have conviction.

PR #124, opened 2024-10-18
Extrinsic v5 definition and specification

georgepisaltu:


gui1117 (approved):

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,155 @@
...
+3. The result of the hash should then be signed with the signature type specified in the extension definition.
+
+```rust
+// Step 1: encode the bytes
+let encoded = (extension_version_byte, call, transaction_extension, transaction_extension_implicit).encode();
+// Step 2: hash them
+let payload = blake2_256(&encoded[..]);
+// Step 3: sign the payload
+let signature = keyring.sign(&payload[..]);
+```
```
carlosala, line 92: This will most likely depend on which origin and kind of signature is used. For instance, ZK circuits might have different levels of proofs and might not be signing the whole payload. This should go to another RFC, which defines every particular `VerifySignature`-like extension.

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,155 @@
...
+consumers downstream in the transition period between these extrinsic versions, we should:
+
+- Remove the `Signed` variant starting with v5 extrinsics
+- Add the `General` variant starting with v5 extrinsics
+- Enable runtimes to support both v4 and v5 extrinsics
+
+## Drawbacks
+
+The metadata will have to accommodate two distinct extrinsic format versions at a given point in
+time in order to provide the new functionality in a non-breaking way for users and tooling.
```
carlosala, line 106: This is not a drawback IMO. Metadata v15 should show v4 and metadata v16 and ahead have a vector of extrinsic versions.

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,155 @@
...
+
+There is no impact on testing, security or privacy.
+
+## Performance, Ergonomics, and Compatibility
+
+This change makes the authorization through signatures configurable by runtime devs in version 5
+extrinsics, as opposed to version 4 where the signing payload algorithm and signatures were
+hardcoded. This moves the responsibility of ensuring proper authentication through
+`TransactionExtension` to the runtime devs, but a sensible default which closely resembles the
+present day behavior will be provided in `VerifySignature`.
```
carlosala, line 118: This is not true. Signature schemes and addresses were configurable by runtime devs through rust generics. For example, Moonbeam uses only ECDSA signatures with EVM-like addresses.
Besides that, I wouldn't mention `VerifySignature` since it should be RFC-ed anyways.

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Unresolved Questions
+
+There is no clear way to expose two different extrinsic versions in the current metadata framework.
+A non-exhaustive list of options discussed so far:
+
+1. Change the `ExtrinsicMetadata` trait to specify a list of versions instead of a single version.
+2. Use the custom fields in the metadata to specify the details of the version 5.
+3. Create a new trait similar to `ExtrinsicMetadata`, but for future versions of the extrinsic
+   format and add it to the metadata.
```
carlosala, line 149: This is irrelevant for the RFC. Again, it should not include implementation details of the `polkadot-sdk`.

carlosala (changes requested):
Nice starting point! Generally I'd try to keep the RFC as far away from the implementation as possible.
Great work George! 👏🏻

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,155 @@
...
+The introduction of `General` transactions allows the authorization of any and all origins through
+extensions. This means that, with the appropriate extension, `General` transactions are capable of
+replicating the same behavior present day v4 `Signed` transactions. Specifically for Polkadot
+chains, an example implementation for such an extension is
+[`VerifySignature`](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/verify-signature),
+introduced in the Transaction Extension
+[PR3685](https://github.com/paritytech/polkadot-sdk/pull/3685). Other extensions can be inserted
+into the extension pipeline to authorize different custom origins. Therefore, a `Signed` extrinsic
+variant is redundant to a `General` one strictly in terms of functionality available to users and
+would eventually need to be deprecated and removed.
```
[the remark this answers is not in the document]
georgepisaltu: This is in the explanation section and I think it adds valuable context to the discussion. I appreciate that you're familiar with the idea of authorization through transaction extensions, but I think it's still a novel topic and the example is useful. However, if others disagree, I will remove it.

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,155 @@
...
+
+### Encoding format for version 5
+
+As with version 4, the encoded v5 extrinsics will still be an array of SCALE encoded bytes, starting
+with the encoded length of the following bytes. The leading byte will determine the version and type
+of extrinsic, as specified by
+[RFC84](https://github.com/polkadot-fellows/RFCs/blob/main/text/0084-general-transaction-extrinsic-format.md),
+with the addition that the `Signed` variant will not be supported for v5 extrinsics, for reasons
+mentioned above.
```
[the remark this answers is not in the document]
georgepisaltu: Good catch, I rephrased it :+1: .

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,155 @@
...
+leading encoding byte. The extension version byte should be included in payloads to be signed by all
+extensions configured by runtime devs to ensure a user's extension version choice cannot be altered
+by third parties.
+
+After the extension version byte, the extensions will be encoded next, followed by the call itself.
+
+A quick visualization of the encoding:
+
+- `Bare` extrinsics: `(extrinsic_encoded_len, 0b0000_0101, call)`
+- `General` transactions: `(extrinsic_encoded_len, , 0b0100_0101, extension_version_byte, extension, call)`
```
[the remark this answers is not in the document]
georgepisaltu: Without the NOTE I'd have to always mention the `Bare` case in all the following sections.

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,155 @@
...
+Transaction Extension which replicates the current `Signed` transactions in v5 extrinsics, namely
+[`VerifySignature`](https://github.com/paritytech/polkadot-sdk/tree/master/substrate/frame/verify-signature).
+This extension leverages the new inherited implication functionality introduced in
+`TransactionExtension` and creates a payload to be signed using the data of all extensions after
+itself in the extension pipeline. In order to run a transaction with a signed origin, a user must
+create the transaction with an instance of the extension which provides a signature. Alternatively,
+if users want to use some other origin, they should create the transaction with this particular
+extension disabled. More on this behavior in the extension documentation. This extension can be
+configured to accept a `MultiSignature`, which makes it compatible with all signature types
+currently used in Polkadot.
```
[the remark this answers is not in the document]
georgepisaltu: The goal of it is to explain how to use the new v5 format to achieve provide the same functionality as signed transactions. I then use the example in `polkadot-sdk` to show this. The previous phrasing was a bit off and the intent wasn't clear, hopefully now it's better.

I don't think it's wrong to have links to `polkadot-sdk` or reference stuff from there. `polkadot-sdk` is used right now in Polkadot and Kusama runtimes, and quite extensively I might add. Even if we might not want to use it to define future functionality, it's inevitable that describing current functionality will point to `polkadot-sdk`. The upstream/consumer dependency is already there, I think making the RFC overly general by avoiding specific examples doesn't help in any way, it just makes it harder to read and understand. This does not in any way mean that the example pulled from `polkadot-sdk` is the only accepted implementation of specified functionality, it just helps me not unfurl a lot of code in this file.

That said, if others consider the examples are not useful, I will remove them.

text/0124-extrinsic-version-5.md
georgepisaltu: Yes, this is just an example on how to achieve `Signed` functionality in v5. I agree that the "chosen" way of providing this functionality in the actual runtimes should be a different RFC, dependent on this one. 

However, there should be only one `VerifySignature`-like extension used to replicate this signed behavior. There is no need to have multiple extensions for this purpose. For custom origins, users can create infinite variations of an authorization extension and that doesn't need an RFC.

text/0124-extrinsic-version-5.md
georgepisaltu: It is something extra to support in the metadata for both the runtime  and users, is this not a drawback?

text/0124-extrinsic-version-5.md
georgepisaltu: Your statement about configurable signature schemes is true. However, I still consider my statement to be true because:
- The signing payload generation algorithm was hardcoded; this is not the case anymore as any extension can take the inherited implication and add or subtract any data to it and mutate it in any way (such as hashing it - or not) before actually creating a signature.
- There are now multiple ways of ending up with a `Signed` origin variant, with arbitrary logic in any `TransactionExtension` being able to authorize that origin; before, a user HAD to provide a transaction signed by a specific account.

All of this static logic is now moved to extensions. The extensions receive the inherited implication, the generation of which is still hardcoded and handled in this RFC, but is not in any way mandatory to be used in any signing scheme.

I'd agree though that the phrasing isn't clear, but I'm not sure how to improve it.

text/0124-extrinsic-version-5.md
georgepisaltu: Reading it again, I think you're right and it makes more sense to remove it.

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,156 @@
...
+present day behavior will be provided in `VerifySignature`.
+
+### Performance
+
+There is no performance impact.
+
+### Ergonomics
+
+Tooling will have to adapt to be able to tell which authorization scheme is used by a particular
+transaction by decoding the extension and checking which particular `TransactionExtension` in the pipeline is enabled to do the origin authorization. Previously, this was done by simply checking whether the transaction is signed or unsigned, as there was only one method of authentication.
```
acatangiu, line 134: nit: this is the only unbroken line in the doc, suggest breaking it at similar length limit as the others

```suggestion
transaction by decoding the extension and checking which particular `TransactionExtension` in the pipeline
is enabled to do the origin authorization. Previously, this was done by simply checking whether
the transaction is signed or unsigned, as there was only one method of authentication.
```

(actually Line 15 - the paragraph under "Motivation" - is also one long unbroken line)

acatangiu (approved):
Looks good! 👍

text/0124-extrinsic-version-5.md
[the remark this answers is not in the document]
carlosala: Yeah, that was my idea. Just explain it without the note.

text/0124-extrinsic-version-5.md
carlosala: Yep, but anyways this should go on a different RFC. I agree that `VerifySignature`-extension requires its own RFC, yes, but this has nothing to do with extrnsics v5 RFC.

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,156 @@
...
+starting with the encoded length of the following bytes in compact format. The leading byte after
+the length determines the version and type of extrinsic, as specified by
+[RFC84](https://github.com/polkadot-fellows/RFCs/blob/main/text/0084-general-transaction-extrinsic-format.md).
+For reasons mentioned above, this RFC removes the `Signed` variant for v5 extrinsics.
+
+NOTE: For `Bare` extrinsics, the following bytes will just be the encoded call and nothing else.
+
+For `General` transactions, as stated in
+[RFC99](https://github.com/polkadot-fellows/RFCs/blob/main/text/0099-transaction-extension-version.md),
+an extension version byte must be added in the next extrinsic version. This byte should allow
```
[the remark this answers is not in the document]
georgepisaltu: Rephrased, it was meant to say "when we bump the format version, this needs to be in that new format". It's redundant indeed, so I removed that.

text/0124-extrinsic-version-5.md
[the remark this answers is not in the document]
georgepisaltu: Done.

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,156 @@
...
+leading encoding byte. The extension version byte should be included in payloads to be signed by all
+extensions configured by runtime devs to ensure a user's extension version choice cannot be altered
+by third parties.
+
+After the extension version byte, the extensions will be encoded next, followed by the call itself.
+
+A quick visualization of the encoding:
+
+- `Bare` extrinsics: `(extrinsic_encoded_len, 0b0000_0101, call)`
+- `General` transactions: `(extrinsic_encoded_len, , 0b0100_0101, extension_version_byte, extension, call)`
```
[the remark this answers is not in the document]
georgepisaltu: The extension itself is usually a tuple of multiple extensions, generally referred to as the extension pipeline. Technically it's only one extension, the `TxExtension` commonly defined in runtimes, but that is always a tuple of extensions like `CheckNonce`, `CheckWeight`, `ChargeTransactionPayment` etc., so it would be only one extension version, as it is the version of the tuple, but there are multiple extension**s** in the pipeline.

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Signatures on Polkadot in General transactions
+
+In order to run a transaction with a signed origin in extrinsic version 5, a user must create the
+transaction with an instance of at least one extension responsible for authorizing `Signed` origins
+with a provided signature. Alternatively, if users want to use some other origin, they should create
+the transaction with this particular extension disabled.
```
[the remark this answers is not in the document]
georgepisaltu: To your question, yes it is entirely possible and, in fact, what I expect to be implemented in most runtimes. The "alternative" wasn't helpful so I removed it.

text/0124-extrinsic-version-5.md
```diff
@@ -0,0 +1,156 @@
...
+I will use this extension as an example on how to replicate current `Signed` transaction
+functionality in the new v5 extrinsic format, though the runtime logic is not constrained to this
+particular implementation.
+
+This extension leverages the new inherited implication functionality introduced in
+`TransactionExtension` and creates a payload to be signed using the data of all extensions after
+itself in the extension pipeline. This extension can be configured to accept a `MultiSignature`,
+which makes it compatible with all signature types currently used in Polkadot.
+
+In the context of using and extension such as `VerifySignature` to replicate current `Signed`
```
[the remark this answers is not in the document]
georgepisaltu: Fixed.

text/0124-extrinsic-version-5.md
georgepisaltu: Added the explanation.

text/0124-extrinsic-version-5.md
georgepisaltu: Formatted all the lines.

text/0124-extrinsic-version-5.md
georgepisaltu: Again, `VerifySignature` is just an example, and also very clearly labeled as an example in the latest revision of the RFC. Because of the explanation above and the fact that it's not obvious from the start what the extension pipeline should look like for v5 extrinsics, I think the example is useful.

acatangiu: I think the example is useful, and since it's also explicitly labeled as an example I see no reason to remove it

PR #126, opened 2024-10-25
Introduce PVQ

indirection42:
Looking for feedback.

## Summary
This proposal introduces PVQ (PolkaVM Query), which aims to serve as an intermediary layer between different chain runtime implementations and tools/UIs, to provide a unified interface for cross-chain queries.

## Related Discussions
https://forum.polkadot.network/t/wasm-view-functions/1045

## PoC implementations
https://github.com/open-web3-stack/XCQ


text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,342 @@
...
+
+## Motivation
+
+In Substrate, runtime APIs facilitate off-chain clients in reading the state of the consensus system.
+However, different chains may expose different APIs for a similar query or have varying data types, such as doing custom transformations on direct data, or differing `AccountId` types.
+This diversity also extends to client-side, which may require custom computations over runtime APIs in various use cases.
+Therefore, tools and UI developers often access storage directly and reimplement custom computations to convert data into user-friendly representations, leading to duplicated code between Rust runtime logic and UI JS/TS logic.
+This duplication increases workload and potential for bugs.
+
+Therefore, A system is needed to serve as an intermediary layer between concrete chain runtime implementations and tools/UIs, to provide a unified interface for cross-chain queries.
```
franciscoaguirre, line 34:
```suggestion
Therefore, a system is needed to serve as an intermediary layer between concrete chain runtime implementations and tools/UIs, to provide a unified interface for cross-chain queries.
```

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,342 @@
...
+
+## Stakeholders
+
+- Runtime Developers
+- Tools/UI Developers
+
+## Explanation
+
+The overall query pattern of the XCQ consists of three components:
```
franciscoaguirre, line 43:
```suggestion
The overall query pattern of XCQ consists of three components:
```

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,342 @@
...
+The query is the query program in PolkaVM program binary format.
+The input is the query arguments that is SCALE-encoded.
+The weight limit is the maximum weight allowed for the query execution.
+- `metadata`: Return metadata of supported extensions (introduced in later section) and methods, serving as a feature discovery functionality
+
+**Example XCQ Runtime API**:
+```rust
+decl_runtime_apis! {
+    pub trait XcqApi {
+        fn execute_query(query: Vec<u8>, input: Vec<u8>, weight_limit: u64) -> XcqResult;
```
franciscoaguirre, line 63:
```suggestion
        fn execute_query(query: Vec<u8>, input: Vec<u8>, gas_limit: u64) -> XcqResult;
```

Is this limit the PVM gas instead of the weight?

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,342 @@
...
+              extension_id: u64,
+              call_ptr: u32,
+              call_len: u32|
+              -> Result<u64, ExtensionError> {
+                  ...
+              });
+```
+- `PermController`: Filters guest XCQ program calling requests, useful for host chains to disable some queries by filtering invoking sources.
+```rust
+pub trait PermController {
```
franciscoaguirre, line 229: Maybe `Authorizer` is a better name? Or `PermissionsController` (not abbreviated)

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,342 @@
...
+      - Accurate conversion between given weight limit and the gas limit of PolkaVM
+    - Negative test cases:
+      - Queries exceeding weight limits
+      - Invoking queries from unauthorized sources
+    - Edge cases:
+      - Queries with minimal or maximum allowed input sizes
+  - Integration tests to ensure proper interaction with off-chain wallets/UI and on-chain XCM, including the aforementioned use cases in **Motivation** section.
+
+- Security:
+  - The XCQ system must enforce a strict read-only policy for all query operations. A mechanism should be implemented to prevent any state-changing operations within XCQ queries.
```
franciscoaguirre, line 313: Could the execution of the program be sandboxed?

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,342 @@
...
+- <https://forum.polkadot.network/t/wasm-view-functions/1045> is the original discussion about having a mechanism to avoid code duplications between the runtime and front-ends/wallets. In the original design, the custom computations are compiled as a wasm function.
+- <https://github.com/paritytech/polkadot-sdk/issues/216> is the issue tracking the `view functions` implementation in runtime implementations
+- <https://github.com/paritytech/polkadot-sdk/pull/4722> is the on-going `view function` pull request. It works at pallet level. If two chains use two different pallets to provide similar functionalities, like pallet-assets and pallet-erc20, we still need to have different codes to support. Therefore, it doesn't conflict with XCQ, and can be utilized by XCQ.
+
+## Unresolved Questions
+
+
+## Future Directions and Related Material
+
+Since XCQ are supported both in off-chain and on-chain, a related XCM-Format RFC should be proposed.
```
franciscoaguirre, line 342: We recently merged XCM RFCs into the fellowship RFCs. So the new instruction and behaviour of XCM could be described here.

text/0126-introduce-xcq.md
indirection42: It seems that the metrics of the PVM gas are in development now, and the gas value is an inner value that should not be exposed to users.

text/0126-introduce-xcq.md
indirection42: Yes, we can do the query in a transaction, and rollback finally.

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,384 @@
...
+}
+```
+
+Note: `ty` is represented by a meta-type system called `xcq-types`
+
+#### xcq-types
+
+`xcq-types` is a meta-type system similar to `scale-info` but simpler.
+It enables different chains with different type definitions to work via a common operation.
+Front-end codes constructs call data to XCQ programs according to metadata provided by different chains.
```
jsdw, line 126: I'd like to know more about `xcq-types`, ie what are the existing issues with `scale-info` here? Since it just describes the shapes of types, one can compare two sets of `scale-info` types across chains already, or is there more to it?

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,384 @@
...
+        }
+        #[allow(unused_variables)]
+        fn balance(asset: <Self::Config as extension_fungibles::Config>::AssetId, who: <Self::Config as extension_fungibles::Config>::AccountId) -> <Self::Config as extension_fungibles::Config>::Balance {
+            100
+        }
+    }
+}
+```
+
+- `ExtensionExecutor`: Connects extension implementations and `xcq-executor`.
```
jsdw, line 211: If one wants to expose pallet view functions or runtime APIs to the XCQ programs, do extensions have to be written which call into them (and then those extensions are what is exposed in the XCQ program)?

Is the reason for having this new extension stuff instead of exposing raw runtime APIs and pallet view functions to XCQ programs that extensions have associated types which can be tweaked by different runtimes, whereas runtime APIs do not? (If that were the only reason, I wonder whether runtime APIs could be extended to optionally support associated types to achieve the same without a new thing being needed?)

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,384 @@
...
+The query is the query program in PolkaVM program binary format.
+The input is the query arguments that is SCALE-encoded.
+The weight limit is the maximum weight allowed for the query execution.
+- `metadata`: Return metadata of supported extensions (introduced in later section) and methods, serving as a feature discovery functionality
+
+**Example XCQ Runtime API**:
+```rust
+decl_runtime_apis! {
+    pub trait XcqApi {
+        fn execute_query(query: Vec<u8>, input: Vec<u8>, weight_limit: u64) -> XcqResult;
```
jsdw, line 63: I guess any user can call this Runtime API and provide their XCQ program and input args? Is it expected that the runtime will have some default gas/weight limit to make it more difficult for users to supply long running/complex programs that DoS the node?

text/0126-introduce-xcq.md
indirection42: A1: Yes, the fundamental APIs exposed to XCQ programs are organized at the extension level.

A2: It's not the only reason. Runtime APIs cannot be recognized via some identifications, while extensions can be recognized via extension ids.

text/0126-introduce-xcq.md
indirection42: The `weight_limit` argument in this single API restricts the usage of a single execution. The overall resource limit is enforced by Runtime config and RPC rate limiting, which is out of the scope for a single RuntimeAPI. As for XCM usage, the gas/weight limit can be enforced by `BuyExecution` instruction.

text/0126-introduce-xcq.md
jsdw: > Runtime APIs cannot be recognized via some identifications

Runtime API methods can all be identified as `$trait_name ++ _ ++ $method_name`, and so we could expose functions with similar names to the PolkaVM program? I suspect I am missing something :)

If Runtime APIs "just worked" natively, it would mean that we could provide a single stable interface to users (who can call the APIs themselves) and XCQ programs (which could call the same APIs), and wouldn't need to add another concept ro Runtimes.

text/0126-introduce-xcq.md
jsdw: Another point here: I'd assume that we would want XCQ metadata to be a part of the frame-metadata that we hand back when users call the `Metadata_metadata_at_version` runtime API? In that case, it feels beneficial to use the existing `scale_info` based type information used by the rest of the metadata over having a separate concept to understand and integrate there.

indirection42: Yes, makes sense. we are experimenting with switching back to `scale-info`.

text/0126-introduce-xcq.md
xlc: The main reason we created XCQ extension is to have content based address. i.e. The extension ID is the hash of the extension definition, which includes the methods and argument types. So we can for sure know the compatibility with a matching extension ID. Runtime API is using versioning for that, which requires extra collaboration.
Besides, Runtime API are designed to be invoked from outside of wasm, which is only half of the use case of XCQ, the other half is XCM. We cannot assume every runtime API are safe to be dispatched by XCM.

jsdw: Thankyou for your reply!

>  The extension ID is the hash of the extension definition, which includes the methods and argument types. So we can for sure know the compatibility with a matching extension ID. Runtime API is using versioning for that, which requires extra collaboration.

How does having a content based address help when compared with non content based addresses (ie the address just being the extension trait  and method name, like Runtime APIs)? 

Looking at some of the possibilities:
- **If I change the semantics but not the type information of some method:** 
   - **Using content based address:** the method will no longer do what is expected, but will still be callable via the same hash. 
   - **Using non-content based address:** same as above
   - **Verdict:** Neither approach is any different here; we require social agreement (mostly a no brainer ofc) to not change these things.
- **If I change the semantics _and_ type information (ie name, input, return types):**
   - ** Using content based address:** calling the "old" address will now fail with some error that it doesn't exist.
   - ** Using non-content based address:** calling the "old" address will still work, but fail with some type error on encoding or decoding.
   - **Verdict:** Both approaches lead to an error; we require social agreement to not change/remove methods if we want to avoid breakage.

In other words, however we approach addressing things, we still need people to agree not to make breaking changes to these functions if we want to avoid breaking compatibility. 

Do content based addresses help in some other way that I am missing (definitely likely!!)? Is the thinking here mainly around XCM ie making it more likely for queries that they call to no longer be accessible?

And a couple more Dumb Qs: what do extensions provide on top of runtime APIs that ensures that they are safe to be dispatched by XCM (or marks them as not being safe)? Might be there some extensions that we want to be able to call in XCQ programs but not via XCM?

xlc: The current polkadot-sdk runtime calls are name based and have many known issues and most of them can be addressed by content based address, which is more commonly used. e.g. EVM selector.

Some advantages:
- The extension ID fixed the name and types, the name based approach only fixed the name
- The extension ID is hash of the whole interface, whereas the name, well, is just whatever the name referring to.
  - e.g. `0x123456` could mean `trait Fungibles { fn balanceOf(who: AccountId) -> U128; fn totalSupply() -> U128 }` but with name, `Fungibles` means a trait named Fungibles but does not constraint the methods. A name of `Fungibles.balanceOf` is just a function name, and not telling anything about other members.
- Unexpected change of type can be detected easily because it changes the extension ID. Not the case for name based ID.
- There is no need to have central name registry. 3 different chains may all want to propose their version of `Fungibles` interface and who decide who got it? Not an issue for hash.

xlc: `PermissionController` defined in the RFC is the mechanism to only allow some extension in some context.

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,354 @@
...
+The weight limit is the maximum weight allowed for the query execution.
+- `metadata`: Return metadata of supported extensions (introduced in later section) and methods, serving as a feature discovery functionality.
+The representation and encoding mechanism is similar to the [`frame-metadata`](https://github.com/paritytech/frame-metadata/), using `scale-info`.
+
+**Example XCQ Runtime API**:
+
+```rust
+decl_runtime_apis! {
+    pub trait XcqApi {
+        fn execute_query(query: Vec<u8>, input: Vec<u8>, weight_limit: u64) -> XcqResult;
```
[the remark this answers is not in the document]
xlc: Proof size doesn't really matter in the context of runtime API but yeah u64 is ambitious. We could use the Substrate Weight type and ignore proof size

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,354 @@
...
+impl extension_fungibles::Config for ExtensionImpl {
+    type AssetId = u32;
+    type AccountId = [u8; 32];
+    type Balance = u64;
+}
+impl_extensions! {
+    impl extension_core::ExtensionCore for ExtensionImpl {
+        type Config = ExtensionImpl;
+        fn has_extension(id: <Self::Config as extension_core::Config>::ExtensionId) -> bool {
+            matches!(id, 0 | 1)
```
franciscoaguirre, line 153: I thought extension ids were hashes of the content? The extension decides its id?

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,354 @@
...
+4. Program Execution:
+   1. Type Checking: The program uses the `return_ty` function to ensure compatibility with supported chain extensions.
+   2. Query Execution: The program executes the query using `host_call` and performs custom computations.
+   3. Result Serialization: The program serializes the result, writes it to shared memory, and returns the pointer and length to the executor.
+
+5. Result Retrieval: The Executor reads the result from shared memory and returns it to the caller.
+
+### XCM integration
+
+The integration of XCQ into XCM is acheived by adding a new instruction to XCM, as well as a new variant of the `Response` type in `QueryResponse` message.:
```
franciscoaguirre, line 253:
```suggestion
The integration of XCQ into XCM is achieved by adding a new instruction to XCM, as well as a new variant of the `Response` type in `QueryResponse` message.:
```

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,354 @@
...
+
+- A new `ReportQuery` instruction
+
+```rust
+ReportQuery {
+  query: SizeLimitedXcq,
+  weight_limit: Option<Weight>,
```
franciscoaguirre, line 260: How can someone get this weight? Is it easy? If not, we could just limit it by the overall fee paid in `PayFees`?
We need to have a good way of using this, otherwise everyone will just use `Unlimited`

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,354 @@
...
+- `XcqResult = 6 (XcqResult)`
+  `XcqResult` is a enum
+  - `Ok = 0 (Vec<u8>)`: XCQ executes successfully with a SCALE-encoded response.
+  - `Err = 1 (ErrorCode)`: XCQ fails with an error code.
+`ErrorCode` is a enum
+- `ExceedMaxWeight = 0`
+- `MemoryAllocationError = 1`
+- `MemoryAccessError = 2`
+- `CallError = 3`
+- `OtherPVMError = 4`
```
franciscoaguirre, line 289: I'm wondering if there's a way to improve these errors. I guess you could run and debug your XCQ program directly on the destination chain.

text/0126-introduce-xcq.md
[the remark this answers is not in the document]
mrshiposha: Maybe we could name it `weight_ref_time_limit` to emphasize that it is intentionally one-dimensional. We could also leave a comment reminding people that the proof size isn't relevant in this context.

text/0126-introduce-xcq.md
mrshiposha: When I read the RFC, I understood the `ExtensionCore` as a "special" "foundational" extension. It seems it will have some generic implementation where the extension IDs will indeed be content-addressed.

Is that so? Is this code just an example of "some return value"?

xlc: this is just a dummy implementation. we should leave it out in the example

text/0126-introduce-xcq.md
xlc: maybe we should modify `execute_query` to have it also return used weight so it can be used to estimate weight usage
We could make `Unlimited` weight means the weight to be limited by fees. Limited weight will be useful for the case that we need to reserve some fees for other instruction, such as transact

text/0126-introduce-xcq.md
xlc: Yes everything are deterministic so replay with extra logging enabled should be the preferred way for debugging. That's how we debug XCM today.

gavofyork:
This is not what I had in mind nor described in the channel when I introduced XCQ. While it may in and of itself be a reasonable proposal worthy of consideration and support, I think it should be given its own name ("view functions", I suppose, since this is appears to be a write up of that proposed as stated in the prior art section).

XCQ is an abstract language similar in nature to XCM. Chains offering XCQ functionality offer an XCQ Translator. This is a small pieces of PVM code which translates an XCQ expression into a) a series of keys whose values must be fetched and b) a means of combining those values and the original expression into a result decodable by SCALE into an XCQ value.

Crucially, XCQ is an *abstraction*. Like XCM it requires no knowledge whatsoever of the underlying chain or what it supports.

xlc:
On an early draft of XCQ, it is instruction based similar to XCM. However, in order to execute some expression to combine/transform the values, the instruction set needs to cover arithmetic and some memory management. Then the general feedbacks I got is that we should just use PVM instead and hence the direction we are working so far.

The extension design is not too much differ to XCM adapters but generalized and more integrated. I believe the current proposal is able to achieve the goals of the original XCQ idea. XCQ program developer nor XCQ program users need to know the chain specific details (unless they are working with chain specific extensions).

gavofyork:
What is proposed has no obvious abstraction capability.

xlc:
The intended usage is that we also define a set of common extensions (e.g. fungibles and non-fungibles) that would be confirmed by all implementation chains. Then people will be able to develop generic XCQ programs with the ability to access this information.
The common extensions are not included in this RFC because they don't need to. This RFC is about the low level primitives that could be used to build the abstractions. After this is accepted, we could have another RFC to define common extensions.
Polkadot-SDK can offer standard implementations for pallets it provides as well. Basically the XCM adapters and related pallets. But the key things is that chains can just do their own implementation without being blocked by polkadot-sdk or any  upstream library.

gavofyork:
That's definitely not XCQ then. XCQ is meant to be a sister-protocol to XCM for inspection rather than mutating operations. The clue is in the name. What you've done here is not what I described nor could it be considered "read-only-XCM".

xlc:
This is indeed lower level than an abstract language for querying state. But the idea is that we can build such abstract language on top of this work (by defining extensions and compile some DSL or Rust code to PVM).

Will try to find another name to better describe this work.

text/0126-introduce-xcq.md
```diff
@@ -0,0 +1,342 @@
...
+     #[polkavm_derive::polkavm_import]
+     extern "C" {
+         fn return_ty(extension_id: u64, call_index: u32) -> u64;
+     }
+     ```
+
+     Results are SCALE-encoded bytes, with the pointer address and length packed similarly to `host_call`.
+
+- Exported Functions:
+  - `main`: The entry point of the XCQ program. It performs type checking, arguments and results passing and executes the query.
```
koute, line 232: This is reasonable as a reference for programmers/as part of a potential SDK, but RFC-wise this should be defined on a much, much lower level (on the level of raw PVM disassembly).

So:

- You should probably refer to the JAM spec where PVM is defined.
- Considering the space constraints (you want these programs to be copy-pastable as hex strings, correct?) you most likely don't want to use a `.polkavm` blob, and just have a raw code blob.
- If you're going to have only a single entry point then it should be defined that the execution starts at address 0 (like JAM does for one of its entry points).
- Calling into the host should be defined in terms of the `ecalli` instruction, with a hardcoded index for each host call (like JAM does), and it should specify which registers are used for passing the arguments (e.g. for three arguments those would be `a0`, `a1` and `a2`, which are 7th, 8th and 9th registers respectively - JAM doesn't define those names, but in common use it's more convenient to refer to the register by their RISC-V ABI names).
- No need to pack multiple values into a u64 when returning; you can just return a tuple/use two registers (`a0` and `a1`).
- The memory map should also be defined. At the bare minimum you need 4k of stack space 
between `0xfffef000` and `0xfffff000`. You also need to think about whether you want the program to be able to have RO data/RW data sections (you don't necessarily need them), and whether you want to hardcode the sizes or make them configurable (by e.g. including them in the header or whatever)

xlc: Is this sort of tweaks easily supported by the current PVM tooling?
It make sense to define the entrypoint on PVM level so it play nicely with alternative language implementations, however it feels a bit too low level as well. For example, I don't think we define host functions in WASM level.

koute: > Is this sort of tweaks easily supported by the current PVM tooling?

All of these are needed for JAM, so yes.

> It make sense to define the entrypoint on PVM level so it play nicely with alternative language implementations, however it feels a bit too low level as well. For example, I don't think we define host functions in WASM level. The C FFI should be defined as part of PVM spec so we can depend on it?

Any higher level definition should be only there for reference and be non-normative. The PVM spec doesn't concern itself with the C FFI, it doesn't know anything about imports or exports or symbol names or anything like that, and as far as I know there are no plans to add any of these concepts to the GP.

WASM *has* a concept of imports, of exports, and of an explicit ABI for FFI. In WASM if you want to export a function you define it by name and explicitly define its arguments and return values, and that gets embedded in the blob (same for imports). In PVM this doesn't exist - all imports and exports are defined implicitly (there's no header/section to define them), there are no names for anything (host calls numbers are hardcoded and are referenced by their number) and the ABI is also completely implicit.

xlc:
@koute if you have the chance, could you take another review?

xlc:
/rfc process 0x481398e45142a1ee9ae206d11bceabaf5ab81698aba2f812a3da40a7d6be1041

PR #135, opened 2025-01-06
Standardize compressed blob prefixes

s0me0ne-unkn0wn:
[Rendered](https://github.com/s0me0ne-unkn0wn/RFCs/blob/s0me0ne/compressed-blob-prefixes/text/0135-compressed-blob-prefixes.md)

### DIFF
--- a/text/0135-compressed-blob-prefixes.md
+++ b/text/0135-compressed-blob-prefixes.md
@@ -1,0 +1,101 @@
+# RFC-0135: Compressed Blob Prefixes
+
+|                 |                                                                    |
+| --------------- | ------------------------------------------------------------------ |
+| **Start Date**  | 2025-01-06 |
+| **Description** | Standardize compressed blob prefixes |
+| **Authors**     | s0me0ne-unkn0wn (13WGadgNgqSjiGQvfhimw9pX26mvGdYQ6XgrjPANSEDRoGMt) |
+
+## Summary
+
+This RFC proposes a change that makes it possible to identify types of compressed blobs stored on-chain, as well as used off-chain, without the need for decompression.
+
+## Motivation
+
+Currently, a compressed blob does not give any idea of what's inside because the only thing that can be inside, according to the spec, is Wasm. In reality, other blob types are already being used, and more are to come. Blob decompression is a blocking operation, and routing the blob to a proper worker should not involve its decompression for the sake of efficiency. Thus, it is necessary to introduce a mechanism allowing to identify the blob type without decompressing it.
+
+This proposal is intended to:
+1. Fill up gaps in the Polkadot spec, increasing its preciseness and putting it in line with mechanisms already being used in practice but not yet standardized;
+2. Support future work enabling Polkadot to execute PolkaVM and, more generally, other-than-Wasm parachain runtimes;
+3. Allow developers to introduce compression methods other than `zstd` seamlessly in the future.
+
+## Stakeholders
+
+Node developers are the main stakeholders for this proposal. It also creates a foundation on which parachain runtime developers will build.
+
+## Explanation
+
+### Overview
+
+The current approach to compressing binary blobs is defined in [subsection 2.6.2](https://spec.polkadot.network/chap-state#sect-loading-runtime-code) of Polkadot spec. It involves using `zstd` compression, and the resulting compressed blob is prefixed with a unique 64-bit magic value specified in that subsection. Said subsection only defines the means of compressing Wasm code blobs; no other compression procedure is currently defined by the spec.
+
+However, in practice, the current de facto protocol uses the said procedure to compress not only Wasm code blobs but also proofs-of-validity. Such a usage is not stipulated by the spec. Currently, having solely a compressed blob, it's impossible to tell what's inside it without decompression, a Wasm blob, or a PoV. That doesn't cause any problems in the current de facto protocol, as Wasm blobs and PoV blobs take completely different execution paths, and it's impossible to mix them.
+
+Changes proposed below are intended to:
+* Bring the spec into line with the currently used de facto protocol;
+* Define the means for distinguishing compressed blob types in a backward-compatible and future-proof way;
+* Implement the means defined;
+* Deprecate the legacy compression/decompression mechanism in favor of the proposed one.
+
+### Proposed spec changes
+
+* Add a subsection titled "Blob Compression" and describe how the binary data is compressed generally and list the following well-known prefixes:
+
+| Prefix name | Prefix bytes | Description |
+| -- | -- | -- |
+| `CBLOB_ZSTD_WASM_OR_POV` | 82, 188, 83, 118, 70, 219, 142, 5 | Wasm code blob or PoV, zstd-compressed |
+| `CBLOB_ZSTD_WASM_CODE`   | 82, 188, 83, 118, 70, 219, 142, 6 | Wasm code blob, zstd-compressed        |
+| `CBLOB_ZSTD_PVM_CODE`    | 82, 188, 83, 118, 70, 219, 142, 7 | PolkaVM code blob, zstd-compressed     |
+
+* Amend [subsection 2.6.2 "Loading the Runtime Code"](https://spec.polkadot.network/chap-state#sect-loading-runtime-code) and make it reference the newly created "Blob Compression" subsection instead of specifying the prefix and the compression method explicitly;
+* Amend [subsection 8.3.2 "Runtime Compression"](https://spec.polkadot.network/chapter-anv#sect-runtime-compression), which currently reads "Not documented yet", and make it describe the actual parachain runtime compression technics, referencing the newly created  "Blob Compression" subsection;
+* In [section 8, "Availability & Validity"](https://spec.polkadot.network/chapter-anv), either in a separate new subsection or in one of the existing ones, introduce a PoV compression concept, linking it to the newly created "Blob Compression" subsection.
+
+### Proposed code changes
+
+No runtime code is changed. Node-side changes are trivial; a PoC already implemented as a part of [SDK PR#6704](https://github.com/paritytech/polkadot-sdk/pull/6704) may be used as an example.
+
+### Timeline
+
+1. The proposed prefix changes are implemented and released. No logic changes yet;
+2. After the supermajority of production networks' nodes upgrades, one more change is released that adds `CBLOB_ZSTD_WASM_CODE` prefix instead of `CBLOB_ZSTD_WASM_OR_POV` when compiling and compressing Wasm parachain runtimes;
+3. Conservatively, wait until no more PVFs prefixed with `CBLOB_ZSTD_WASM_OR_POV` remain on-chain. That may take quite some time. Alternatively, create a migration that alters prefixes of existing blobs;
+4. Rename `CBLOB_ZSTD_WASM_OR_POV` to `CBLOB_ZSTD_POV`, as the PoVs remain the only entity compressed with this prefix by that time. Thus no collator protocol changes are needed.
+
+## Drawbacks
+
+Currently, the only requirement for a compressed blob prefix is not to coincide with Wasm magic bytes (as stated in code comments). Changes proposed here increase prefix collision risk, given that arbitrary data may be compressed in the future. However, it must be taken into account that:
+* Collision probability per arbitrary blob is ≈5,4×10⁻²⁰ for a single random 64-bit prefix (current situation) and ≈1,6×10⁻¹⁹ for the proposed set of three 64-bit prefixes (proposed situation), which is still low enough;
+* The current de facto protocol uses the current compression implementation to compress PoVs, which are arbitrary binary data, so the collision risk already exists and is not introduced by changes proposed here.
+
+## Testing, Security, and Privacy
+
+As the change increases granularity, it will positively affect both testing possibilities and security, allowing developers to check what's inside a given compressed blob precisely. Testing the change itself is trivial. Privacy is not affected by this change.
+
+## Performance, Ergonomics, and Compatibility
+
+### Performance
+
+The current implementation's performance is not affected by this change. Future implementations allowing for the execution of other-than-Wasm parachain runtimes will benefit from this change performance-wise.
+
+### Ergonomics
+
+The end-user ergonomics is not affected. The ergonomics for developers will benefit from this change as it enables exact checks and less guessing.
+
+### Compatibility
+
+The change is designed to be backward-compatible. 
+
+## Prior Art and References
+
+[SDK PR#6704](https://github.com/paritytech/polkadot-sdk/pull/6704) (WIP) introduces a mechanism similar to that described in this proposal and proves the necessity of such a change.
+
+## Unresolved Questions
+
+None
+
+## Future Directions and Related Material
+
+This proposal creates a foundation for two future work directions:
+* Proposing to introduce other-than-Wasm code executors, including PolkaVM, allowing parachain runtime authors to seamlessly change execution platform using the existing mechanism of runtime upgrades;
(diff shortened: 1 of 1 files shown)


text/0135-compressed-blob-prefixes.md
```diff
@@ -0,0 +1,101 @@
...
+| **Description** | Standardize compressed blob prefixes |
+| **Authors**     | s0me0ne-unkn0wn (13WGadgNgqSjiGQvfhimw9pX26mvGdYQ6XgrjPANSEDRoGMt) |
+
+## Summary
+
+This RFC proposes a change that makes it possible to identify types of compressed blobs stored on-chain, as well as used off-chain, without the need for decompression.
+
+## Motivation
+
+Currently, a compressed blob does not give any idea of what's inside because the only thing that can be inside, according to the spec, is Wasm. In reality, other blob types are already being used, and more are to come. Blob decompression is a blocking operation, and routing the blob to a proper worker should not involve its decompression for the sake of efficiency. Thus, it is necessary to introduce a mechanism allowing to identify the blob type without decompressing it.
```
eskimor, line 15: While I think this RFC makes sense in general, I feel like we should be able to find better arguments than an implementation detail in the Parity implementation?

s0me0ne-unkn0wn: Fair enough. I'll fix the wording.

text/0135-compressed-blob-prefixes.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Proposed spec changes
+
+* Add a subsection titled "Blob Compression" and describe how the binary data is compressed generally and list the following well-known prefixes:
+
+| Prefix name | Prefix bytes | Description |
+| -- | -- | -- |
+| `CBLOB_ZSTD_WASM_OR_POV` | 82, 188, 83, 118, 70, 219, 142, 5 | Wasm code blob or PoV, zstd-compressed |
+| `CBLOB_ZSTD_WASM_CODE`   | 82, 188, 83, 118, 70, 219, 142, 6 | Wasm code blob, zstd-compressed        |
+| `CBLOB_ZSTD_PVM_CODE`    | 82, 188, 83, 118, 70, 219, 142, 7 | PolkaVM code blob, zstd-compressed     |
```
bkchr, line 48: Why don't we just introduce a new variant for `PVM` and say that the old variant is only used for WASM?

s0me0ne-unkn0wn: Because the old variant is currently used for PoVs as well.
Yes, we could leave the old variant as-is and just add PVM to minimize the change footprint. But to me, it looks like a bad design. If we add granularity, let's do it correctly right away, not leaving the technical debt behind. When this table grows, that "wasm-or-pov" thing will look like an obvious legacy and will most probably be harder to remove than now.

s0me0ne-unkn0wn:
> The spec is currently outdated and not maintained by anyone.

That's some bad news. I mean, if I want to create my implementation of Polkadot from scratch, where do I go?

text/0135-compressed-blob-prefixes.md
```diff
@@ -0,0 +1,101 @@
...
+
+Node developers are the main stakeholders for this proposal. It also creates a foundation on which parachain runtime developers will build.
+
+## Explanation
+
+### Overview
+
+The current approach to compressing binary blobs is defined in [subsection 2.6.2](https://spec.polkadot.network/chap-state#sect-loading-runtime-code) of Polkadot spec. It involves using `zstd` compression, and the resulting compressed blob is prefixed with a unique 64-bit magic value specified in that subsection. Said subsection only defines the means of compressing Wasm code blobs; no other compression procedure is currently defined by the spec.
+
+However, in practice, the current de facto protocol uses the said procedure to compress not only Wasm code blobs but also proofs-of-validity. Such a usage is not stipulated by the spec. Currently, having solely a compressed blob, it's impossible to tell what's inside it without decompression, a Wasm blob, or a PoV. That doesn't cause any problems in the current de facto protocol, as Wasm blobs and PoV blobs take completely different execution paths, and it's impossible to mix them.
```
bkchr, line 32: The argumentation "it's impossible to mix them" is not really that strong :P

text/0135-compressed-blob-prefixes.md
bkchr: Now I get what you mean. Then I would say we introduce a third variant that is directly `CBLOB_ZSTD_POV` and doesn't do this dance of renaming it in some months or whatever. Basically deprecating the old prefix.

s0me0ne-unkn0wn: That was my first thought. But the bad news is that the PoVs are compressed by collators. We couldn't remove the old prefix until all the collators upgraded and started using the new one, which could take... Forever

s0me0ne-unkn0wn: We could name it `CBLOB_ZSTD_POV` from the beginning and leave a code comment that it is also used for Wasm blobs, but such usage is deprecated. Removing a comment is a little bit less effort than renaming a constant :upside_down_face:

bkchr: > That was my first thought. But the bad news is that the PoVs are compressed by collators. We couldn't remove the old prefix until all the collators upgraded and started using the new one, which could take... Forever

Yeah and I don't see any real problem with this. We have soo many code that is 5 years old that handles stuff that maybe someone still using somewhere.

text/0135-compressed-blob-prefixes.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Unresolved Questions
+
+None
+
+## Future Directions and Related Material
+
+This proposal creates a foundation for two future work directions:
+* Proposing to introduce other-than-Wasm code executors, including PolkaVM, allowing parachain runtime authors to seamlessly change execution platform using the existing mechanism of runtime upgrades;
+* Proposing to use new compression methods that are faster/better/more secure than `zstd` as soon as they appear, streamlining the adoption of new methods and deprecation of the old ones.
```
koute, line 101: As far as I know it's somewhat unlikely that a significantly better general purpose compression codec than `zstd` will appear, at least probably in the next few decades, so this is somewhat of a weak motivation.

However, it would be feasible to have a *dedicated* (so not a general purpose) codec to compress certain types of blobs that could result in better compression and/or speed; for example, precomputing a zstd dictionary to compress `.wasm` blobs would most likely give us measurable improvements, but (even though it would still be compressed with zstd) it would require a mechanism (like what you propose here) to specify that such a precomputed dictionary should be used (because the dictionary wouldn't be transmitted, so it wouldn't be "normal" zstd compression).

text/0135-compressed-blob-prefixes.md
koute: Personally I agree with Basti. I would probably have a `CBLOB_ZSTD_LEGACY` (or whatever you'd like to call it) which would be the current prefix, and then introduce three new prefixes: one for PVFs, one for WASM, and one for PVM.

s0me0ne-unkn0wn:
@bkchr, returning to that "let's follow the typical RFC format" part and explaining why I'd like to keep it that formal.

According to the Polkadot Whitepaper, _"A system of RFCs, not unlike the Python Enhancement Proposals, will allow a means of publicly collaborating over protocol changes and upgrades"_. Thus, the RFCs are proposals to change the protocol. But what protocol? We must understand what we're changing. I understand that currently, we're more or less in limbo between the Polkadot Whitepaper and the JAM Gray paper, and that would be acceptable if we found ourselves in this situation, say, for a couple of months. But JAM is not arriving in a couple of months, and if the spec is not currently maintained, that means we're flying in the Bitcoin mode with an implementation-defined spec. What protocol is changed by any given RFC, then? That one which had been implemented by the time when the RFC was first published?  Or will we count from the latest stable release?

Again, if a team wants to start its very own implementation of Polkadot tomorrow, where does it start? What should they implement? Should they just read our Rust code and do the same? They may not even know Rust, after all. Or should they take the latest outdated spec and apply all the RFCs over it? Even the infamous EU bureaucracy doesn't treat you like that, you can always get a full text of legislation that is in force right now instead of getting the very first one from 1990s and a hundred amendments to compile everything yourself.

That's why I believe the spec is super important. The spec should be a representation of the current state, and RFCs should propose changes to the spec, not the code. Yes, we're not living in a perfect world, and it may become more or less stale sometimes, but if we drop its support altogether, a lot of things lose sense, including alternative implementations.

Alternatively, if we wanted to follow the IETF's RFC model, we could have a set of "root RFCs" defining the protocol and other RFCs amending the root ones, but I believe for a protocol as complex as Polkadot it's a counterproductive approach as the set of root RFCs would just become the spec on its own. The IETF's TCP protocol had 10 RFCs on its way before it was rewritten from scratch to include all the amendments, and that happened over a time span of 40+ years. We in Polkadot, on the other hand, have already had 60+ RFCs since Fellowship was formed, roughly in a year and a half. The IETF's approach wouldn't work here.

I'll go ahead and tag @gavofyork here 'cause I believe it's a really bad situation. If the spec is not maintained, we should change it today or tomorrow at the latest.

bkchr:
There are two things, the spec not being maintained and then there are like the standard RFC format. While I'm also not happy that the spec is not maintained and I would like to change this, it is complicated right now and I don't think this will change until we have JAM. Better to accept the reality instead of denying it. 
For the RFC I would just follow the format that everyone is using and not trying to push for the special way for no real benefit right now.

bkchr:
> Polkadot Fellowship does not accept reality, it forms it! 😁 Fellowship is a resourceful organization and could acquire both internal and external resources to get that job done. What happens now is just a lack of understanding of the importance of the task. Having an up-to-date spec is a basic need, not a luxury.

You are free to take this on and make it happen :)

text/0135-compressed-blob-prefixes.md
```diff
@@ -0,0 +1,87 @@
...
+| `CBLOB_ZSTD_PVM_CODE`    | 82, 188, 83, 118, 70, 219, 142, 8 | PolkaVM code blob, zstd-compressed     |
+
+No runtime code changes should be needed to imnplement this proposal. Node-side changes are trivial; a PoC already implemented as a part of [SDK PR#6704](https://github.com/paritytech/polkadot-sdk/pull/6704) may be used as an example.
+
+### Timeline
+
+1. The proposed prefix changes are implemented and released. No logic changes yet;
+2. After the supermajority of production networks' nodes upgrades, one more change is released that adds `CBLOB_ZSTD_WASM_CODE` prefix instead of `CBLOB_ZSTD_LEGACY` when compiling and compressing Wasm parachain runtimes, and `CBLOB_ZSTD_POV` instead of `CBLOB_ZSTD_LEGACY` when compressing PoVs;
+3. Conservatively, wait until no more PVFs prefixed with `CBLOB_ZSTD_LEGACY` remain on-chain. That may take quite some time. Alternatively, create a migration that alters prefixes of existing blobs;
+4. Removing `CBLOB_ZSTD_LEGACY` prefix will be possible after all the collators in all the networks switch to compression with `CBLOB_ZSTD_POV` prefix.
```
bkchr, line 49: *Every* parachain node also needs to upgrade to support the new prefixes. We will need to think on how to nudge the people to switch to this option, but it will take some time. Which makes the coordination more complicated.

bkchr (approved):
Generally looks good. 

Just need to be even more conservative when it comes to the time lines as this is not only affecting the relay chain or the collators, but every node.

text/0135-compressed-blob-prefixes.md
s0me0ne-unkn0wn: Then I have to ask again, why don't you want to avoid this problem altogether by reusing the existing prefix for PoVs?

s0me0ne-unkn0wn:
> You are free to take this on and make it happen :)

I don't think a single person can do it because I don't believe there exists a single person who knows all the aspects of the current implementation in depth. I think I could design a pipeline that streamlines the spec updates. But first we should communicate and agree on the procedure anyway. Sounds like one more RFC by itself.

text/0135-compressed-blob-prefixes.md
bkchr: It is not about PoVs. When a runtime would use a new prefix, old nodes would fail to decompress and thus stall. The Substrate executor is also using this prefix to detect the compressed runtime.

bkchr:
I did not meant that you should keep the spec updated on your own. Someone needs to setup the process, find the people to update the spec, find out what is outdated etc.

text/0135-compressed-blob-prefixes.md
s0me0ne-unkn0wn: Yeah now I got it at last. You're right, there's no point in reusing the legacy prefix in that case. We should incentivize people to upgrade, and the best incentive may be delivering PolkaVM runtimes and lowering PVF deposits for them.

xlc: Are there any downsides on having `CBLOB_ZSTD_LEGACY` stay for a while? All the nodes will be forced to upgrade to JAM when it is live and can we do it at that time or that's too long?

s0me0ne-unkn0wn: No _real_ downsides indeed, just a reluctance to leave any legacy behind if that can be avoided. I come from the Perl world, and legacy is my pain point :sweat_smile:

text/0135-compressed-blob-prefixes.md
```diff
@@ -0,0 +1,87 @@
...
+3. Conservatively, wait until no more PVFs prefixed with `CBLOB_ZSTD_LEGACY` remain on-chain. That may take quite some time. Alternatively, create a migration that alters prefixes of existing blobs;
+4. Removing `CBLOB_ZSTD_LEGACY` prefix will be possible after all the nodes in all the networks cease using the prefix which is a long process, and additional incentives should be offered to the community to make people upgrade.
+
+## Drawbacks
+
+Currently, the only requirement for a compressed blob prefix is not to coincide with Wasm magic bytes (as stated in code comments). Changes proposed here increase prefix collision risk, given that arbitrary data may be compressed in the future. However, it must be taken into account that:
+* Collision probability per arbitrary blob is ≈5,4×10⁻²⁰ for a single random 64-bit prefix (current situation) and ≈2,17×10⁻¹⁹ for the proposed set of four 64-bit prefixes (proposed situation), which is still low enough;
+* The current de facto protocol uses the current compression implementation to compress PoVs, which are arbitrary binary data, so the collision risk already exists and is not introduced by changes proposed here.
```
eskimor, line 55: Would it make sense to de-risk further, by deprecating blank/unprefixed payloads entirely?

s0me0ne-unkn0wn: That's a good point :thinking: But I'm afraid that would involve changing the PoV format in the collator protocol which is one more change that could take years to propagate, and we'd need to support both versions all the way. Optimistically, it would take more time to stabilize than for JAM to arrive :slightly_smiling_face:

s0me0ne-unkn0wn: Thinking a little bit more about it... Might it be vice versa? Maybe it's time to deprecate uncompressed payloads? They don't make much sense to me anyway, uncompressed Wasm would waste onchain storage and uncompressed PoV — the network throughput. If we ever need something uncompressed, we could introduce a no-op compression prefix.

@bkchr @koute WDYT?

ggwpez:
We have two refs for this now, can we close one? https://collectives.subsquare.io/fellowship/referenda/302 and https://collectives.subsquare.io/fellowship/referenda/301

s0me0ne-unkn0wn:
301 should be closed (not sure if I can do that myself)

s0me0ne-unkn0wn:
Status update: A fully functional implementation is part of https://github.com/paritytech/polkadot-sdk/pull/6704, which is unlikely to be merged before #145 is implemented and merged. It will most probably require a separate RFC on its own.

If anyone feels it should be separated from that integration, let me know.

PR #146, opened 2025-05-20
Deflationary Transaction Fee Model for Relay- and System Chains

jonasW3F:
This RFC proposes burning 80% of transaction fees on the Relay Chain and System Chains, adding to the existing deflationary capacity.


Tomen:
while we are at it, we could set the rate that goes to validators. If OpenGov wants to change inflation params in the future, it might want to give more tx fees to validators

jonasW3F:
> while we are at it, we could set the rate that goes to validators as OpenGov param. If OpenGov wants to change inflation params in the future, it might want to give more tx fees to validators

I don't think this is worth the effort. As the RFC states, with acceptance of RFC32, the goal is to remove almost all activity away from the Relay Chain (which is already happening). Therefore, I'd expect not any significant volume of fees on the Relay Chain itself.

text/0146-deflationary-fee-proposal.md
ultracoconut, line 1: I like the proposal, but I recommend changing some concepts that could lead to misunderstandings by regulators.

ultracoconut:
I like the proposal, but I recommend changing some concepts that could lead to misunderstandings by regulators.

xlc:
Can we accurately evaluate the impact of this without taking consideration of collator reward mechanism?

burdges:
Afaik, there is no reason for collator rewards to be anything more comple than the parachain tallying up some fixed rewards each block, and later messaging the talley to AssetHub.

jonasW3F:
> I like the proposal, but I recommend changing some concepts that could lead to misunderstandings by regulators.

could you elaborate? I'm not sure what you mean.

ultracoconut:
> > I like the proposal, but I recommend changing some concepts that could lead to misunderstandings by regulators.
> 
> could you elaborate? I'm not sure what you mean.
Yes, of course. I'm referring to anything that suggests the token is considered a security. For example, we must avoid implying that the token provides a return without performing any type of work. Instead of expressing that it's deflationary and that this generates some type of profit for holders, we must emphasize that it only offsets inflation.

jonasW3F:
> Can we accurately evaluate the impact of this without taking consideration of collator reward mechanism?

System Chain collators are being paid retrospectively from the infrastructure bounty and the overwhelming part of their income come from that process. The fees don't play a meaningful part of the rewards right now and reducing that part should therefore not change much. With higher activity on AH, I'd expect that collators earn much more with 20% then than with 100% now.

text/0146-deflationary-fee-proposal.md
```diff
@@ -0,0 +1,47 @@
+# RFC-0146: Deflationary Transaction Fee Model for Relay- and System Chains
```
seadanda, line 1: Probably a nit since it's consistent and clear from context, but "System Chains" != "System Parachains", usually "System Chains" for a given network would include the Relay Chain.

text/0146-deflationary-fee-proposal.md
```diff
@@ -0,0 +1,47 @@
...
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 20th May 2025                                                            |
+| **Description** | This RFC proposes burning 80% of transaction fees on the Relay Chain and System Chains, adding to the existing deflationary capacity.     |
+| **Authors**     | Jonas Gehrlein                                                           |
+
+## Summary
+
+This RFC proposes **burning 80% of transaction fees** accrued on Polkadot’s **Relay Chain** and, more significantly, on all **System Chains**. The remaining 20% would continue to incentivize Validators (on the Relay Chain) and Collators (on System Chains) for including transactions.
```
seadanda, line 11: Is there any reasoning behind this specific 80:20 ratio?

seadanda (commented):
I support this high level, but think it would be worth motivating the ratio

text/0146-deflationary-fee-proposal.md
jonasW3F: Thanks for the insight, that would mean "Deflationary Transaction Fee Model for Polkadot's System Chains" is a more precise title?

text/0146-deflationary-fee-proposal.md
jonasW3F: Yes, it would leave validator rewards untouched and yield a unified ratio across all System Chains.

text/0146-deflationary-fee-proposal.md
jonasW3F: Okay, since I assume that I'm not the only one that did not know this nuance, I'll update the RFC to talk about the "Relay Chain and its System Parachains".

text/0146-deflationary-fee-proposal.md
jonasW3F: I added this part excplicitly.

xlc:
> System Chain collators are being paid retrospectively from the infrastructure bounty and the overwhelming part of their income come from that process.

Isn't that just a short-term solution?

jonasW3F:
> > System Chain collators are being paid retrospectively from the infrastructure bounty and the overwhelming part of their income come from that process.
> 
> Isn't that just a short-term solution?

I don't think so. It seems to me things are working according to [RFC0007](https://polkadot-fellows.github.io/RFCs/approved/0007-system-collator-selection.html). @joepetrowski might be the best person to ask whether there are plans for a different system.

Income would be so volatile that I don't see how a fee-based compensation can be suitable in any case. So, I don't think this is affecting this RFC.

ggwpez (approved):

kianenigma (approved):

jonasW3F:
As an addition to the discussion: There's a seminal paper, ["Transaction Fee Mechanism Design for the Ethereum Blockchain: An Economic Analysis of EIP-1559" (Roughgarden, 2020)](https://timroughgarden.org/papers/eip1559.pdf), which highlights the importance of burning transaction fees to prevent off-chain agreements (OCAs) between block producers and users. Roughgarden writes:

> Transferring the revenue from the base fee of a block to the miner of that block is economically equivalent to having no base fee. In this sense, a base fee provides UX improvements only if it is burned (or otherwise withheld from the miner).

This theoretical insight applies to Polkadot, as well. By increasing the portion of fees that are burned, we strengthen the protocol's resistance to such collusion and reinforce the credibility of the base fee as a protocol-defined price signal. This proposal brings Polkadot closer in line with these theoretical results while remaining the practical advantage to provide an incentive to include transactions in the absence of those OCAs.

ggwpez:
Voting ongoing here https://collectives.subsquare.io/fellowship/referenda/355

bkchr:
/rfc process 0xa0fe5aeb81003c8cc50317e6cbfd7089224e95a32b0ba407da5a0e877a50b650

PR #149, opened 2025-05-30
Renewal adjustments to RFC-1

eskimor:
Proposal to fix the issues we are seeing with Agile Coretime in production right now.

### DIFF
--- a/text/0149-rfc-1-renewal-adjustment.md
+++ b/text/0149-rfc-1-renewal-adjustment.md
@@ -166,3 +166,15 @@ This RFC is altering RFC-1 and taking ideas from RFC-17, mainly the introduction
 This RFC should solve the immediate problems we are seeing in production right
 now. Longer term, improvements to the market in terms of price discovery
 (RFC-17) should be considered, especially once demand grows.
+
+There is an edge case remaining for situations like the following:
+
+1. Not enough demand.
+2. Prices drop to the end price.
+3. Someone buys all the cores relatively cheaply & renews.
+4. Market deprived of cores, but price only goes up by renewal bumps (all cores are renewed).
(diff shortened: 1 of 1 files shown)


text/0149-rfc-1-renewal-adjustment.md
```diff
@@ -0,0 +1,168 @@
...
+keep increasing, eventually the least profitable projects will give up their
+cores, freeing up resources for new entrants. 
+
+
+Recommendations for deployment:
+
+1. Deploy with a price curve enforcing a minimum end price, we suggest 10 DOT
+   and 1 KSM. This is to ensure prompt recovery to sound market conditions.
+2. Together with a minimum price, it should be safe to add more cores. We
+   recommend adding at least 10 additional cores, 20 would be better. This
```
seadanda, line 109:
```suggestion
   recommend adding at least 10 additional cores, 20 would be better. This
```

Maybe it is best to specify network-specific numbers for the number of cores to add, since we have more headroom on Polkadot than on Kusama, and with more cores we avoid the low-liquidity risks of the model more effectively (while still having a small reserve)

seadanda (approved):

pandres95:
/rfc process 0x5c3ad5e5c8a9082cfd76dc8b357b8f25b7b8cd4e9179c7a68658155db11b4a30

ltfschoen:
For consideration in RFC-0149 discussions here's an economic analysis of potential market distortions and alternative approaches that may maintain stability while addressing hoarding https://raw.githubusercontent.com/ltfschoen/absolute-red-open-challenge/refs/heads/main/absolute-red.pdf with feedback welcome under CC BY-SA 4.0 attribution terms.

PR #150, opened 2025-06-05
Allow Voting While Delegating

PolkadotDom:
[Rendered](https://github.com/PolkadotDom/RFCs/blob/dom/voting-while-delegating/text/0150-voting-while-delegating.md)

[PR](https://github.com/paritytech/polkadot-sdk/pull/9026)


text/0150-voting-while-delegating.md
```diff
@@ -0,0 +1,111 @@
...
+
+`Users`: We will want users to be aware of the new functionality, though not required.
+
+`Technical Writers`: This change will require rewrites of documentation and tutorials. 
+
+## Explanation
+
+### New Data & Runtime Logic
+
+The [Voting Enum](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/conviction-voting/src/vote.rs#L225) is first collapsed, as there's no longer a distinction between the variants. Then a `(poll index -> retracted votes count)` data item would be added to the user's voting data stored in [VotingFor](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/conviction-voting/src/lib.rs#L165). This would keep track of the per poll balance that has been clawed back from the user by those delegating to them.
```
bkchr: I would like to have the following edge cases covered here in the document:

- The delegate votes, the delegator votes as well, the delegate removes their vote, the delegator removes their delegation, the delegate votes again
- Generally the case of delegator voting on their own, delegate removing their vote and adding it back
- Delegator voting first and then the delegate

There are probably some other edge cases we need to think of.

bkchr (commented):
Thank you! Generally a highly requested feature and I think the general idea looks good. I wrote up some edge cases, for which I already have some ideas how to solve them (probably), but would be really good to have them included.

text/0150-voting-while-delegating.md
PolkadotDom: Thank you for the comment. I'll address here and also update the document.

In general these edge cases are handled by the persistence of the poll_index -> retracted_votes state even when the delegate doesn't have a recorded vote for that poll_index. To showcase, let's take a look at the second edge case: The delegator votes -> their vote is added, the retracted vote count is added to their delegate, tally is updated accordingly. Next, the delegate removes their vote -> The vote is removed, tally is updated in accordance with retracted votes, and crucially retracted votes data persists because it is not yet 0. Finally, the delegate adds their vote back -> The new vote can take into account the retracted votes as it has persisted.

The other two cases are similar and I leave up to the reader.

An important note: Because poll_index -> retracted_votes data can persist when there is no vote, it's important that vote removal by delegators also cleans up the delegate's voting state (when retracted votes is 0 and vote is None), otherwise the delegate has no immediate monetary incentive to clean their voting state as no lock is incurred by that data item. Will add this to document as well.

Both of these are handled in the current WIP. The way I've chosen to handle poll_index -> retracted_votes persistence is by bundling that data with the pre-existing vote data tuple and making the actual vote an Option. [old](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/conviction-voting/src/vote.rs#L232) vs [new](https://github.com/PolkadotDom/polkadot-sdk/blob/dom/vote-while-delegating/substrate/frame/conviction-voting/src/vote.rs#L191-L200). Then the logic just checks that the vote is None and the retracted_votes is 0 before fully removing a vote. 

Edit: RFC is updated

Edit 2: Unit tests are up. The cases pointed out by Basti above can be found [here](https://github.com/PolkadotDom/polkadot-sdk/blob/dom/vote-while-delegating/substrate/frame/conviction-voting/src/tests.rs#L378)

text/0150-voting-while-delegating.md
```diff
@@ -0,0 +1,115 @@
...
+
+`Users`: We will want users to be aware of the new functionality, though not required.
+
+`Technical Writers`: This change will require rewrites of documentation and tutorials. 
+
+## Explanation
+
+### New Data & Runtime Logic
+
+The [Voting Enum](https://github.com/paritytech/polkadot-sdk/blob/939fc198daaf5e8ae319419f112dacbc1ea7aefe/substrate/frame/conviction-voting/src/vote.rs#L256-L264) is first collapsed, as there's no longer a distinction between the variants. Then a `(poll index -> retracted votes count)` data item would be added to the user's voting data stored in [VotingFor](https://github.com/paritytech/polkadot-sdk/blob/939fc198daaf5e8ae319419f112dacbc1ea7aefe/substrate/frame/conviction-voting/src/lib.rs#L165). This would keep track of the per poll balance that has been clawed back from the user by those delegating to them.
```
josepot: > The [Voting Enum](https://github.com/paritytech/polkadot-sdk/blob/939fc198daaf5e8ae319419f112dacbc1ea7aefe/substrate/frame/conviction-voting/src/vote.rs#L256-L264) is first collapsed, as there's no longer a distinction between the variants.

Then how do we know who someone is delegating towards? Wouldn't a new storage entry be needed for keeping track of the delegations?

> Then a `(poll index -> retracted votes count)` data item would be added to the user's voting data stored in [VotingFor](https://github.com/paritytech/polkadot-sdk/blob/939fc198daaf5e8ae319419f112dacbc1ea7aefe/substrate/frame/conviction-voting/src/lib.rs#L165). This would keep track of the per poll balance that has been clawed back from the user by those delegating to them.

I think that it doesn't make sense to have this inside the `VotingFor` data-structure. I think this belongs into the new storage that previously mentioned (the one that defines the delegations). Otherwise, if I'm delegating to Bob and then I cast my vote to ref X before Bob casts their vote to that same ref, then my vote will be computed twice b/c the storage entry for Bob's vote on that ref didn't exist when I voted.

text/0150-voting-while-delegating.md
```diff
@@ -0,0 +1,115 @@
...
+
+The implementation must allow for the `(poll index -> retracted votes)` data to exist even if the user does not currently have a vote for that poll. A simple example that highlights the necessity is as follows: A delegator votes first, then the delegate does. If the delegator is not allowed to create the retracted votes data on the delegate, the tally count would be corrupted when the delegate votes.
```
josepot: > The implementation must allow for the `(poll index -> retracted votes)` data to exist even if the user does not currently have a vote for that poll.

This is IMO too vague. How would that be accomplished without a new storage on this pallet?

text/0150-voting-while-delegating.md
```diff
@@ -0,0 +1,115 @@
...
+
+A new error is to be introduced that signals MaxVotes was reached specifically for the delegate's voting data.
+
+### Locked Balance
+
+A user's locked balance will be the greater of the delegation lock and the voting lock.
+
+### Migrations 
+
+A runtime migration is necessary, though simple considering voting and delegation are currently separate.
```
josepot: I think that the RFC is missing a detailed explanation on how the migration from the old to the new system would take place.

text/0150-voting-while-delegating.md
josepot: Oh, wait, I _think _ that I get it now!

I think that what you are saying is that instead of having an Enum, there would be a struct, where one fields would be `votes`, and the other field would be `delegation`, and the type of the `delegation` would be an `Enum` more or less like this?

```rust
enum Delegation<Balance, AccountId, ReferendumIndex, MaxVotes: Get<u32>> {
  Delegating(Option<(AccountId, Balance, Conviction)>),
  Receiving {
    delegations: Balance,
    clawback: BoundedVec<(ReferendumIndex, Balance), MaxVotes>
  },
}
```

and then the `votes` would probably be something more or less like this?

```rust
struct Votes {
  votes: BoundedVec<(ReferendumIndex, AccountVote<Balance>), MaxVotes>,
  prior: PriorLock<BlockNumber, Balance>,
}
```

Ok, in that case, that makes sense... Also, I think that with this change then we could consolidate the `prior`, so that it's only a property on the `Votes`.

If what you are proposing is something similar to this, then would you mind explaining this a bit better? 🙏

PolkadotDom: Yes looking through it again I do believe it was missing some needed clarity. Thank you for the feedback on this. I've updated the RFC, let me know if it still follows poorly.

And yes! Something like that, where the underlying fields from the enums are consolidated into one struct. Though, in the end, mine looked like this:

```rust
pub struct Voting<Balance, AccountId, BlockNumber, PollIndex, MaxVotes>
where
	MaxVotes: Get<u32>,
{
	/// The current voting data of the account.
	pub votes: BoundedVec<VoteRecord<PollIndex, Balance>, MaxVotes>,
	/// The amount of balance delegated to some voting power.
	pub delegated_balance: Balance,
	/// A possible account to which the voting power is delegating.
	pub maybe_delegate: Option<AccountId>,
	/// The possible conviction with which the voting power is delegating. When this gets
	/// undelegated, the relevant lock begins.
	pub maybe_conviction: Option<Conviction>,
	/// The total amount of delegations that this account has received, post-conviction-weighting.
	pub delegations: Delegations<Balance>,
	/// Any pre-existing locks from past voting/delegating activity.
	pub prior: PriorLock<BlockNumber, Balance>,
}
```

With the retracted votes held in the vote record:

```rust
pub struct VoteRecord<PollIndex, Balance> {
	/// The poll index this information concerns
	pub poll_index: PollIndex,
	/// The vote this account has cast. Can be none if only retracted_votes info is needed
	pub maybe_vote: Option<AccountVote<Balance>>,
	/// The amount of votes retracted from this user for this poll. Can't be more than is
	/// delegated to them. Votes are retracted when a delegator votes in stead of their delegate.
	pub retracted_votes: RetractedVotes<Balance>,
}
```

I would think the final implementation would be more of a PR conversation though.

text/0150-voting-while-delegating.md
PolkadotDom: I believe answered by the conversation above, but it would be held in the new voting data struct. So it would slot in with the pre-existing [VotingFor](https://github.com/paritytech/polkadot-sdk/blob/939fc198daaf5e8ae319419f112dacbc1ea7aefe/substrate/frame/conviction-voting/src/lib.rs#L165) storage item.

text/0150-voting-while-delegating.md
PolkadotDom: Fair 👍  I've updated this section. The migration itself is fairly simple, so still not a lot to say about it, but let me know if the new version is not detailed/clear enough.

PolkadotDom:
@bkchr I had talked about it/called for reviews during the last OpenDev fellowship call. I wrote about it in the fellowship channel a while back, but that would have been a couple months ago now.

text/0150-voting-while-delegating.md
```diff
@@ -0,0 +1,115 @@
...
+
+A new error is to be introduced that signals MaxVotes was reached specifically for the delegate's voting data.
+
+### Locked Balance
+
+A user's locked balance will be the greater of the delegation lock and the voting lock.
+
+### Migrations 
+
+A runtime migration is necessary, though simple considering voting and delegation are currently separate. It would iterate over the [VotingFor](https://github.com/paritytech/polkadot-sdk/blob/939fc198daaf5e8ae319419f112dacbc1ea7aefe/substrate/frame/conviction-voting/src/lib.rs#L165) storage item and convert the [old vote data structure](https://github.com/paritytech/polkadot-sdk/blob/939fc198daaf5e8ae319419f112dacbc1ea7aefe/substrate/frame/conviction-voting/src/vote.rs#L256-L264) to the [new structure](https://github.com/PolkadotDom/polkadot-sdk/blob/dom/vote-while-delegating/substrate/frame/conviction-voting/src/vote.rs#L227-L243).
```
ggwpez: Could you give an estimate of how many items need to be iterated (>1000?)? I would guess that it is more than what we can do in a single Parachain block, so it would probably need to be a Multi-Block-Migration.

ggwpez (approved):
Looks good, thanks for the details.

text/0150-voting-while-delegating.md
PolkadotDom: Thank you for the call out. This is my assumption as well and my current plan. 

However, looking at the voting records of high profile referendums (Chainalysis, hard pressure), it's no more than 500 voters, so it may actually be <1000 in the end. Though of course that's just for the relay.

ggwpez: But that is per referendum? And we need to migrate all referenda at once? Or do you want to migrate them lazily?  
Then it would be 500 * Number of Referenda.

PolkadotDom: Ah okay yes I got it a bit backwards. So the number of items migrated will be Σ(tracks with a current voting record) over 1..number of current voters-- Basically, each voter holds a storage item for each track they're currently voting on. So there's probably ~500 active voters and there's 16 tracks, for a loose upper bound of 8000 (though likely an OOM lower as most tracks are voter sparse).

That being said, each item update will be heavy (both in compute and PoV weight) due to the potentially [MaxVotes](https://github.com/paritytech/polkadot-sdk/blob/939fc198daaf5e8ae319419f112dacbc1ea7aefe/substrate/frame/conviction-voting/src/vote.rs#L229) sized vector held in each one.

Overall I think even if a single block migration did work with a current relay snapshot, we can't guarantee it's soundness for parachain teams, so probably stuck with the MBM.

I hadn't considered a lazy migration 🤔 You mean check for storage format at extrinsic time and update then?

ggwpez: I was just throwing lazy migration into the mix but it is normally not so good since UIs need to fetch and understand both formats.

But as MBM it should be fine, yes.

text/0150-voting-while-delegating.md
```diff
@@ -0,0 +1,115 @@
...
+
+The retracted amount is always the full delegated amount. For example, if Alice delegates 10 UNITS to Bob and then votes with 5 UNITS, the full 10 UNITS is still added as a clawback to Bob for that poll. This is both for simplicity and to ensure we don't make unnecessary assumptions about what Alice wants.
+
+Because you need to add the clawback, a delegator's vote can affect a delegate's voting data. If a delegator's vote or delegation makes the delegate's voting data exceed [MaxVotes](https://github.com/paritytech/polkadot-sdk/blob/939fc198daaf5e8ae319419f112dacbc1ea7aefe/substrate/frame/conviction-voting/src/lib.rs#L138), the transaction will fail. In practice, this means this new system is somewhere between the old and the ideal. However, this will incentivize delegates to stay on top of voting data clearance. And given our current referenda rates and MaxVotes set to [512](https://github.com/polkadot-fellows/runtimes/blob/34ecb949660704ccf139a06afb075c6a729b1295/relay/polkadot/src/governance/mod.rs#L43), it would be difficult to hit this limit.
```
bkchr: > However, this will incentivize delegates to stay on top of voting data clearance

In which way is the delegate incentivized? It could prevent someone from voting for a proposal this way.

bkchr: It is more that teh delegator would be incentivized to call `removeOtherVote`.

text/0150-voting-while-delegating.md
```diff
@@ -0,0 +1,115 @@
...
+
+### New Data & Runtime Logic
+
+The [Voting Enum](https://github.com/paritytech/polkadot-sdk/blob/939fc198daaf5e8ae319419f112dacbc1ea7aefe/substrate/frame/conviction-voting/src/vote.rs#L256-L264), which currently holds the user's vote data, would first be collapsed and it's underlying fields consolidated, as there would no longer be a distinction between the enum's variants. A `(poll index -> retracted votes count)` field would then be added to the resulting structure - It's role to keep track of the per poll balance that has been clawed back from the user by those delegating to them. See [here](https://github.com/PolkadotDom/polkadot-sdk/blob/f9af95133534c18dfde990cb9d775c325c2c6ebf/substrate/frame/conviction-voting/src/vote.rs#L227-L244) for a potential implementation.
+
+The implementation must allow for the `(poll index -> retracted votes)` data to exist even if the user does not currently have a vote for that poll. A simple example that highlights the necessity is as follows: A delegator votes first, then the delegate does. If the delegator is not allowed to create the retracted votes data on the delegate, the tally count would be corrupted when the delegate votes.
```
bkchr: This sentence doesn't bring any value to this RFC.

text/0150-voting-while-delegating.md
```diff
@@ -0,0 +1,115 @@
...
+
+`App developers`: Apps like Subsquare and Polkassembly would need to update their user interface logic. They will also need to handle the new error.
+
+`Users`: We will want users to be aware of the new functionality, though not required.
+
+`Technical Writers`: This change will require rewrites of documentation and tutorials. 
+
+## Explanation
+
+### New Data & Runtime Logic
```
bkchr, line 49: Could you please rephrase this entire section. It should explain how the algorithm is working and not how the data structures are looking. It is right now possible to come up with this already, but the RFC is not laying out exactly how it works.

It would be nice to have the logic layed out for the different states, aka delegator voting first, delegate voting first, delegator removing votes etc.

bkchr (commented):
Still reading better, but it would be really nice to get a proper explanation of the algorithm and not just some "guessing" on how it will work.

text/0150-voting-while-delegating.md
PolkadotDom: > It could prevent someone from voting for a proposal this way.

Luckily the same property that allows for voting while delegating also allows for undelegating / delegating while having a voting record. So if this attack was attempted, the user could just undelegate. Presumably after a `removeOtherVote` attempt. Really the only person who can lose in this situation is the delegate.

With that in mind, I believe there is sufficient political power gained from being a delegate that the delegate will attempt to retain as many delegators as possible (by keeping their voting record clean).

text/0150-voting-while-delegating.md
PolkadotDom: Okay I've updated the explanation section. Let me know what you think 🙌

PolkadotDom:
There's an open question here revolving around a small cost symmetric grief-

A delegator can delegate to someone, then vote on votes the delegatee is not voting on to maximally fill their vote record state (as much as the incoming referenda rate will allow). Then they take the hit of having their dot locked, while waiting for the delegatee to pay for removal of the vote record, costing the delegatee some money.

This can be solved by simply adding a per user flag 'canDelegatorsVote' which just disallows voting for the delegator and blocks the attack (intuitively it just reverts the logic to a similar state as it currently is).

However, with the actual risk of the attack being somewhere around a few cents per referenda (which is a low count as well). It feels nominal and thus I've opted not to add the extra complexity. But leaving it open to discussion here.

bkchr (approved):

PR #152, opened 2025-07-07
RFC-0152: Decentralized Convex-Preference Coretime Market for Polkadot

labormedia:
This RFC proposes a decentralized market mechanism for allocating Coretime on Polkadot, replacing the existing to date Dutch auction method (RFC17). The proposed model leverages convex preference interactions among agents, eliminating explicit bidding and artificial timing schedules for price determination, bringing desirable properties to the pricing mechanism. This further reassures principled motivations of fairness, transparency, and decentralization in resource allocation.

### DIFF
--- a/text/0152-decentralized-convex-preference-coretime-market-for-polkadot.md
+++ b/text/0152-decentralized-convex-preference-coretime-market-for-polkadot.md
@@ -1,0 +1,166 @@
+# RFC-0152: Decentralized Convex-Preference Coretime Market for Polkadot
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 2025-06-30                                                                                  |
+| **Description** | This RFC proposes a decentralized market mechanism for allocating Coretime on Polkadot, replacing the existing Dutch auction method (RFC17). The proposed model leverages convex preference interactions among agents, eliminating explicit bidding and centralized price determination. This ensures fairness, transparency, and decentralization.         |
+| **Conflicts-With| RFC-0017                                                                                    |
+| **Authors**     | Diego Correa Tristain <algoritmia@labormedia.cl>                                            |
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


bkchr:
CC @jonasW3F

jonasW3F:
I've read your proposal but as it stands it's far from being a RFC that can be proposed to the technical fellowship. It lacks both theoretical foundation and practical detail. I’ve tried to fill in some of the many gaps myself and tried to think where these ideas could lead, but I don't think it is our task to do so.

The only concrete mechanism proposed is that agents declare preferences over tokens and Coretime using a Cobb-Douglas utility function. Beyond that, the proposal offers no explanation of how prices are determined, how allocations are computed, or why this method would be superior to a more traditional auction.

I've got some more specific comments:

- Your motivation rests on the claim that bidding in an auction is cognitively complex, yet your mechanism asks agents to specify an α parameter in a Cobb-Douglas utility function. That seems significantly more complex and less intuitive.
- What justifies the assumption that agent preferences can be captured by a Cobb-Douglas function at all? What if agents have different functional forms, discontinuities, or non-convex preferences? Additionally, Cobb-Douglas utility functions are only ordinal and do not tell us much about the intensity of preference, something that makes it yet again more difficult for agents to express them.
- Even if we grant that agents can articulate their preferences in this way, requiring them to publish these preferences publicly invites substantial strategic manipulation. Other agents could easily exploit this information to shape their own declared preferences and manipulate outcomes. Additionally, announcing the preferences before the final allocation would compromise the strategy-proofness of the mechanism, which means that nobody would have an incentive to state their preferences truthfully in the first place.

In general, the quality of this proposal is very low, and I strongly encourage you to invest more effort before asking others to provide feedback. Otherwise, it risks wasting valuable time from people who are here to seriously contribute.

If you genuinely believe in this approach, addressing the comments above should already give you some direction to flesh out your ideas. Additionally, it would help to motivate it with any real-world example where such a price-finding mechanism has been successfully used. None come to my mind, and the proposal does not provide any.

PR #155, opened 2025-09-11
pUSD

xlc:
Polkadot native stablecoin on Asset Hub

[rendered](https://github.com/polkadot-fellows/RFCs/blob/d4dded97db656884ed8e9fea937e2a81d0917af1/text/0155-pUSD.md)

### DIFF
--- a/text/0155-pUSD.md
+++ b/text/0155-pUSD.md
@@ -1,4 +1,4 @@
-# RFC-0155: Polkadot$
+# RFC-0155: pUSD (Polkadot USD over-collateralised debt token)
 
 |                 |                                                                 |
 | --------------- | --------------------------------------------------------------- |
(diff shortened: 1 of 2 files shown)


ggwpez (commented):
Pollar :D? Or too similar to Hollar maybe.

xlc:
I didn't link to existing honzon docs because while a good chunk of the code will be reused, they will be adapted specifically for this. e.g. no more multi collateral. So a lot of information will be incorrect and I fear it only introduce more confusion.

Also it wasn't me decided on honzon, it was Gav's idea. He might be able to share more insights.

Another note, this is basically starting fresh on AssetHub. But just like anyone starting a new projects today, they don't code everything by hand from scratch. A boilerplate or starter project is used. You can view honzon was used as a over-collateralized stablecoin starter and it will be heavily modified to fits the requirements here.

gavofyork:
The code which this stable utilizes is mature and the point of having this RFC process is to avoid "rushed" decisions. That said, I am of the well-considered opinion that Polkadot needs a fully-collateralised decentralised stablecoin, that it needs it ASAP and that it is of strategic importance.

This is a proposal for a DOT native stable, which resides on Hub and for which *issuing liquidity would not result in HDX holders getting a windfall*. Hollar, at least, does not fulfil the last point.

I appreciate that you (and perhaps some others) would desire your own products to have a monopoly within this ecosystem, but I'm afraid I, at least, consider that unacceptable for a strategic service of the Polkadot network.

jak-pan:
Having spoken with Gavin at length on this matter, we will not be proposing an alternative for pUSD - we will continue our focus on collaboration and driving liquidity into the Polkadot ecosystem.

sourabhniyogi:
Mmm, 9/11.  I seriously doubt there can be a repeat of whatever happened to aUSD in pUSD, but:
 (1) people never really got a great post-mortem on aUSD, and without it will be super suspicious of pUSD.  I think its useful to exorcise the ghost of aUSD so pUSD can quickly get as much community love as Hydration enjoys right now.   This might even require _propaganda_ =): "Who controls the past controls the future. Who controls the present controls the past." means ... if you erase or alter what people know about what has already happened, you can steer what they expect and accept in the future.   
 (2) This is a serious political issue.  I think DOT tokenholders should weigh on this, so I posted [OpenGov #1761](https://polkadot.polkassembly.io/referenda/1761) with this RFC verbatim.    
 (3) God Bless USD and God Bless America!

PR #156, opened 2025-10-15
BLS12-381 host functions

s0me0ne-unkn0wn:
[Rendered](https://github.com/s0me0ne-unkn0wn/RFCs/blob/bls/text/0156-bls-signatures.md)

CC @drskalman @davxy

### DIFF
--- a/text/0156-bls-signatures.md
+++ b/text/0156-bls-signatures.md
@@ -69,41 +69,25 @@ Generates a BLS12-381 Proof Of Possession for a given public key and owner ident
 
 The function returns `0` on success. On error, `-1` is returned, and the output buffer should be considered uninitialized.
 
-### ext_crypto_bls381_num_public_keys
+### ext_crypto_bls381_public_keys
 
-Retrieves the number of BLS12-381 keys of the given type available in the keystore.
-
-#### Prototype
-
-```wat
-(func $ext_crypto_bls381_num_public_keys_version_1
- (param $id i32) (result i32))
-```
-
-#### Arguments
-
-* `id` is a pointer ([Definition 215](https://spec.polkadot.network/chap-host-api#defn-runtime-pointer)) to a key type identifier ([Definition 220](https://spec.polkadot.network/chap-host-api#defn-key-type-id)). The function will panic if the identifier is invalid.
-
-#### Result
-
-The result represents a (possibly zero) number of keys of the given type known to the keystore.
-
-### ext_crypto_bls381_public_key
-
-Retrieves a BLS12-381 public key of a given type at a given index from the keystore.
+Retrieves all BLS12-381 public keys of a given type from the keystore.
 
 #### Prototype
 
 ```wat
-(func $ext_crypto_bls381_public_key_version_1
- (param $id i32) (param $index i32) (param $out))
+(func $ext_crypto_bls381_public_keys_version_1
+ (param $id i32) (param $out i64) (result i32))
 ```
 
 #### Arguments
 
 * `id` is a pointer ([Definition 215](https://spec.polkadot.network/chap-host-api#defn-runtime-pointer)) to a key type identifier ([Definition 220](https://spec.polkadot.network/chap-host-api#defn-key-type-id)). The function will panic if the identifier is invalid;
-* `index` is an index of the key in the keystore. If the index is out of bounds (determined by the value returned by the `ext_crypto_bls381_num_public_keys` function), the function will panic;
-* `out` is a pointer ([Definition 215](https://spec.polkadot.network/chap-host-api#defn-runtime-pointer)) to an output buffer, 144 bytes long, where the key will be written.
+* `out` is a pointer-size ([Definition 216](https://spec.polkadot.network/chap-host-api#defn-runtime-pointer-size)) to a buffer where the public keys of the given type known to the keystore will be stored consecutively. The value is actually stored only if the buffer is large enough. Otherwise, the buffer is not written into, and its contents are unchanged.
+
+#### Result
+
+The result is an unsigned 32-bit integer representing the total size in bytes required to store all public keys of the given type. The number of keys can be determined by dividing this value by the known key size (144 bytes for BLS12-381). A value of `0` indicates that no keys of the given type are known to the keystore.
 
 ### ext_crypto_bls381_sign
 
@@ -119,7 +103,7 @@ Signs an input message using a given BLS12-381 key.
 #### Arguments
 
 * `id` is a pointer ([Definition 215](https://spec.polkadot.network/chap-host-api#defn-runtime-pointer)) to a key type identifier ([Definition 220](https://spec.polkadot.network/chap-host-api#defn-key-type-id)). The function will panic if the identifier is invalid;
-* `pub_key` is a pointer ([Definition 215](https://spec.polkadot.network/chap-host-api#defn-runtime-pointer)) to public key bytes (as returned by `ext_crypto_bls381_public_key` function);
+* `pub_key` is a pointer ([Definition 215](https://spec.polkadot.network/chap-host-api#defn-runtime-pointer)) to public key bytes (as returned by `ext_crypto_bls381_public_keys` function);
 * `msg` is a pointer-size ([Definition 216](https://spec.polkadot.network/chap-host-api#defn-runtime-pointer-size)) to a message that is to be signed;
 * `out` is a pointer ([Definition 215](https://spec.polkadot.network/chap-host-api#defn-runtime-pointer)) to an output buffer, 112 bytes long, where the signature will be written.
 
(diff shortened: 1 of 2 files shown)


text/0156-bls-signatures.md
```diff
@@ -0,0 +1,171 @@
...
+#### Arguments
+
+* `id` is a pointer ([Definition 215](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer)) to a key type identifier ([Definition 220](https://polkadotspec.dev/chap-host-api#defn-key-type-id)). The function will panic if the identifier is invalid;
+* `out` is a pointer-size ([Definition 216](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer-size)) to a buffer where the public keys of the given type known to the keystore will be stored consecutively. The value is actually stored only if the buffer is large enough. Otherwise, the buffer is not written into, and its contents are unchanged.
+
+#### Result
+
+The result is an unsigned 32-bit integer representing the total size in bytes required to store all public keys of the given type. The number of keys can be determined by dividing this value by the known key size (177 bytes for ECDSA & BLS12-381). A value of `0` indicates that no keys of the given type are known to the keystore.
+
+### ext_crypto_ecdsa_bls381_sign
```
drskalman: We also agreed for this one also to go. Similar reason with the keccak version. We can get the functionality composing othe host calls.

text/0156-bls-signatures.md
```diff
@@ -0,0 +1,171 @@
...
+* `out` is a pointer ([Definition 215](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer)) to an output buffer, 144 bytes long, where the public key will be written.
+
+### ext_crypto_bls381_generate_proof_of_possession
+
+Generates a BLS12-381 Proof Of Possession for a given public key and owner identifier.
+
+#### Prototype
+
+```wat
+(func $ext_crypto_bls381_generate_proof_of_possession_version_1
```
drskalman, line 57: Because we have bls381_sign now we could implement proof of ownership using sign and then this function can do only proof of possession (in contrast to what is doing now as generating both proof of ownership and proof of possession).  If that is the case then this function does not need get owner as input. Perhaps we should ask for @AlistairStewart 's wisdom on this.

drskalman (changes requested):
One more function needs to go and another possible change which I'm not sure if it is worth it.

text/0156-bls-signatures.md
s0me0ne-unkn0wn: Removed in https://github.com/polkadot-fellows/RFCs/pull/156/commits/16bd1efaecaaa95e20561b0ddf25654839294c50

text/0156-bls-signatures.md
s0me0ne-unkn0wn: I'll be waiting for input on this one, then. CC @P1sar

AlistairStewart: > Because we have bls381_sign now we could implement proof of ownership using sign and then this function can do only proof of possession (in contrast to what is doing now as generating both proof of ownership and proof of possession). If that is the case then this function does not need get owner as input. Perhaps we should ask for @AlistairStewart 's wisdom on this.

I think this should be up to you. So what is the tradeoff? Without an owner, we can generate proofs of possession without proofs of ownership - useful if thee are non-validator uses of BLS. With it, we need to add different code to the runtime to support BLS session keys that is not the same as other crypto.

text/0156-bls-signatures.md
```diff
@@ -0,0 +1,148 @@
...
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 2025-10-15                                                                                  |
+| **Description** | Introduce BLS12-381 host functions                                                          |
+| **Author**      | Someone Unknown                                                                             |
+
+## Summary
+
+Introduce new host functions allowing runtimes to generate BLS12-381 keys, signatures, and proofs of possession.
```
drskalman:
```suggestion
Introduce new host functions allowing runtimes to generate BLS12-381 keys, signatures, and proofs of possession. Additionally it allows runtime to generate paired (ECDSA,BLS12-381) keys to enable BEEFY to support both signatures at the same time.
```

text/0156-bls-signatures.md
```diff
@@ -0,0 +1,148 @@
...
+```wat
+(func $ext_crypto_bls381_generate_proof_of_possession_version_1
+ (param $id i32) (param $pub_key i32) (param $out i32) (result i64))
+```
+
+#### Arguments
+
+* `id` is a pointer ([Definition 215](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer)) to a key type identifier ([Definition 220](https://polkadotspec.dev/chap-host-api#defn-key-type-id)). The function will panic if the identifier is invalid;
+* `pub_key` is a pointer ([Definition 215](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer)) to a public key, 144 bytes long;
+* `out` is a pointer ([Definition 215](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer)) to an output buffer, 224 bytes long, where the proof of possession will be written.
```
drskalman:
```suggestion
* `out` is a pointer ([Definition 215](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer)) to an output buffer, 112 bytes long, where the proof of possession will be written.
```

text/0156-bls-signatures.md
```diff
@@ -0,0 +1,148 @@
...
+* `msg` is a pointer-size ([Definition 216](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer-size)) to a message that is to be signed;
+* `out` is a pointer ([Definition 215](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer)) to an output buffer, 112 bytes long, where the signature will be written.
+
+#### Result
+
+The function returns `0` on success. On error, `-1` is returned, and the output buffer should be considered uninitialized.
+
+### ext_crypto_ecdsa_bls381_generate
+
+Generates a combination ECDSA & BLS12-381 key for a given key type using an optional `seed`, stores it in the keystore, and returns the corresponding public key.
```
drskalman:
```suggestion
Generates a pair of (ECDSA,BLS12-381) keys for a given key type using an optional `seed`, stores it in the keystore, and returns the corresponding public key. The keystore and the runtime regard this pair as a single key.
```

text/0156-bls-signatures.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+#### Arguments
+
+* `id` is a pointer ([Definition 215](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer)) to a key type identifier ([Definition 220](https://polkadotspec.dev/chap-host-api#defn-key-type-id)). The function will panic if the identifier is invalid;
+* `out` is a pointer-size ([Definition 216](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer-size)) to a buffer where the public keys of the given type known to the keystore will be stored consecutively. The value is actually stored only if the buffer is large enough. Otherwise, the buffer is not written into, and its contents are unchanged.
+
+#### Result
+
+The result is an unsigned 32-bit integer representing the total size in bytes required to store all public keys of the given type. The number of keys can be determined by dividing this value by the known key size (177 bytes for ECDSA & BLS12-381). A value of `0` indicates that no keys of the given type are known to the keystore.
```
drskalman:
```suggestion
The result is an unsigned 32-bit integer representing the total size in bytes required to store all public keys of the given type. The number of keys can be determined by dividing this value by the known key size (177 bytes for (ECDSA,BLS12-381)). A value of `0` indicates that no keys of the given type are known to the keystore.
```

text/0156-bls-signatures.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### Arguments
+
+* `id` is a pointer ([Definition 215](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer)) to a key type identifier ([Definition 220](https://polkadotspec.dev/chap-host-api#defn-key-type-id)). The function will panic if the identifier is invalid;
+* `seed` is a pointer-size ([Definition 216](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer-size)) to a SCALE-encoded Option value ([Definition 200](https://polkadotspec.dev/id-cryptography-encoding#defn-option-type)) containing a BIP-39 seed which must be valid UTF-8. The function will panic if the seed is not a valid UTF-8;
+* `out` is a pointer ([Definition 215](https://polkadotspec.dev/chap-host-api#defn-runtime-pointer)) to an output buffer, 177 bytes long, where the public key will be written.
+
+### ext_crypto_ecdsa_bls381_public_keys
+
+Retrieves all ECDSA & BLS12-381 public keys of a given type from the keystore.
```
drskalman:
```suggestion
Retrieves all paired (ECDSA,BLS12-381) public keys of a given type from the keystore.
```

drskalman (changes requested):

drskalman (approved):
LGTM 👍 Thank you :-)

AlistairStewart (approved):
This looks ok.

acatangiu:
On-chain RFC proposal voting live: https://collectives.subsquare.io/fellowship/referenda/500

s0me0ne-unkn0wn:
/rfc process 0xb4aab1c3d443a103869f4bf8358c01e8c401dfa9d4f2e4628dd830d24ddf5bc9

PR #163, opened 2026-01-23
RFC 0163: EC Host Functions

davxy:
Proposal for a set of host functions for performing computationally intensive elliptic curve operations in Polkadot runtimes. These host functions enable efficient execution of cryptographic primitives that would otherwise be significantly expensive when executed in the runtime.

[Rendered](https://github.com/polkadot-fellows/RFCs/blob/main/text/0163-ec-host-functions.md)

[Implementation](https://github.com/paritytech/polkadot-sdk/pull/10830)


text/0163-ec-host-functions.md
```diff
@@ -0,0 +1,294 @@
...
+Affine representation has been chosen for:
+- Simplicity and interoperability with external systems and other libraries
+- Reduced serialization overhead compared to projective coordinates
+- Direct usability without requiring coordinate conversion after the operation
+
+##### Return Values
+
+All host functions return `Result<Vec<u8>, ()>`, where:
+- On success, the `Ok` variant contains the result encoded using the ArkScale codec as described above
+- On error, the `Err` variant contains an empty unit type `()`
```
ggwpez: Why not define an error enum? I assume there could be some error like decoding.

text/0163-ec-host-functions.md
```diff
@@ -0,0 +1,294 @@
...
+## Drawbacks
+
+1. **Maintenance Overhead**: Each curve adds multiple host functions, increasing the maintenance
+   burden and expanding the attack surface.
+2. **Upgrade Coordination**: Changes to host functions require coordinated runtime and node upgrades.
+3. **Library Dependency**: The implementation relies on Arkworks, which has not undergone formal
+   security audits.
+4. **Curve Selection**: Future cryptographic advances may deprecate some curves or require new ones,
+   potentially leading to API churn.
+
```
ggwpez, line 359: 5. Slower than "fat" host functions like `verify_bls_signature` that do the work in one host-call and possibly use multiple threads.

text/0163-ec-host-functions.md
davxy: Yeah. There are currently 3 possible error variants:
- decode
- msm args length mismatch
- multi miller loop args length mismatch

I'll add the error enum

text/0163-ec-host-functions.md
```diff
@@ -0,0 +1,294 @@
...
+This RFC proposes a set of host functions for performing computationally intensive elliptic curve
+operations in Polkadot runtimes. These host functions enable efficient execution of cryptographic
+primitives that would otherwise be significantly expensive when executed in the runtime.
+
+The proposal covers the following elliptic curves:
+- **BLS12-381**: Pairing-friendly curve widely used for BLS signatures and zkSNARKs
+- **Ed-on-BLS12-381-Bandersnatch**: Twisted Edwards curve for in-circuit operations within BLS12-381 zkSNARKs
+
+**TODO**: Which additional curves should be included in this first iteration (e.g. Ed25519, BLS12-377, etc)?
```
drskalman: Quoting @@AlistairStewart we need Pastas curves for variety of reasons in order to use Curve Tree for RingProof in order to scale ring size to O(10^6). It also open us to use a considerable amount of well implemented audited zk code from Zcash. I think @coax1d has already implemented those host call for both curves. I'm not sure where they are.

davxy: They've been added to [arkworks-extensions](https://github.com/paritytech/arkworks-extensions/tree/main/curves) crate. I can add these to the hostcalls

text/0163-ec-host-functions.md
```diff
@@ -0,0 +1,294 @@
...
+  - Takes an affine point G and an unbounded big integer scalar I.
+  - Returns an element of G.
+
+For pairing-friendly curves with distinct G1 and G2 groups, `msm` and `mul` are provided separately
+for each group (e.g., `msm_g1`, `msm_g2`).
+
+### Curve Specifications
+
+**TODO**: Finalize the set of curves to include based on ecosystem needs.
+**TODO**: Should Polkadot and Kusama expose different curve sets? Consider exposing a broader set on Kusama for experimentation.
```
ggwpez: I think its fine to keep them the same for now. Kusama OpenGov could request more - if needed.

text/0163-ec-host-functions.md
ggwpez: Yes lets add them here as well

text/0163-ec-host-functions.md
davxy: https://github.com/polkadot-fellows/RFCs/pull/163/commits/eda2b747b4fb1f560264b968f3a694d2a8ae6fbd

text/0163-ec-host-functions.md
```diff
@@ -0,0 +1,297 @@
...
+Coordinates are scalars representing elements of the curve's base field.
+This applies to both input and output parameters across all host functions.
+Affine representation has been chosen for:
+- Simplicity and interoperability with external systems and other libraries
+- Reduced serialization overhead compared to projective coordinates
+- Direct usability without requiring coordinate conversion after the operation
+
+##### Return Values
+
+All host functions return `Result<Vec<u8>, ()>`, where:
```
koute: Note: this has to be changed to be in style of [RFC-145](https://github.com/polkadot-fellows/RFCs/pull/145), i.e. 1) do not use the host allocator, and 2) detail what exactly happens on the host <=> runtime boundary to serialize the data (including adding WASM WAT-style definitions)

text/0163-ec-host-functions.md
davxy: https://github.com/polkadot-fellows/RFCs/pull/163/commits/067233ab5a38d4037e282fd515887d03726b0e90

text/0163-ec-host-functions.md
davxy: https://github.com/polkadot-fellows/RFCs/pull/163/commits/216c84f8cd1038350151e125fa5ef227004dcbbb

text/0163-ec-host-functions.md
ggwpez: I think it depends on priority on whether we can do 145 first or not. 145 has quite a few other things attached that will probably delay us here if we have to wait for it.

koute: We don't have to wait for 145 if we just design the host calls so that they don't require the host allocator in the first place.

text/0163-ec-host-functions.md
```diff
@@ -0,0 +1,336 @@
...
+- On success, the `Ok` variant contains the result encoded using the ArkScale codec as described above
+- On error, the `Err` variant contains one of the following error codes:
+
+```rust
+enum Error {
+    /// Input data decoding failed.
+    DecodeError = 1,
+    /// Input sequences have different lengths.
+    /// Applies to `msm` and `multi_miller_loop` operations.
+    LengthMismatch = 2,
```
ggwpez: Are all other operations infallible? Otherwise we can add an `Unknown` placeholder variant in case something unexpected pops up in the code.

text/0163-ec-host-functions.md
```diff
@@ -0,0 +1,336 @@
...
+## Summary
+
+This RFC proposes a set of host functions for performing computationally intensive elliptic curve
+operations in Polkadot runtimes. These host functions enable efficient execution of cryptographic
+primitives that would otherwise be significantly expensive when executed in the runtime.
+
+The proposal covers the following elliptic curves:
+- **BLS12-381**: Pairing-friendly curve widely used for BLS signatures and zkSNARKs
+- **Ed-on-BLS12-381-Bandersnatch**: Twisted Edwards curve for in-circuit operations within BLS12-381 zkSNARKs
+
```
ggwpez: Need to mention the Pasta stuff :)

text/0163-ec-host-functions.md
davxy: @koute @ggwpez So in practice we need to use [this](docs.rs/sp-runtime-interface/latest/sp_runtime_interface/pass_by/struct.PassFatPointerAndReadWrite.html)?
I.e. pass the pointer to a WASM-allocated buffer  where the host needs to write the stuff?

So in the implementation like:

FROM:

```rust
	fn bls12_381_multi_miller_loop(
		a: PassFatPointerAndRead<&[u8]>,
		b: PassFatPointerAndRead<&[u8]>,
	) -> AllocateAndReturnByCodec<Result<Vec<u8>, Error>> {
           ...
	}
```

TO:

```rust
	fn bls12_381_multi_miller_loop(
		a: PassFatPointerAndRead<&[u8]>,
		b: PassFatPointerAndRead<&[u8]>,
		out: PassFatPointerAndReadWrite<&mut [u8]>,
	) -> AllocateAndReturnByCodec<Result<(), Error>> {
           ...
	}
```

???

text/0163-ec-host-functions.md
davxy: Yeah, currently all other ops are infallible. But if you like I can add an `Unknown = 255` variant?

text/0163-ec-host-functions.md
koute: There cannot be any use of any strategy with `Allocate` in the name.

If you're going to return an error code (e.g. the `Result<(), Error>` in this case) then this need to be passed as just an integer under the hood.

If you need to return a blob of bytes then you should just pass a pointer to which the host writes.

Implementation-wise use the `PassFatPointerAndWrite` strategy from [this PR](https://github.com/paritytech/polkadot-sdk/pull/8641) (if this RFC is "ready" first then that implementation can be just copy-pasted into a PR for this RFC).

davxy: @koute What about this: https://github.com/paritytech/polkadot-sdk/pull/10969

Note 1: At the moment, only `PassFatPointerAndReadWrite` is implemented in polkadot-sdk, but should be fine I guess...
Note 2: The PR applies the strategy to a single host call for now (bls12-381 multi-miller-loop), just to get feedback before propagate the design to all the others EC hostcalls

davxy: https://github.com/polkadot-fellows/RFCs/pull/163/commits/8ac0c8471b7efe10214e92dfcbea4a07455895a8

text/0163-ec-host-functions.md
```diff
@@ -0,0 +1,343 @@
...
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | 21 January 2026                                                                             |
+| **Description** | Host functions for elliptic curve operations in Polkadot runtimes                           |
+| **Authors**     | Davide Galassi                                                                              |
+
+## Summary
+
+This RFC proposes a set of host functions for performing computationally intensive elliptic curve
+operations in Polkadot runtimes. These host functions enable efficient execution of cryptographic
+primitives that would otherwise be significantly expensive when executed in the runtime.
```
burdges: We propose a family of host functions for Polkadot runtimes that provide native execution for selected operations upon elliptic curve.  

We shall still require that many elliptic curve operations occur within runtime code itself, but here we select 2-4 operations per curve that cover the vast majority of the running time for almost all verifier algorithms in elliptic curve cryptography.

text/0163-ec-host-functions.md
```diff
@@ -0,0 +1,340 @@
...
+All [Arkworks](https://github.com/arkworks-rs) types passed on the runtime/host boundary are serialized
+using [ArkScale](https://github.com/paritytech/ark-scale), a SCALE encoding wrapper for Arkworks types.
+ArkScale internally uses Arkworks' `CanonicalSerialize`/`CanonicalDeserialize` traits, ensuring
+compatibility with the broader Arkworks ecosystem.
+
+The codec is configured with the following settings:
+- **Not validated**: Points are not validated on deserialization for performance (caller responsibility)
+- **Not compressed**: Uncompressed point representation for faster deserialization
+
+##### Scalar Encoding
```
davxy: Add note about diff between ScalarField and BigInteger args

text/0163-ec-host-functions.md
```diff
@@ -0,0 +1,325 @@
...
+
+Detailed benchmark results comparing native host function execution versus pure runtime execution,
+along with practical integration examples (Groth16 verification, Bandersnatch ring-VRF),
+are available in the [Polkadot Arkworks Extensions](https://github.com/davxy/polkadot-arkworks-extensions)
+repository.
+
+### Compatibility
+
+The polkadot-sdk implementation is based on the Arkworks library ecosystem, which is widely used.
+The serialization format is designed for compatibility with other Arkworks-based implementations.
```
davxy: Not only arkworks-based

ggwpez (approved):

PR #164, opened 2026-03-02
RFC 164: Aura: Allow anyone to author blocks

bkchr:


text/0164-aura-allow-anyone-to-author-blocks.md
```diff
@@ -0,0 +1,78 @@
...
+The current pre digest for the `AURA_ENGINE_ID` is just the slot type which is not versionable. So, a new engine ID `aur2` should be introduced that uses an enum as pre digest. To stay compatible with old runtimes,
+we would continue to pass the `slot` via the old `AURA_ENGINE_ID`.
+The block import logic and especially the authority tracking logic needs to be adapted as well to the possibility of anyone building a block. It should follow the same logic as the block production to determine
+the eligible block author. 
+After a block was imported that had the `InactiveOverride` digest included, block production and import should allow any author to produce blocks. The nodes should restrict the number of blocks to one per slot.
+As anyone can build blocks, the node should stop importing blocks at the same height/slot if there are too many blocks. It is then the job of the relay chain to determine on which will be the canonical block.
+The nodes should allow any author until the runtime has announced the next authority list. After the new authority list was announced, Aura continues to produce blocks with the default logic for determining the
```
acatangiu, line 48: So once the "inactivity threshold" is reached for a particular collator set (authority list), then block production becomes free for all until the collator set is renewed (authority list change), right?

The concerns I can think of are 1. potentially many forks, 2. that an "attacker" can race to "outbuild" honest nodes, but this would be costly to do and costly to sustain, at some point some honest node will mine the Whitelist/Upgrade/InvulnerablesChange/etc transaction.

(just starting a conversation here in case others have other concerns)

acatangiu (approved):

text/0164-aura-allow-anyone-to-author-blocks.md
bkchr: > 1. potentially many forks,

Yes this could happen, but I hope this "escape hatch" will never be used. This is really just for some hopefully never happening event. And if this happens, blocks will still be produced. It then depends on the number of nodes actually wanting to collate. This is basically the old "relay chain provided" consensus that we used in the early days for parachains.

> 2. that an "attacker" can race to "outbuild" honest nodes, but this would be costly to do and costly to sustain, at some point some honest node will mine the Whitelist/Upgrade/InvulnerablesChange/etc transaction.

There is always the possibility that you "force include" a transaction by sending a XCM message from the relay chain or any other parachain. The collator is forced to include it or the candidate would be invalid.

text/0164-aura-allow-anyone-to-author-blocks.md
```diff
@@ -0,0 +1,78 @@
...
+authority that isn't on the authority list to claim a slot and thus, build a block it will need to set some special new digest: `AuraDigestItem::InactiveOverride`. This digest should be passed
+alongside the `slot` as a pre digest to the runtime.
+The current pre digest for the `AURA_ENGINE_ID` is just the slot type which is not versionable. So, a new engine ID `aur2` should be introduced that uses an enum as pre digest. To stay compatible with old runtimes,
+we would continue to pass the `slot` via the old `AURA_ENGINE_ID`.
+The block import logic and especially the authority tracking logic needs to be adapted as well to the possibility of anyone building a block. It should follow the same logic as the block production to determine
+the eligible block author. 
+After a block was imported that had the `InactiveOverride` digest included, block production and import should allow any author to produce blocks. The nodes should restrict the number of blocks to one per slot.
+As anyone can build blocks, the node should stop importing blocks at the same height/slot if there are too many blocks. It is then the job of the relay chain to determine on which will be the canonical block.
```
davxy, line 47: > It is then the job of the relay chain to determine on which will be the canonical block

I'm not sure whether it is worth specifying the criteria used by the RC to select the canonical block.
(i.e. if there are N proposals at the same height).

One option is for the InactiveOverride digest to include a VRF signature derived from a well-known input (e.g., the block author's public key plus other deterministic data). The RC would verify the VRF proof and select the proposal with the lowest resulting value.

bkchr: What I meant by this is that the relay chain backers will validate whatever is thrown at them. The candidate with the highest votes wins. This is already how it is implemented today. Nothing changes there.

davxy: Alright, I thought there was a specific dedicated path for these

davxy (approved):

eskimor (approved):
Nice! Really glad we are finally doing this. Love it.

PolkadotDom:
I'm concerned this creates a new incentive structure to attack the collator set. If my expected return from MEV is higher than the cost of a DDoS attack (or other), wouldn't I just always do this?

bkchr:
> I'm concerned this creates a new incentive structure to attack the collator set. If my expected return from MEV is higher than the cost of a DDoS attack (or other), wouldn't I just always do this?

Not sure I understand what you want to say here?

PolkadotDom:
> Not sure I understand what you want to say here?

If I'm a non-block producing collator on AssetHub and my expected gains from producing however many blocks over the session is higher than the cost of crippling the collator set, I would make money by doing so. If I'm understanding correctly this RFC sets the price to be a collator at the cost of taking down the current authority set, no? Perhaps I'm misunderstanding some internals here.

bkchr:
> If I'm understanding correctly this RFC sets the price to be a collator at the cost of taking down the current authority set, no?

If you can take down the collators longer than whatever is the configured time window before the set is opened, yes. But as I see this here just as some sort of lifeline, I could imagine setting the "offline time window" to e.g. 24 hours. So, you would need to keep the other nodes down for 24 hours. We should for sure not set the value to 5min or similar. Maybe you could even not give the new collator any rewards.

PolkadotDom:
> to e.g. 24 hours

May be being a bit of a stickler but I worry with this approach we're just kicking the problem down the road. Perhaps if we blocked all non governance or authority set related transactions but that seems a bit intense. 

To me it seems this RFC bolsters (much needed) defense against collator collusion at the cost of increasing the motive for external attack. I'd currently be preferential to having Parity or someone under contract run a collator until we can switch to a more capital efficient PoP that will scale collators as needed. Then use Claude 6 or whatever to merge any non economically viable system chains into asset hub at that time.

Or honestly if we wanted to use aura2 and an attack on the collator set as a flag for when switching to PoP or PoS is appropriate, maybe that works too 🤷 

Anyway, just my two cents. Thank you for the work.

bkchr:
> May be being a bit of a stickler but I worry with this approach we're just kicking the problem down the road. Perhaps if we blocked all non governance or authority set related transactions but that seems a bit intense.

I can not follow what you are writing here? Why block these transactions?

PolkadotDom:
> Why block these transactions?

(All but those transactions). 

Just an idea to minimize the expected financial gain from joining the collator set through attack, while still keeping the extrinsics we need to fix the situation open- Whichever those are, I haven't thought on it exhaustively. Basically if I can't mine MEV I have no more reason to attack than I have today.

The 'kicking down the road' is just referring to the idea that no matter the duration we set it doesn't change the underlying dynamics of 'is expected gain > cost'

eskimor:
@PolkadotDom not a bad idea in principle - would be a good safe-guard, but I am not sure the complexity is worth it. In principle we should really make sure that it is not feasible to DOS all collator for 24 hours. If that were possible, governance would be at risk as then obviously perfect censorship is possible. We should work on collators being DOS resilient - especially when it comes to block production, this should be realistic. You might be able to swamp the tx pool (also this should be made as hard as possible, via e.g. a large full node network doing the filtering before they arrive at the collators over good rep connections pre-filtered), but block production (with empty blocks worst case), should be easier to protect.

In a way, incentivizing attacks via an MEV possibility might even be net beneficial: If someone showed us a weakness, gaining some profit via MEV, that would still be better than using that weakness to attack governance. (Optimal situation of course not even having that weakness in production in the first place.)

PolkadotDom (approved):

text/0164-aura-allow-anyone-to-author-blocks.md
```diff
@@ -0,0 +1,78 @@
...
+the function `fn inactive_override_slot_diff() -> u32`. A return value of `0` would mean that the functionality is disabled and only the runtime configured authority list is allowed. If the logic allows some
+authority that isn't on the authority list to claim a slot and thus, build a block it will need to set some special new digest: `AuraDigestItem::InactiveOverride`. This digest should be passed
+alongside the `slot` as a pre digest to the runtime.
+The current pre digest for the `AURA_ENGINE_ID` is just the slot type which is not versionable. So, a new engine ID `aur2` should be introduced that uses an enum as pre digest. To stay compatible with old runtimes,
+we would continue to pass the `slot` via the old `AURA_ENGINE_ID`.
+The block import logic and especially the authority tracking logic needs to be adapted as well to the possibility of anyone building a block. It should follow the same logic as the block production to determine
+the eligible block author. 
+After a block was imported that had the `InactiveOverride` digest included, block production and import should allow any author to produce blocks. The nodes should restrict the number of blocks to one per slot.
```
skunert, line 46: Probably good enough for an escape hatch, but one thing to consider are new nodes that sync. For example, a newly warp-synced collator might sync to a point after the `InactiveOverride` digest was seen. It would only learn of the digest during gap sync.

Maybe it should not be a one-time signal but every block produced in this mode should include the digest.

PR #165, opened 2026-04-13
Hybrid post-quantum accounts (DRAFT)

burdges:
We've had folks asking about this again,so figured I'd write up a quick draft.  We'll want fewer opinions in the final maybe, but I've confined those to the appendix. 

We cannot deploy anything here anyways until NIST standardises FN-DSA (Falson), because even the key derivation remains unfinalized.  See https://github.com/pornin/rust-fn-dsa/pull/5)

It turns out staying so close to BIP32 created a minor mess in key derivation.  I disliked doing that for-the-record, but folks were worried, maybe about exchanges or something more embedded in the BTC world, so maybe that mistake was useful somehow.  It's anyways easily fixed by having a parallel key derivation tree now.

We do not discuss post-quantum consensus signatures here, likely those involve different constraints, like enforced key rotation instead of hybrid.  We do have post-quantum schemes for Sassafras too, but they remain ugly so no rush there.

As for elves, we already have the first discussion RFC for making that post-quantum up at https://github.com/polkadot-fellows/RFCs/pull/144 but it'll probably change some whenever Alistair takes another closer look.

All this stuff feels less important than the product push imho, but it turns up in the news, and folks shall keep asking, so here is a place to discuss. 

[rendered](https://github.com/burdges/Polkadot-RFCs/blob/pq_accounts/text/0000-pq_accounts.md)

### DIFF
--- a/text/0000-pq_accounts.md
+++ b/text/0000-pq_accounts.md
@@ -1,0 +1,189 @@
+# RFC-0000: Hybrid post-quantum accounts (DRAFT)
+
+|                 |                                                                                             |
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | Date of initial proposal                                                                    |
+| **Description** | Add host calls for post-quantum signatures and support for hybrid post-quantum accounts.    |
+| **Authors**     | Jeff Burdges                                                                                |
+
+## Summary
+
+After NIST standardizes FN-DSA (Falcon-512) in 2026 or 2027, we shall add one or more host calls for FN-DSA, probably with public key recovery or message recovery (Falcon-MRM), and probably a host call for ML-DSA (Dilithium) too.
+
+AssetHub adds support hybrid post-quantum user accounts that require both Falcon, probably with some recovery, as well as one of ed25519 or sr25519.  We allow hybrid upgrading an existing account by attacking a Falcon key without changing the account's existing elliptic curve key, which does not cont as a transfer and does not require unstaking, unlocking, etc.
+
+Vault adds support for upgrading an existing elliptic curve key to a hybrid post-quantum key, by deriving the Falcon secret key using a selerate parallel purely hard derivation tree, perhaps only if the existing key decends only through hard derivations.  Vault adds support for this hybrid upgrade transaction.
+
+## Motivation
+
+If large enough quantum computers existed, then they could break all the public key cryptography used in most secure network protocols, including Polakdot.  We do not know if reasonable size quantum computer are really possible, or how soon they might be developed, or how they'd scale even after being developed.  We do have reasonable post-quantum KEMs and post-qantum signatures, but they should be used in hybird combinations with elliptic curve KEXs and signatures.
+
+We elaborate upon those observations in the appendix but focus here upon more direct motivations:
+
+1)  People want to feel secure, so if they want to adopt hybrid post-quantum account keys now, and pay say 5% higher tx fees, the why stop them?  In this, hybrid post-quantum account keys would not significantly increase blocksizes or transaction sizes, roughly 1.5 kb per treansation.  If we worried about blocksizes now then we'd adopt NOMT which should improve overall blocksizes by a factor of three or four, dwarfing this.
+
+2)  We could avoid blanace transfers and have users merely add a post-quantum key attached to their existing sr25519 or ed25519 key, without doing any balance transfers or fighitng with staking and locks.  This solves our largest adoption hurdle. 
+
+3)  We do pay development and maintannace costs for doing this properly once, because anything that touches acocunt cryptocraphy requires careful debugging and testing, but we belive that (a) the non-transfer hybrid transition makes roll backs possible, and (b) Kusama reduces these cost, relative to what our compeditors pay.  We doubt that quantum computers disapear from the news since researchers shall continue making some progress, however incremental or overhyped.
+
+We choose FN-DSA (Falcon) because its the smallest easily deployable post-quantum signature scheme:  An FN-DSA signature plus public key winds up being almost 2.5 x smaller than for ML-DSA, but still 16 x larger than for sr25519/ed25519.  FN-DSA and ML-DSA both have similar verificaiton times, roughly half that of sr25519 and ed25519.  FN-DSA signers are roughly 4 x slower than ML-DSA signers, which is already 5 x slower than sr25519/ed25519.  We do not care about signers being somewhat slow however, since our signatures are verified many more times than signed. 
+
+## Stakeholders
+
+<!-- A brief catalogue of the primary stakeholder sets of this RFC, with some description of previous socialization of the proposal. -->
+?? Signal on key or message recovery ??
+
+?? Tor for EC keys as names for PQ keys ??
+
+## Explanation
+
+### Host calls
+
+We add a host call for [FN-DSA-512](https://datatracker.ietf.org/doc/draft-ietf-cose-falcon/) (Falcon) signature verification once its standardized. 
+
+We should explore adding variants that do key or message recovery.  Signal & others have done work on message recovery in [iacr/2026/420](https://eprint.iacr.org/2026/420), but we're unsure how this beats key recovery.  We're unsure if any recovery schemes shall be deployed or stnadardized, or when, or the future code quality, including whether constant-time signers exist.
+
+We suggest a host call for [ML-DSA-44](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.204.pdf) (Dilithium) too, because of its heavy adoption outside the blockchain space, and possible usage in concensus protocol.  We doubt ML-DSA sees much adoption for account signatures in other blockchains though, so blochain related features maybe less likely there too.
+
+We should ideally research if more abstract interfaces make sense for either FN-DSA or ML-DSA.  It'd entail a larger performance hit than https://github.com/polkadot-fellows/RFCs/pull/163 though, unless the wasm boundary handles this smoothly.
+
+### Secret key derivation
+
+We require deterministic key derivation for all account keys, so that users can save their keys in human readable ways, and have key derivation trees.  We cannot yet deterministically derive secret keys for FN-DSA (Falcon) because [NIST has not yet standardized the secret key format and generation](https://github.com/pornin/rust-fn-dsa/pull/5) and indeed other details could change in Falcon, but we expect this should be corrected before 2028.  We should not deploy on polkadot until we know the standards situation, even if we choose non-standard modes like recoveries, but we could deploy other experemental versions to Kusama, provided we accepted the risks of doing roll backs.
+
+As a hybrid post-quantum signature using derivation paths, we need a deterministically derived FN-DSA secret key for each possible derivaiton of the existing elliptic curve key.
+
+We cannot derive a post-quantum key from the scalar in an (expanded) secret key for an elliptic curve signature, because the quantum comouter could break the soft parent, and then derive the secret child.  In principle, a "truly" hard derivation path constructed from hashes alone works fine here, but this never exists in practice.  In schnorrkel for example, we sadly accepted the poor design choice of allowing hard derivations based upon soft derivations, which prompted using the scalar for all derivations in `schnorrkel::{SecretKey,Keypair,MiniSecretKey}::hard_derive_mini_secret_key`.  As I recall, BIP32 works this way too, so a couple stakeholders requested the hard-soft derivations of BIP32, instead of true hard derivations. 
+
+As a result, we need an entirely seperate true hard derivation path that uses only hash functions.  We could explore some hybrid combiners for the derivation path that intermix the elliptic curve key somehow, but afaik this brings no advantage over a seperate parallel purely hard derivation tree.  As soft derivations loose their privacy against quantum computers, we could simply abandon them and not derive FN-DSA secret keys below soft derivations.  If we do derive below soft derivations, then we retain soft derivation anonymity only against classical adversaries.
+
+It's possible someone constructs a soft derivation for FN-DSA eventually, but if so they would necessarily entwine the FN-DSA and elliptic curve for hybrid post-quantum unlinkability, so we'd require some third hybrid post-quantum soft derivation type.
+
+### Accounts ###
+
+Hybrid post-quantum accounts should be identified by the hash of both their public keys keys together, ala `blake2b("hpq_account",a.fn_dsa_pk,a.sr25519_pk)`.  We should avoid two hybrid post-quantum accounts having the same elliptic curve public key, so each hybrid post-quantum account needs a free-ish alias named by its elliptic curve public key.  We could optimise this off-chain at RFC nodes for inocming payments, but we'll likely need some on-chain implementation too.
+
+A quantum adversary could register their own FN-DSA key for any unhybridized accounts, but this concern could mostly be ignored (see Appendix).  We do however have a problem that hackers could hybid an existing unhybrided account, which makes our past rescue techniques useless.  If this borthers us, then we could add some waiting period before hybridizing takes effect, so then a hacked user could keep unhybridizing.
+
+At some point in the future, we could require some STARK proof that the FN-DSA and sr25519 derivation paths agree.  We caution aginst this effort now for several reasons, like development being expensive, and the hybrid transition proofs being massive.  Also, there maybe better options in the 
+non-imminent future but still long before quantum computers appear, like entwined hybrid soft derivations, lattice only protocols, or simply changes in how the community views unclaimed DOTs. 
+
+### Batch verification for sr25519 and ed25519
+
+We could expand https://github.com/polkadot-fellows/RFCs/pull/163 to ed25519-dalek, including some ark-ed25519-dalek shim, and with this add batch verification for sr25519 and ed25519.  If done well, this reduces our current CPU overhead by roughly what 100% hybrid post-quantum accounts costs.
+
+## Drawbacks
+
+We necessarily spend considerably more bandwidth sending the post-qauntum public keys and signatures (see performance below). 
+
+ECDSA key recovery wound up being an incorrect design choice for blockchains like Ethereum, because it obstructs batch verification.  ECDSA key recovery only saves 32 bytes per signature, while batch verification yields massive savings in CPU time, given chains wind up bound by the CPU time of signature verificaiton (cite ourselves?). 
+
+We think key or message recovery sound safe for Falcon for several reasons:  First, batch verification seems much harder and less rewarding for lattice based signatures like FN-DSA nd ML-DSA.  Second, lattice based signatures already have fast verification, usually like half of Ed25519.  in fact, if we add batch verification for sr25519, then FN-DSA plus sr25519 should cost roughly the CPU time of our unbatched sr25519 or ed25519 today.  Third, Signal's message recovery saves roughly 226 bytes, over 7 x more than ECDSA key recovery saves.  
+
+Also conversely, we've care little about saving 226 bytes per signature when Polkadot has not even adopted NOMT yet, which would reduce almost parablock size by a factor of four.  We might prefer to adhere more strictly to the NIST standard, which perhaps grants some advantages, like maybe FIPS or maybe FIPS demands ML-DSA.
+
+## Testing, Security, and Privacy
+
+There is little reason to directly support post-quantum for accounts that do not use air gapped hardware wallets, out of which we only countrol Vault. 
+
+There are high debugging and testing costs for anything that touches account cryptography.  If however we make some mistake, then we could roll back all the non-transfer upgrades into hybrid post-quantum accounts, and then correct the situation.  We'd require roll back support in Vault for this thought, which sounds dangerous ala downgrade attack, but more likely represents how Vault would work anyways.  If we require multiple Vault upgrades to allow and then disallow downgrades, then we might simply keep this on Kusama for longer.  And anyways Kusama should help us do this sanely by any measure.
+
+We think [NIST level 1 schemes](https://postquantum.com/quantum-security-pqc/post-quantum/nist-pqc-security-categories/) suffice because higher level schemes represent much more powerful quantum adversaries.  
+
+There is no privacy in regular transperent blockchain accounts and this does nothing to change that.
+
+## Performance, Ergonomics, and Compatibility
+
+*Performance:*  We necessarily spend considerably more bandwidth sending the post-qauntum public keys and signatures.  
+
+[FN-DSA-512](https://datatracker.ietf.org/doc/draft-ietf-cose-falcon/) (Falcon) has 666 bye signature (21 x ed25519) and 867 byte public keys (27 x ed25519), and 1281 byte private keys.  In concret terms, a Falcon public key and signature together with recovery should costs roughly the same bandwidth as 1.5 NOMT state reads into 1 billion accounts.  It's anyways vastly less than one state read using our current inefficent storage.
+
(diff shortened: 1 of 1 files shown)


text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Drawbacks
+
+We necessarily spend considerably more bandwidth sending the post-qauntum public keys and signatures (see performance below). 
+
+ECDSA key recovery wound up being an incorrect design choice for blockchains like Ethereum, because it obstructs batch verification.  ECDSA key recovery only saves 32 bytes per signature, while batch verification yields massive savings in CPU time, given chains wind up bound by the CPU time of signature verificaiton (cite ourselves?). 
+
+We think key or message recovery sound safe for Falcon for several reasons:  First, batch verification seems much harder and less rewarding for lattice based signatures like FN-DSA nd ML-DSA.  Second, lattice based signatures already have fast verification, usually like half of Ed25519.  in fact, if we add batch verification for sr25519, then FN-DSA plus sr25519 should cost roughly the CPU time of our unbatched sr25519 or ed25519 today.  Third, Signal's message recovery saves roughly 226 bytes, over 7 x more than ECDSA key recovery saves.
```
PolkadotDom, line 81: The 'First..' sentence was confusing I thought 🤔 What's stated there is being backed up by the 'Second..' and 'Third..' sentences. May be less confusing if the 'less rewarding' is moved out to the initial statement that's then followed by the proof in the three following reasons.

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+We think key or message recovery sound safe for Falcon for several reasons:  First, batch verification seems much harder and less rewarding for lattice based signatures like FN-DSA nd ML-DSA.  Second, lattice based signatures already have fast verification, usually like half of Ed25519.  in fact, if we add batch verification for sr25519, then FN-DSA plus sr25519 should cost roughly the CPU time of our unbatched sr25519 or ed25519 today.  Third, Signal's message recovery saves roughly 226 bytes, over 7 x more than ECDSA key recovery saves.  
+
+Also conversely, we've care little about saving 226 bytes per signature when Polkadot has not even adopted NOMT yet, which would reduce almost parablock size by a factor of four.  We might prefer to adhere more strictly to the NIST standard, which perhaps grants some advantages, like maybe FIPS or maybe FIPS demands ML-DSA.
+
+## Testing, Security, and Privacy
+
+There is little reason to directly support post-quantum for accounts that do not use air gapped hardware wallets, out of which we only countrol Vault.
```
PolkadotDom, line 87: Wouldn't mind a short reasoning trace for this for those less familiar with security. Not immediately obvious why this is the case.

burdges: We anyways need a parallel hard derivation tree for post-quantum keys, because BIP32 makes even hard derivations vulnerable.  We only control the derivation paths used by vault.  

Ledger could copy or alter whatever we do in Vault, using the same on-chain infrastructure.  We cannot tell them what to do though.

Afaik only air gaped keys hold balances for decades, so no need to complicate life for niche use cases that maybe use soft derivations, like store fronts.

text/0000-pq_accounts.md
burdges: These three reasons are independent and largely unrelated.  Lattice batch verification would save less, even if one found some sane method.  This has nothing to do with lattices being faster in the first place.

burdges:
TODO:  Cite https://eprint.iacr.org/2026/1300 or other threshold Falcon schemes once we understand their constraints, like if they really handle so many signers securely.

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+*Known:*  If large enough quantum computers existed, then they could break all the public key cryptography used in most secure network protocols, including Polakdot.  
+
+An end-to-end encrypted (e2ee) messnager like Signal must adopt hybrid post-quantum cryptography now-ish, because some adversaries record and store traffic, and they might decrypt these messages in the future once quantum computer arises (and after their nation has become a fascist dystopia).  We do not have this harvest-now-decrypt-later problem for signatures, which includes the public key cryptography used in blockchains like Polakdot.  We could adopt hybrid post-quantum cryptography much more slowly, but adoption by accound would be slow too.
```
rcarback, line 175: Signal launched PQXDH last year: https://signal.org/blog/spqr/

Small typo: "accound" -> "account"

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Future Directions and Related Material
+
+Describe future work which could be enabled by this RFC, if it were accepted, as well as related RFCs. This is a place to brain-dump and explore possibilities, which themselves may become their own RFCs.
+
+## Appendex:  Motivation Backstory
+
+*Unknowns:*  We do not know if quantum computer are really possible, or how soon they might be developed, or how they'd scale even after being developed.
+
+We have only a few fringe superdeterminism theories of quantum physics that predict that quantum computers might be fundementally impossible, like [Gerard 't Hooft](https://arxiv.org/abs/1405.1548), so physicists would typically expected them to be possible.  We do have small quantum computers or something similar, but we lack any demonstration of ["quantum supremacy"](https://en.wikipedia.org/wiki/Quantum_supremacy) yet, aka some demonstration that quantum computers can really be faster.  See [Scott Aaronson's Quantum Supremacy FAQ!](https://scottaaronson.blog/?p=4317).
```
rcarback, line 163: Just after you posted this, Scott Aaronson began sounding the alarm (2qb fidelity results on neutral atom are now well within range of closing the correlated-noise objections which was the key argument against scale): 

https://scottaaronson.blog/?p=9718

There are arguably already 5 examples of "quantum supremacy" from 2025:

1. Google Willow from October 2025, where 65 qbits were used for measuring out-of-time-order correlators. The paper claims it is 13,000x faster than the Frontier supercomputer: roughly 2 hours against an estimated 3.2 years. They also computed hydrogen-atom distances in 15- and 28-atom molecules and checked them against NMR data. I emphasize that the results here are peer-reviewed, gate-based, and validated against something other than a second quantum computer: https://www.nature.com/articles/s41586-025-09526-6
2. D-Wave, March 2025 using spin-glass dynamics. The Advantage2 annealer simulated time evolution of spin glasses across 2D, 3D, and infinite-dimensional lattices. Claimed roughly 20 minutes against one million years on Frontier. There are a couple of publications that contest this result, but they do not actually produce algorithms and results on the same problems and none of them were able to produce results within the 20 mins D-Wave demonstrated. While this is an annealer, not gate-based, and the "one million years" figure is probably overstated, no one has yet beaten the result: https://www.science.org/doi/10.1126/science.ado6285
3. USTC, March 2025 using the Zuchongzhi 3.0. Using 83 qubits with random circuit sampling. They claim it is 10^15 times faster than Frontier and 10^6 times faster than Sycamore. This result is uncontested so far, but random circuit sampling is a benchmark rather than a useful computation: https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.134.090601
4. USTC, August 2025 using the Jiuzhang 4.0. They used gaussian boson sampling with 3,050 photon detection events in 25.6 microseconds. Running the best known matrix product state algorithm on El Capitan would take more than 10^42 years just to build the required tensor network. Like the annealer result, also not a general-purpose gate-based system: https://arxiv.org/abs/2508.09092
5. Quantinuum and UT Austin, September 2025 by a group associated with Scott Aaronson using a one-way communication complexity approach. They used 12 qubits on the H1-1 trapped-ion system performing a task whose most space-efficient classical algorithm provably needs between 62 and 382 bits of memory. This is a memory result rather than speed, but I think relevant: https://arxiv.org/abs/2509.07255

I think the willow result is probably the strongest but the D-Wave result is a close second given that you can potentially run optimization problems with it (if they have rugged topology and fit in the annealer, a big if but it means you can use it for things today, and Anduril claims publicly that they do.

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+Do we care about FIPS compliance? 
+
+FN-DSA:  How close is key generation to standardization?  How soon could Falcon be deployed even without deterministic keys?
+
+Recovery:  Is Signal going to deploy message recovery?  Is it oging to become standard?  How would we integrate message recovery?  How much more does message recovery save vs key recovery?
+
+ML-DSA:  Do we need an ML-DSA host call?  Accessible in contracts?  Are other chains deploying ML-DSA now?  Bridges?
+
+Should we care about STARKs that proove key derivations?  Imho this seems silly.
```
rcarback, line 153: proove -> prove

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Explanation
+
+### Host calls
+
+We add a host call for [FN-DSA-512](https://datatracker.ietf.org/doc/draft-ietf-cose-falcon/) (Falcon) signature verification once its standardized. 
+
+We should explore adding variants that do key or message recovery.  Signal & others have done work on message recovery in [iacr/2026/420](https://eprint.iacr.org/2026/420), but we're unsure how this beats key recovery.  We're unsure if any recovery schemes shall be deployed or stnadardized, or when, or the future code quality, including whether constant-time signers exist.
+
+We suggest a host call for [ML-DSA-44](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.204.pdf) (Dilithium) too, because of its heavy adoption outside the blockchain space, and possible usage in concensus protocol.  We doubt ML-DSA sees much adoption for account signatures in other blockchains though, so blochain related features maybe less likely there too.
```
rcarback, line 46: concensus -> consensus 

Also, isn't ML-DSA-44 Level 2?

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+| --------------- | ------------------------------------------------------------------------------------------- |
+| **Start Date**  | Date of initial proposal                                                                    |
+| **Description** | Add host calls for post-quantum signatures and support for hybrid post-quantum accounts.    |
+| **Authors**     | Jeff Burdges                                                                                |
+
+## Summary
+
+After NIST standardizes FN-DSA (Falcon-512) in 2026 or 2027, we shall add one or more host calls for FN-DSA, probably with public key recovery or message recovery (Falcon-MRM), and probably a host call for ML-DSA (Dilithium) too.
+
+AssetHub adds support hybrid post-quantum user accounts that require both Falcon, probably with some recovery, as well as one of ed25519 or sr25519.  We allow hybrid upgrading an existing account by attacking a Falcon key without changing the account's existing elliptic curve key, which does not cont as a transfer and does not require unstaking, unlocking, etc.
```
rcarback, line 13: attacking -> attaching?

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+Vault adds support for upgrading an existing elliptic curve key to a hybrid post-quantum key, by deriving the Falcon secret key using a selerate parallel purely hard derivation tree, perhaps only if the existing key decends only through hard derivations.  Vault adds support for this hybrid upgrade transaction.
```
rcarback, line 15: `selerate` -> `separate`?

There's a "sleeve" construction where you can generate a PQ secret and then use that as proof you generated a classical secret: https://github.com/xxfoundation/sleeve

That is useful in a post-hack recovery case, not the proposed upgrade, but might be useful as new default behavior for vault. 

That idea is also interesting and perhaps applicable if you could provide a kind of post-hack recovery scenario without pre-generation. I know that's half-baked, but there might be a way to deterministically bind the two keys post generation of the classical key, and that would be neat.

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Motivation
+
+If large enough quantum computers existed, then they could break all the public key cryptography used in most secure network protocols, including Polakdot.  We do not know if reasonable size quantum computer are really possible, or how soon they might be developed, or how they'd scale even after being developed.  We do have reasonable post-quantum KEMs and post-qantum signatures, but they should be used in hybird combinations with elliptic curve KEXs and signatures.
```
rcarback, line 19: I think evidence has recently flipped that they are possible, but when they will be developed is definitely still unknown, but worst case probability is ~5% in **2028** or 10% in **2030**. QuEra, IonQ, and Oratomic are all roadmaping the delivery NISQ systems in the ~10,000 physical qubit range in the next 18months, and only ~1000-1500 "perfect" (logical) qubits are required based on the algorithms, so we enter a gray zone where the physical qubits are present but just not good enough (yet) to run the algorithms at that time. 

I 100% agree with you on hybrid. It is the conservative design choice. An available CRQC diminishes the security of ECDSA to the cost to run the attack (which is not likely to be trivial). At which
point the PQC side protects. Even given the above developments, it will likely be a few years or more even under the most aggressive assumptions before such CRQCs become common place. High quality vetted implementations of the core primitives are necessary, and the impls should exist in some quantity before relying on a pure PQ-only construction.

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+We elaborate upon those observations in the appendix but focus here upon more direct motivations:
+
+1)  People want to feel secure, so if they want to adopt hybrid post-quantum account keys now, and pay say 5% higher tx fees, the why stop them?  In this, hybrid post-quantum account keys would not significantly increase blocksizes or transaction sizes, roughly 1.5 kb per treansation.  If we worried about blocksizes now then we'd adopt NOMT which should improve overall blocksizes by a factor of three or four, dwarfing this.
```
rcarback, line 23: With SQISign you can get down to ~300bytes and with SHRINCS or other stateful methods you can get the per-tx size down to ~1kb albeit there needs to be a full-sized signature (~25-30kb) as a backup. 

(I don't want to advert my work here but if you want references/links to our implementations on some of the above, happy to provide).

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+2)  We could avoid blanace transfers and have users merely add a post-quantum key attached to their existing sr25519 or ed25519 key, without doing any balance transfers or fighitng with staking and locks.  This solves our largest adoption hurdle.
```
rcarback, line 25: blanace -> balance

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+1)  People want to feel secure, so if they want to adopt hybrid post-quantum account keys now, and pay say 5% higher tx fees, the why stop them?  In this, hybrid post-quantum account keys would not significantly increase blocksizes or transaction sizes, roughly 1.5 kb per treansation.  If we worried about blocksizes now then we'd adopt NOMT which should improve overall blocksizes by a factor of three or four, dwarfing this.
+
+2)  We could avoid blanace transfers and have users merely add a post-quantum key attached to their existing sr25519 or ed25519 key, without doing any balance transfers or fighitng with staking and locks.  This solves our largest adoption hurdle. 
+
+3)  We do pay development and maintannace costs for doing this properly once, because anything that touches acocunt cryptocraphy requires careful debugging and testing, but we belive that (a) the non-transfer hybrid transition makes roll backs possible, and (b) Kusama reduces these cost, relative to what our compeditors pay.  We doubt that quantum computers disapear from the news since researchers shall continue making some progress, however incremental or overhyped.
```
rcarback, line 27: maintannace -> maintenance, acocunt -> account, cryptocraphy -> cryptography.

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+We choose FN-DSA (Falcon) because its the smallest easily deployable post-quantum signature scheme:  An FN-DSA signature plus public key winds up being almost 2.5 x smaller than for ML-DSA, but still 16 x larger than for sr25519/ed25519.  FN-DSA and ML-DSA both have similar verificaiton times, roughly half that of sr25519 and ed25519.  FN-DSA signers are roughly 4 x slower than ML-DSA signers, which is already 5 x slower than sr25519/ed25519.  We do not care about signers being somewhat slow however, since our signatures are verified many more times than signed.
```
rcarback, line 29: Regarding the signer time, this is a concern especially for hardware signers. SHRINCS and related methods are the only game in town right now, with sign times sub second on ledger and potentially possible via burner when using SHA-2.

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+We should ideally research if more abstract interfaces make sense for either FN-DSA or ML-DSA.  It'd entail a larger performance hit than https://github.com/polkadot-fellows/RFCs/pull/163 though, unless the wasm boundary handles this smoothly.
+
+### Secret key derivation
+
+We require deterministic key derivation for all account keys, so that users can save their keys in human readable ways, and have key derivation trees.  We cannot yet deterministically derive secret keys for FN-DSA (Falcon) because [NIST has not yet standardized the secret key format and generation](https://github.com/pornin/rust-fn-dsa/pull/5) and indeed other details could change in Falcon, but we expect this should be corrected before 2028.  We should not deploy on polkadot until we know the standards situation, even if we choose non-standard modes like recoveries, but we could deploy other experemental versions to Kusama, provided we accepted the risks of doing roll backs.
```
rcarback, line 52: Deterministic Falcon is vulnerable to a differential fault attack (https://eprint.iacr.org/2023/422 - Bauer & De Santis). Roughly 25 faulty signatures from the same PRNG seed recover the key, and faulty signatures still verify, so verify-before-output does not catch them.

The mitigation (I have a spec here with more details, link on request) is sign-once-and-cache: sign exactly once per (key, nonce) pair and return the cached result thereafter, capping the attacker at one signature. I think this matters most for Vault, since it is an air-gapped system where physical fault injection is the realistic threat.

text/0000-pq_accounts.md
```diff
@@ -0,0 +1,189 @@
...
+
+We think [NIST level 1 schemes](https://postquantum.com/quantum-security-pqc/post-quantum/nist-pqc-security-categories/) suffice because higher level schemes represent much more powerful quantum adversaries.  
+
+There is no privacy in regular transperent blockchain accounts and this does nothing to change that.
+
+## Performance, Ergonomics, and Compatibility
+
+*Performance:*  We necessarily spend considerably more bandwidth sending the post-qauntum public keys and signatures.  
+
+[FN-DSA-512](https://datatracker.ietf.org/doc/draft-ietf-cose-falcon/) (Falcon) has 666 bye signature (21 x ed25519) and 867 byte public keys (27 x ed25519), and 1281 byte private keys.  In concret terms, a Falcon public key and signature together with recovery should costs roughly the same bandwidth as 1.5 NOMT state reads into 1 billion accounts.  It's anyways vastly less than one state read using our current inefficent storage.
```
rcarback, line 99: For the hybrid combiner itself, I recommend:

https://datatracker.ietf.org/doc/draft-ietf-lamps-pq-composite-sigs/

It also provides:

- *Domain separation.* Bind each component signature to a versioned label so a component signature cannot be replayed in another context.
- *Key non-reuse.* Component keys must not be used outside the composite, per the draft's Section 9. This is stronger than the `blake2b` account id proposed here, which binds identity but not per-signature context.

For key size shaving, you can pre-encode length in the first byte. This saves serialization overhead at zero
cryptographic risk. These techniques are independent and compose.

Also:`867` -> `897` IIRC.

rcarback (commented):
This is an excellent overview of the general issues and a good approach. I think signing time should be considered for hardware devices (not just vault impls), and would say to consider SHRINCS or another hash-based stateful approach in addition to this. 

Also, while the RFC targets account signing, it does not cover consensus canonicality. If the design later reaches validators or light-client justifications, it will need deterministic signatures with a verifiable nonce. I have suggestions and a spec doc for this if interested.

text/0000-pq_accounts.md
burdges: Interesting, but this sounds less bad than the existing fault injection attacks on Ed25519, no?  And ML-DSA would've similar fault injection attacks, no?   

Also sr25519 shields against the ed25519 fault injections by being randomized.  I've noticed discussions on using system randomness in FN-DSA and ML-DSA, maybe nobody does this, but we could do this if desired. 

As a rule, fault injection attacks are risky for cloud machines and browsers.  An existing ed25519 account is secure when run on a hardware signing device, like Vault or Ledger (if they supports Ed25519 on polkadot).  We could apply the same reasoning to FN-DSA right?

burdges:
TODOs:  Incorporate the better benchmarks in https://blog.cloudflare.com/ml-dsa-will-have-to-do/ which show ML-DSA being even worse vs FN-DSA (Falcon).  

As I understand it, Falcon has a [constant-time fixed-base signer](https://eprint.iacr.org/2019/893) that's only 2x slower than is floating point signers.  Assuming so, if you have 17 or more verifications then the constant-time fixed-base FN-DSA outperforms ML-DSA in overall CPU time.  See also [2024/1709](https://eprint.iacr.org/2024/1709).

We expect like 40 verifiers from the relay chain plus parachain.  Vault has plenty of spare CPU cycles, so signer time matters little for us, making this 17 less than 1, and FN-DSA is over 40x cheaper in CPU time than ML-DSA.  

Importantly that blog post promotes Cloudflares position that ML-DSA and Hawk are "mainstream" because they have cheaper signers, while FN-DSA is "niche" because it has cheaper verifiers.  This is clearly bullshit.  We typically expect signatures get verified more than signed, because that's what happens in certificates, app stores, git commits, etc.

A correct statement would be that ML-DSA and Hawk are "niche" for having slow verifiers and fast signers, but cloudflare prefers ML-DSA on FN-DSA anyways, because they do one signature per TLS session, and that FN-DSA would cost more CPU time overall if used in their context.

text/0000-pq_accounts.md
burdges: Appears the benchmarks in https://blog.cloudflare.com/ml-dsa-will-have-to-do/ maybe better than what I cited here.  I made a TODO below to integrate them.

Vault has no problem with FN-DSA or anything else, even [isogenies](https://sqisign.org).  lol

I've still not looked closely at SHRINCS, but I I remain worried that stateful too many integration complexities., but..

We could pick a stateful signer security level that exceeds a human's victims capability to operate their offline signer, like say 10,000 signatures for a break.  We then construct an infinite-ish chain of 1 layer XMSS signatures tuned to that security level.  A valid signature by `account = (original_pk, previous_pk, next_pk)` would sign `(msg, next_next_pk, next_pk)` using `previous_pk` and set `account = (original_pk, next_pk, next_next_pk)` on chain.  

An evil RPC node could tell the victim that  `account = (original_pk, previous_pk)` remains on-chain, thereby forcing the victim to sign hundreds of different messages.  Yet, the human victim cannot physically enter their pin the 10,000 times required for a break, so they would write off their funds before compromising their account, and the evil RPC node never wins anything, making this only a griefing attack.

Around limited usage hash based signatures:  https://csrc.nist.gov/pubs/sp/800/230/ipd 

Also stateful would be less problematic if one required hybrid.

burdges:
Appendix TODO:

We should've some remark on Ewin Tang and how some groups like D-Wave and Google's Quantum AI were forced to pivot once [she disproved their whole field.](https://www.reddit.com/r/crypto/comments/1t76s99/at_18_ewin_tang_wrecked_the_field_of_quantum/)

We do not need anything in the RFC about all the startup drama, but there is a long history of that in this field:

https://qbnets.wordpress.com/2019/12/30/scott-aaronson-excoriates-two-quantum-startups-xanadu-and-zapata/

IonQ was maybe the one sending Scott Aaronson threats?  I forget but..

https://scottaaronson.blog/?p=9425 
https://scottaaronson.blog/?p=9344

text/0000-pq_accounts.md
burdges: Afaik annealing should never become relevant here.  The supremacy results are interesting, but not immediately concerning.

PR #166, opened 2026-05-28
RFC-0166: Snowbridge Emergency Pause RFC

claravanstaden:
Proposes a permissionless, deposit-gated emergency halt for Snowbridge. Any account can stop new flow through the bridge by reserving a DOT deposit; the halt lands on BridgeHub, AssetHub, and the Ethereum Gateway, with automatic retry of any leg that fails. Fellowship adjudicates genuine (refund) vs malicious (slash); an on-chain backstop auto-unhalts if Fellowship is incapacitated.

[Rendered](https://github.com/claravanstaden/RFCs/blob/bd46fe14258346d5085c7e98b11b30e975f9a962/text/0166-snowbridge-emergency-pause-pallet.md)


yrong:
From AI review, just FYI.

1. resolve_genuine leaves State::Normal while the bridge remains halted

  The RFC says resolve_genuine refunds and transitions to Normal, but does not resume. Then resume() is callable when State::Normal.

  That means State::Normal does not mean "bridge normal"; it means "pause pallet idle". This is documented in the enum, but it creates UX and safety ambiguity:

  - A second trigger() could be called while Snowbridge is already halted after a genuine incident.
  - The caller could lock 100k DOT for no additional halt effect.
  - Fellowship would then need to classify a trigger that was redundant but maybe well-intentioned.

  I would suggest adding one of:

  - explicit ResolvedGenuineAwaitingResume state, where only resume() / force_extend-like governance actions are allowed; or
  - require trigger() to check actual Snowbridge operating modes and reject if already halted; or
  - explicitly document that repeated triggers during a genuine halt are allowed and define their resolution semantics.

  2. Backstop auto-resume is risky during Fellowship failure

  The RFC acknowledges this, but I think it deserves stronger treatment. If Fellowship is unreachable for 7 days during a real exploit or unresolved incident, automatic resume may reopen the bridge while the vulnerability still exists.

  The current argument is "if Fellowship was unreachable for the full backstop, the trigger was probably bad". That is not necessarily true during a severe governance, infra, or social-coordination incident.

  Alternatives worth discussing:

  - backstop refunds deposit but does not auto-resume;
  - backstop transitions to a state where only a lower-latency governance/root path can resume;
  - require a much longer timeout for auto-resume than for deposit refund;
  - make auto-resume configurable and possibly disabled on mainnet.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,250 @@
...
+
+1. `EthereumInboundQueue::set_operating_mode(Halted)` (BH local), blocks E→P V1 dispatch on Polkadot. Covers inbound-queue exploits and any in-flight V1 message arriving before the Gateway halt lands.
+2. `EthereumInboundQueueV2::set_operating_mode(Halted)` (BH local), same for V2.
+3. `EthereumOutboundQueue::set_operating_mode(Halted)` (BH local), rejects V1 P→E exports when AH's XCM `ExportMessage` lands on BH. Closes the gap between trigger time and the AH frontend halt landing.
+4. `EthereumOutboundQueueV2::set_operating_mode(Halted)` (BH local, new). Same as call 3 but for V2 P→E exports. Closes the symmetric in-flight gap before the Gateway halt lands on Ethereum.
```
vgeddes: It's dangerous to halt these pallets before halting AH P->E XCM, as users could permanently lose funds if their messages are dropped.

I would consider not halting these pallets

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,250 @@
...
+
+Encoding resume on-chain in this pallet (rather than relying solely on an external SDK preimage tool to assemble the seven calls) means the resume logic is versioned and tested alongside the halt logic. If the set of pallets that need flipping ever changes (e.g., a new outbound queue version is added), `trigger()` and `resume()` move together as one pallet upgrade; there is no second artifact to keep in sync.
+
+The pallet has three resume paths, with different policies for who resumes the bridge:
+
+* **`resolve_genuine`**: the pallet refunds the deposit and transitions to `Normal`, but does **not** auto-resume. The trigger was a legitimate incident response, so resuming requires an explicit `resume()` call once the underlying issue is fixed. Fellowship issues `resume()` when ready; this is fast (Fellowship XCM voice, the same path as the resolve calls) and keeps incident-response decisions deliberate.
+* **`resolve_malicious`**: slashes the deposit and transitions directly to `Resuming`. The bridge auto-resumes via `on_initialize` without waiting on a separate `resume()` call. The trigger was bogus, so no separate decision is needed.
```
vgeddes: Simpler and cleaner to have a single `resolve` extrinsic that takes a policy enum as a parameter.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,250 @@
...
+* **Backstop timeout**: refunds the deposit, transitions to `Resuming`, and auto-resumes (same shape as `resolve_malicious` for the resume side). The reasoning: if Fellowship has been unreachable for the full backstop window, the default policy is "the trigger was probably bad", since a real incident would have been resolved well before then. The per-asset Gateway-side velocity caps from the companion preventive layer remain in force regardless and continue to bound value-at-risk after the bridge resumes.
+
+The existing SDK governance resume preimage still works and remains the fallback if the pallet itself is somehow wedged (governance can always submit `set_operating_mode(Normal)` directly to the underlying Snowbridge pallets via root). The `resume()` extrinsic is the fast, on-chain path for the common case; the SDK preimage is the slow, out-of-band escape hatch.
+
+### Pallet structure
```
vgeddes: I don't think it's useful to include actual implementation code in the proposal. It's a distraction. The value proposition and high-level design should stand on their own

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,250 @@
...
+* **`BackstopDuration` value.** Default is 7 days. Open question whether a longer value (e.g. 2-3 weeks) is warranted as additional caution against premature auto-resume under a prolonged real incident, weighed against the longer worst-case downtime in the Fellowship-unreachable case. Set in runtime config; can be tuned without changing the pallet.
+
+## Future Directions and Related Material
+
+* **V2 outbound queue governance bypass (V1-deprecation follow-up).** Call 7 currently routes through V1's `EthereumSystem`, leveraging V1's existing `PRIMARY_GOVERNANCE_CHANNEL` bypass. When V1 is deprecated and removed, call 7 will need to use `EthereumSystemV2` instead, at which point the V2 outbound queue must gain a governance bypass on its send path (since otherwise call 4's V2 halt would block call 7). V2's `Message` type carries `origin: H256` rather than V1's `channel_id`, so the bypass mechanism will need to be new infrastructure, e.g. checking message origin against the runtime's governance origin, or an explicit `is_governance` flag on the V2 `Message`. This work is deferred until V1 deprecation; this RFC explicitly chooses to keep the V2 outbound queue diff minimal today.
```
vgeddes: This point is convoluted and hard to understand while also being a minor implementation detail.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,250 @@
...
+* **XCM integration tests** with a wedged AH HRMP queue, asserting `trigger()` still transitions to `Triggered`, records the AH leg as pending, retries on subsequent blocks, and clears once HRMP drains.
+* **End-to-end simulation** (chopsticks fork): trigger with all seven legs healthy, with one failed leg, with the BH outbound queue at capacity, with AH unreachable. Each scenario should yield "bridge closed via whatever subset landed, remaining legs pending".
+* **Deposit reservation paths** under sufficient/insufficient balance.
+* **Resolution paths** (`resolve_genuine`, `resolve_malicious`, timeout, `force_extend`).
+* **No new privacy surface.** All events are public; the deposit caller is already implicit in the transaction signer.
+* **Security posture:** the pallet creates a new attack surface (anyone with 100k DOT can halt). This is the intended design, calibrated against the asymmetric harm of being unable to halt during an active drainage.
```
vgeddes: It's obvious that all of this will need to be tested in a proposed implementation.

But this is a high level design doc, so listing testing steps seems out of place here, makes this doc come across as a generic  checklist.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Performance, Ergonomics, and Compatibility
```
vgeddes, line 97: This whole section seems unnecessary

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: @alistair-singh made a point that the first step should be to halt all message on Bridge Hub, reason being, that immediately stops any funds from flowing, even if it leaves the bridge in an inconsistent state and we need some migration/rectifying governance proposal later.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: It's mandated as part of the RFC spec template here: https://github.com/polkadot-fellows/RFCs/blob/main/0000-template.md#testing-security-and-privacy

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Again, part of the RFC template: https://github.com/polkadot-fellows/RFCs/blob/main/0000-template.md#performance-ergonomics-and-compatibility

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Added in https://github.com/polkadot-fellows/RFCs/pull/166/commits/34a5c04d86524fa1f3aaaa3b3d23c13ac899563d

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Removed in https://github.com/polkadot-fellows/RFCs/pull/166/commits/896cf61ccc40247f28d291689d120f710ade26a9

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Removed in https://github.com/polkadot-fellows/RFCs/pull/166/changes/40d19b748e518445981ad90a3744f1d84fdf5939

text/0166-snowbridge-emergency-pause-pallet.md
alistair-singh: Yeah so i suggested this as for like an emergency procedure. Like we know we are currently being attacked. In which case we want to stop the bridge now and then recon later to restore funds of valid messages. We have recovered from this kind of error before, replaying transactions to recover user funds.

alistair-singh: Maybe the emergency force stop everything now should require 150,000 DOT and trigger should take in a parameter to specify graceful vs ungraceful bridging stopping.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -44,99 +32,75 @@ A permissionless DOT deposit triggers a complete Snowbridge halt, in response to
...
 
-### What we reuse from prior work
+To prevent censoring the bridge should the Technical Fellowship being unavailable for an extended amount of time, the bridge should autoresume after a set duration, set in the pallet config (suggested around 2 weeks).
 
-**From `pallet-safe-mode` and [PR #1164](https://github.com/polkadot-fellows/runtimes/pull/1164)'s AssetHub wiring:** the deposit / reserve / release pattern, the `EnsureXcm<IsVoiceOfBody>` origin pattern for `ResolveOrigin`, the 100k DOT deposit calibration, the duration / extension machinery (`EnterDuration` + `ExtendDuration` + `on_initialize` deadline check).
+The pallet should also have an extend extrinsic, callable by the Fellowship, to extend the halt by the provided duration.
```
yrong: I'd prefer to remove this. A safety mechanism should fail toward "stay halted," not "reopen on a timer." 

There's already a governance backstop. All these are governance-settable operating modes — that's literally the existing halt/resume path the RFC describes in its Motivation. So OpenGov (Root) can always force-resume. The bridge can't actually be permanently censored as long as OpenGov remains functional.

Auto-resume only adds unique value if both the Fellowship and OpenGov are inactive for an extended period (e.g., 2+ weeks) — at which point the chain has far bigger problems than this bridge, and automatically reopening an unreviewed bridge is not obviously the right move IMHO.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -44,99 +32,75 @@ A permissionless DOT deposit triggers a complete Snowbridge halt, in response to
...
 
-None of the seven calls block the trigger: the pallet attempts each, logs successes and failures, and re-attempts pending legs in later blocks via `on_initialize`.
+- `Immediate`: The immediate option would immediately block all incoming and outgoing bridge traffic on BridgeHub. This might have inconvenient effects on both sides of the bridge - Ethereum transactions might have been submitted and will not be processed on Polkadot, and transactions from other parachains might have to initiated but will not be completed due to the immediate block on BridgeHub. In an active exploit scenario, the upside (blocking all traffic, immediately) outweighs the downside, given that it can be corrected with a follow-up recon proposal.
+- `Gradual`: The gradual option sends the required cross-chain messages to block messages incoming from AssetHub and Ethereum, so that the bridge is halted gracefully, with inflight transactions being allowed to complete before halting the bridge. This case would make more sense to use in a situation where a vulnerability is detected but not actively exploited.
```
yrong: I'm assuming the enum simply selects which set of calls gets dispatched (1–7 for Immediate vs. 6–7 for Gradual), or does it introduce a distinct on-chain control path? I'd prefer to keep it simple — a parameterized call set rather than separate logic — to avoid added complexity.

One concern about escalation, if a caller triggers Gradual and then realizes the incident is an active exploit, can they upgrade to Immediate without waiting for the Fellowship? As written, follow-up calls fail once the pallet is Triggered, which seems to block that escalation. It would be good to allow moving toward a stricter halt (Gradual → Immediate) while still forbidding de-escalation.

Given the above, I'm not sure whether a two-value enum is the right level of granularity. It feels like the awkward middle: too blunt to be precise, yet too sharp to be safe-by-default. Two cleaner options seems to be:
- Always halt everything — simplest, and the safest default.
- A fully target list later — which is effectively the v2 "per-extrinsic granular pause" is already moving toward.

Happy to be corrected if I’m misunderstanding the design.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: I don't have a strong opinion about this, I think not auto resuming is OK, I agree if the fellowship isn't available we have bigger problems and the bridge should likely not resume under those conditions.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: > I'm assuming the enum simply selects which set of calls gets dispatched (1–7 for Immediate vs. 6–7 for Gradual), or does it introduce a distinct on-chain control path? I'd prefer to keep it simple — a parameterized call set rather than separate logic — to avoid added complexity.

This moves complexity to the caller though, I'm not sure that's good either, especially considering the conditions under which this will be called.

> One concern about escalation, if a caller triggers Gradual and then realizes the incident is an active exploit, can they upgrade to Immediate without waiting for the Fellowship? As written, follow-up calls fail once the pallet is Triggered, which seems to block that escalation. It would be good to allow moving toward a stricter halt (Gradual → Immediate) while still forbidding de-escalation.

Good idea, I'll add.

> Given the above, I'm not sure whether a two-value enum is the right level of granularity. It feels like the awkward middle: too blunt to be precise, yet too sharp to be safe-by-default. Two cleaner options seems to be

I don't agree, I think the halt behaviour will generally fall into those two cases and having the pallet dictate the behaviour is better than someone forgetting to add a call and then the halt is partial. I'll mull over this a bit but this is my first thought

text/0166-snowbridge-emergency-pause-pallet.md
alistair-singh: If auto resuming makes the code have greater complexity or attack surface the we should probably not auto-resume.

alistair-singh:
Since we use the MessageQueue pallet for P -> E, and each message cycles through phases Accepted and then Committed. When the bridge is paused we essentially accept messages but never commit them so they queue. I think this is safe because nonces are only issued on committing of a message. At unpause time fresh nonces will be issued in recent blocks so we would not have to worry about stale light clients.

For E -> P we can do the same, if the bridge is paused queue them in the MessageQueue pallet instead of immediately dispatching. This is also safe, even though the nonce is issues on the ethereum side, the message is relayed and then queued meaning we also do not have to worry about stale light client in this direction as well.

When the bridge is unpaused then all messages get dequeued and sent. We could then audit these queues to remove malicious messages via migration before unpausing.

I think this makes things simpler because it removes the need to do a complex multi-chain co-ordinated pause. I mean we would still do the multi-chain pausing anyway, but any messages inflight would safely be queued. So there would be no difference between halting immediately vs gradual. We could also potentially use the pause pallet for this.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Removed auto resume in https://github.com/polkadot-fellows/RFCs/pull/166/commits/ee4749bfd0a6f1884494275acdada39edfb1503c.

claravanstaden:
@alistair-singh this makes sense for P->E. For E->P, it doesn't use the message queue pallet, so it won't work there. But because we would pause the Inbound Queues, maybe that is acceptable?

claravanstaden:
@alistair-singh updated, as discussed.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Summary
+
+At the moment, there is no way for Snowbridge to be halted immediately. The best course of action to halt the bridge should an exploit be detected, is to halt the bridge through a whitelisted caller proposal, through OpenGov. This has obvious drawbacks - even if a Snowbridge exploit is detected, there is no way to halt the bridge on-chain (off-chain relayers can be switched off but it is obviously not a fool-proof stopgap). This RFC proposes a permissionless, instant Snowbridge halt if the caller deposits a large sum of DOT, to be slashed if paused maliciously. This proposal is a reactive security measure (i.e. a exploit or vulnerability first need to be visible for this functionality to be useful). Another proposal, [Snowbridge Circuit Breakers](https://github.com/polkadot-fellows/RFCs/pull/167), is proposed alongside this RFC for a more proactive approach.
+
+## Motivation
+
+Snowbridge has no near-immediate halt path today. Existing governance halt routes require a referendum  and Fellowship action (hours-to-days latency). Both are too slow for an active drainage exploit and to stop activity during investigation.
```
acatangiu:
```suggestion
Snowbridge has no near-immediate halt path today. Existing governance halt routes require a referendum and Fellowship action (hours-to-days latency). Both are too slow for an active drainage exploit and to stop activity during investigation.
```

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,120 @@
...
+
+Investigation into the new TX Pause pallet and Safe Mode pallet ([polkadot-fellows/runtimes PR1164](https://github.com/polkadot-fellows/runtimes/pull/1164)) revealed parts that can be reused and referenced, but they do not resolve Snowbridge's need directly. Pallet Safe Mode blocks all calls on the chain, including unrelated parts of the chain, which could have unintended effects for the rest of the chain. Besides this, Snowbridge requires a multi-chain freeze that spans Ethereum contracts, Bridge Hub and Asset Hub. Neither of these two existing pallets support inter-chain messaging. Similarly, pallet TX Pause requires a privileged origin. Snowbridge requires a permissionless pausing mechanism, given a sizeable, slashable deposit.
+
+## Stakeholders
+
+* **Polkadot Fellowship**, the `ResolveOrigin` and the body that decides between genuine vs malicious triggers. The most likely callers of a permissionless trigger during an incident.
```
acatangiu:
```suggestion
* **Polkadot Fellowship**, the `ResolveOrigin` and the body that decides between genuine vs malicious triggers.
```

When we say "Polkadot Fellowship" in the context of it being the origin of some action, the meaning usually refers to some fellowship on-chain governance track running a referendum and its potential success being executed with the respective fellowship (sub-)origin.

This RFC proposes a permissionless trigger, so "the fellowship" is actually very **unlikely** to be the "caller". The caller will likely be a single individual or a multisig belonging to parties having vested interest in Polkadot's success and thus willing to put up monetary stake and trigger this "Good Samaritan" action. That person or group may or may not be part of the Polkadot Fellowship, but it is ultimately irrelevant. Ergo, I am for removing this section.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Explanation
+
+### Goal
+
+A permissionless DOT deposit triggers a complete Snowbridge halt, in response to possible exploit (stop new activity while investigating) and active exploits (attacker is actively draining value).
+
+### Implementation
+
+The proposed implementation starts with an entry point extrinsic on Bridge Hub (in a new pallet). The extrinsic requires a DOT deposit. Once a valid deposit has been reserved, the pallet state changes to `Halted` and the bridge is halted in both directions. The halt is graceful: messages that were already in flight, in either direction, are held and sent once the bridge resumes rather than being lost (see below). Once in the `Halted` state, follow-up calls to the same extrinsic will fail.
```
acatangiu: > Once in the `Halted` state, follow-up calls to the same extrinsic will fail.

which "same extrinsic"? the one to halt? maybe be more explicit here.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,120 @@
...
+
+For P→E transfers, in-flight messages are held using the MessageQueue. Outbound messages only get a nonce when they are committed for relay to Ethereum, so while the bridge is halted they are accepted into the queue but not committed. When the bridge resumes, they are committed with fresh nonces, which avoids any stale light client proofs. A small number of P→E messages may have been committed and relayed just before the halt landed, which cannot be held on Bridge Hub; to cover those, we should also consider a bridge operating mode check in the Ethereum `submitV1` and `submitV2` contracts, so they do not process on Ethereum while the bridge is halted.
+
+For E→P transfers, halting holds messages in storage until the bridge resumes. While the bridge is halted, incoming messages are still verified but, rather than being forwarded to Asset Hub, they are kept in the pallet's storage. When the bridge resumes, the held messages are sent on. This applies to the V2 inbound path only, since the older V1 path is being deprecated.
```
acatangiu, line 56: I would expand on both directions about the scope of "in-flight messages" and describe the mechanisms that effectively limit their number.

Afaiu, in the P->E direction, AH stops accepting & producing new bridge messages once `snowbridgeSystemFrontend::set_operating_mode(Halted)` is called (by this new BH emergency pause pallet). This does put an upper bound to "in-flight messages", but this RFC **does not describe what happens with messages generated on other chains destined for Ethereum that then start failing on AH**.

In the E->P direction, things are nicer because once the Gateway is paused on Ethereum, it simply rejects transactions (on Ethereum) before they have on-chain effects. Furthermore, setting `EthereumBeaconClient::set_operating_mode(Halted)` also effectively "freezes" Polkadot's (specifically BH's) view of Ethereum. Maybe make this clearer in the RFC here.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,120 @@
...
+
+In both directions, the held messages can be inspected while the bridge is halted and any malicious ones removed via a governance approved migration before it resumes.
+
+### Resuming the bridge
+
+Resume is the symmetric inverse of the halt. Resuming the bridge and resolution of the halt deposit are separate extrinsics, to allow granular control over the shape of the recovery. The Fellowship will likely bundle the two concerns in a single whitelisted caller proposal (e.g. resume + slash), but in some cases, the specific scenario might require a longer halted bridge state. In that case, the halting account may be refunded, but the bridge should not be resumed yet.
```
acatangiu: Unclear if resuming can be done at any time by the original caller (the one with stake on the line), or  Fellowship has to do it, or it needs OpenGov.

The first is option is easiest, but it potentially opens other attack avenues, where the attacker themselves can halt, and control when and how it resumes.

Fellowship having to do it muddles the water with operational duties that go beyond technical stewardship duties which is its purpose.

So I am leaning on OpenGov having to resume. Or also the first option is good iff halts can be "stacked" so that an honest good samaritan can also add a halt (with separate DOT stake) on top of original one if needed, and it only resumes once all halt callers also call resume (or through OpenGov sudo which is the ultimate fallback), but this just adds more complexity. So in the interest of practicality I propose bridge can only be resumed by OpenGov once it has been halted.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Performance
+
+Performance is not really a concern of this RFC, since the halt is gated by a large deposit and is unlikely to ever receive high traffic. That said, Bridge Hub local operations are O(1) storage writes. The outbound calls to Ethereum and AssetHub are well-defined and there is no performance concern with them.
+
+### Ergonomics
+
+The permissionless halt trigger is an extrinsic with large (to be determined, around 100k) DOT in the signer's account. Offchain relayers should implement watching events for the new pallet, and also stop relaying messages once the pallet `Halted` state is discovered.
+
+The second user of this new function is the Fellowship, who will likely interact with this pallet through whitelisted caller proposals, to resume, slash, refund or extend the halt.
```
acatangiu: the fellowship verifies the technical correctness of already built calls and chooses whether to whitelist/accelerate them, that's it. so it doesn't really directly interact with the ergonomics of the new pallet other than evaluating the way they are used.. anyway, my point is to either remove this section or center it around users that interact with the new pallet either directly or through OpenGov.

acatangiu (commented):
Looks good, just a few comments

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Motivation
+
+Snowbridge has no near-immediate halt path today. Existing governance halt routes require a referendum and Fellowship action (hours-to-days latency). Both are too slow for an active drainage exploit and to stop activity during investigation.
+
+Investigation into the new TX Pause pallet and Safe Mode pallet ([polkadot-fellows/runtimes PR1164](https://github.com/polkadot-fellows/runtimes/pull/1164)) revealed parts that can be reused and referenced, but they do not resolve Snowbridge's need directly. Pallet Safe Mode blocks all calls on the chain, including unrelated parts of the chain, which could have unintended effects for the rest of the chain. Besides this, Snowbridge requires a multi-chain freeze that spans Ethereum contracts, Bridge Hub and Asset Hub. Neither of these two existing pallets support inter-chain messaging. Similarly, pallet TX Pause requires a privileged origin. Snowbridge requires a permissionless pausing mechanism, given a sizeable, slashable deposit.
```
PolkadotDom: Believe safe-mode has a filter list that can exclude pallets from being blocked, fwiw

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,120 @@
...
+
+In both directions, the held messages can be inspected while the bridge is halted and any malicious ones removed via a governance approved migration before it resumes.
+
+### Resuming the bridge
+
+Resume is the symmetric inverse of the halt. Resuming the bridge and resolution of the halt deposit are separate extrinsics, to allow granular control over the shape of the recovery. The Fellowship will likely bundle the two concerns in a single whitelisted caller proposal (e.g. resume + slash), but in some cases, the specific scenario might require a longer halted bridge state. In that case, the halting account may be refunded, but the bridge should not be resumed yet.
+
+It was considered to have the bridge auto resume after a set duration in case the Technical Fellowship should be unavailable for an extended time. The counter argument is that if the Technical Fellowship is unavailable, Polkadot would likely have bigger problems than resuming Snowbridge, and so it was removed from this spec.
```
PolkadotDom: As stated feels more like a footnote, but wouldn't mind a more thorough thought line on the placement of power and why fellowship over alternative structures.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,120 @@
...
+
+The resume extrinsic should do the inverse of all the operations expressed in the previous section, and set the pallet state to `Normal`. While the async calls execute, the bridge might actually be in `Halted` still, but since this is short in duration (1-2 mins) the temporary inconsistency is allowable.
+
+### Releasing or slashing the deposit
+
+The pallet should add two extrinsics to resolve the halting deposit - `slash` and `refund`. Slashing the deposit should send the deposit to Treasury on Asset Hub. Refunding the deposit should release the funds back to the sender. It might be worthwhile to capture a bounded text reason, to capture the reason behind the slash or refund onchain.
```
PolkadotDom: Would think slashing should send to DAP. Could even use the dap-satellite pallet :D

PolkadotDom: Mights should be left out of the main RFC itself I would think. Could open comment on PR if you feel is worth discussing

text/0166-snowbridge-emergency-pause-pallet.md
PolkadotDom: I'm okay w/ Fellowship + OpenGov sudo. I feel we're likely to keep giving fellowship more duties until that operating mode breaks and we spin out separate bureaucracies

text/0166-snowbridge-emergency-pause-pallet.md
acatangiu: Yes, sending to DAP is better/closer to burning, than sending to treasury.

text/0166-snowbridge-emergency-pause-pallet.md
acatangiu: > I feel we're likely to keep giving fellowship more duties until that operating mode breaks and we spin out separate bureaucracies

No. The fellowship has clear purpose of: technical maintenance, technical research/development and technical knowledge preservation. It has no operational duties or responsibilities. We should not break that delimitation even with small bits and pieces like proposed here.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Added in https://github.com/polkadot-fellows/RFCs/pull/166/commits/f4b3bebade96e7892c43cc21faaf07dd6c7718a0.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Good catch, added in https://github.com/polkadot-fellows/RFCs/pull/166/commits/35d460e79b5131b57c08672b99cccd70e7b3062a.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Agreed. Updated in https://github.com/polkadot-fellows/RFCs/pull/166/commits/0d97406ef44c6d618095242e4a626b503b379a97.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Removed unnecessary ergonomics detail in https://github.com/polkadot-fellows/RFCs/pull/166/commits/d84ad4b952875ed1e88d561cbcb8ef2697ef92df.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,120 @@
...
+Investigation into the new TX Pause pallet and Safe Mode pallet ([polkadot-fellows/runtimes PR1164](https://github.com/polkadot-fellows/runtimes/pull/1164)) revealed parts that can be reused and referenced, but they do not resolve Snowbridge's need directly. Pallet Safe Mode blocks all calls on the chain, including unrelated parts of the chain, which could have unintended effects for the rest of the chain. Besides this, Snowbridge requires a multi-chain freeze that spans Ethereum contracts, Bridge Hub and Asset Hub. Neither of these two existing pallets support inter-chain messaging. Similarly, pallet TX Pause requires a privileged origin. Snowbridge requires a permissionless pausing mechanism, given a sizeable, slashable deposit.
+
+## Stakeholders
+
+* **Polkadot OpenGov**, the resolution authority that resumes the bridge and decides between genuine (refund) and malicious (slash) triggers.
+* **Snowbridge maintainers**, who implement and operate the halt path.
+* **Snowbridge users and integrators**, who experience a halt as the bridge being closed at submit time on both Ethereum and AssetHub.
+* **Polkadot Treasury**, the destination of slashed deposits on malicious triggers.
```
acatangiu: Treasury or DAP, personally I am fine with either, but flagging in case someone else has stronger opinion
cc @kianenigma

acatangiu (approved):

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Good catch, added to https://github.com/polkadot-fellows/RFCs/pull/166/commits/936c965811c1e9b8a76406c417d23803b50dfb4f.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Added in https://github.com/polkadot-fellows/RFCs/pull/166/commits/77d86824cad31d6bbb0b3e21f525172889113ae2.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: I think this would be nice to have, so added it more decisively in https://github.com/polkadot-fellows/RFCs/pull/166/commits/f94e5ef36fe64aa86a00a56adc7f8990cc7b0319.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: I changed it to DAP, will keep an eye out for new comments. cc @PolkadotDom

PolkadotDom:
@claravanstaden 'Rendered' is pointing to a specific commit, not latest.

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Stakeholders
+
+* **Polkadot OpenGov**, the resolution authority that resumes the bridge and decides between genuine (refund) and malicious (slash) triggers.
+* **Snowbridge maintainers**, who implement and operate the halt path.
+* **Snowbridge users and integrators**, who experience a halt as the bridge being closed at submit time on both Ethereum and AssetHub.
+* **The Dynamic Allocation Pool (DAP)**, the destination of slashed deposits on malicious triggers, consistent with where other Polkadot slashes now go.
```
PolkadotDom, line 24: Frontend? - I'd think we'll want to integrate this into some UI somewhere so it's simple for anyone to call

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,126 @@
...
+
+The halt blocks new transfers from entering the bridge:
+
+1. `snowbridgeSystemFrontend::set_operating_mode(Halted)`: Sets the frontend's export mode on Asset Hub, blocking new P→E transfers there.
+2. Outbound governance command issued from Bridge Hub via `snowbridge-pallet-system-v2`: Sets the Gateway's operating mode, blocking new E→P transfers on Ethereum.
+3. `EthereumBeaconClient::set_operating_mode(Halted)` - a local Bridge Hub call that stops new Ethereum header ingestion, as defense-in-depth.
+
+Messages that were already in flight when the halt landed are not rejected. Rejecting them would leave the bridge inconsistent, for example an asset burned on one side with nothing minted on the other. Instead, in-flight transfers are held and sent on resume, as described below.
+
+These calls are all best-effort, and failure does not prevent the other calls from being executed. The pallet attempts each, logs successes and failures, and re-attempts pending calls in later blocks via `on_initialize`.
```
PolkadotDom, line 50: Not critique, more discussion. Any concerns here regarding filling up state if one of the halts fails?

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,126 @@
...
+
+The resume extrinsic should do the inverse of all the operations expressed in the previous section, and set the pallet state to `Normal`. While the async calls execute, the bridge might actually be in `Halted` still, but since this is short in duration (1-2 mins) the temporary inconsistency is allowable.
+
+### Releasing or slashing the deposit
+
+The pallet should add two extrinsics to resolve the halting deposit, `slash` and `refund`, both voted on by OpenGov. Slashing the deposit should send it to the Dynamic Allocation Pool (DAP), where other Polkadot slashes now go, routed from Bridge Hub via the `dap-satellite` pallet. Refunding the deposit should release the funds back to the caller. The reason behind the slash or refund should be captured on-chain as text.
+
+### Threat model coverage
+
+- **New E→P entry**, Gateway halt (call 2) stops new transfers starting on Ethereum.
```
PolkadotDom: nit. wouldn't mind if ordered call 1 -> call 2, etc. If that's the way it's supposed to be understood

text/0166-snowbridge-emergency-pause-pallet.md
```diff
@@ -0,0 +1,126 @@
...
+- **E→P in-flight + inbound-queue exploit**, the inbound hold. Covers exploits that bypass the Gateway entirely (malformed proofs, payload-decode bugs, MMR weaknesses): messages are verified and held, and malicious ones can be dropped before reaching Asset Hub.
+- **New P→E entry**, AH frontend halt (call 1) stops new transfers at Asset Hub.
+- **P→E in-flight**, the outbound hold keeps messages queued and uncommitted until resume.
+- **Beacon-client exploit**, beacon client halt (call 3).
+
+## Drawbacks
+
+- **Griefing**: This proposal adds permissionless halting, guarded by a slashable deposit. Someone who is willing to lose funds to censor the bridge, could repeatedly call the permissionless halt. In practice, this seems unlikely. Should this happen, the deposit amount can be upped as a further deterrent.
```
PolkadotDom: 'unlikely' is unsubstantiated. I feel it would be more accurate to say it's a tradeoff we're willing to make right now or some other variant

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: I understand frontends are not part of the core technology, so it can be considered but not in the RFC.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Good question! 

- E→P: Messages accumulate in pallet storage while halted, but it is bounded because the beacon-client halt stops new headers being imported, so the inbound verification pipeline dries up. Even without that, inbound is gated by finalized Ethereum headers, so practically speaking it is bounded by that too.
- P→E: In the outbound MessageQueue, accepted but uncommitted message are held. This side stays bounded if the frontend halt executes on Asset Hub. If it doesn't, it could cause problems, but yeah if it doesn't succeed there might be bigger problems anyway.

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Cleaned up in https://github.com/polkadot-fellows/RFCs/pull/166/commits/bd46fe14258346d5085c7e98b11b30e975f9a962

text/0166-snowbridge-emergency-pause-pallet.md
claravanstaden: Rephrased in https://github.com/polkadot-fellows/RFCs/pull/166/commits/bd46fe14258346d5085c7e98b11b30e975f9a962

PolkadotDom (approved):

claravanstaden:
/rfc process 0x7df76dd4676a2909a789de224f55e9eec0837ee0fe25840ae290a332a7cdc685

PR #167, opened 2026-05-28
RFC-0167: Snowbridge Circuit Breaker

claravanstaden:
Proposes per-asset rate-limiting for Snowbridge in both directions: on the Ethereum Gateway for P→E and on Asset Hub for E→P. Each asset has a governance-set cap that refills over 24 hours. Transfers that fit go through, transfers that don't are held until capacity returns, and transfers larger than the cap are delayed in proportion to their size.

[Rendered](https://github.com/claravanstaden/RFCs/blob/4433518331465cbafa9f28097149ccc122a24d87/text/0167-snowbridge-circuit-breakers.md)


text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,118 @@
...
+Snowbridge's honey pot is primarily on Ethereum - all locked funds bridged from Ethereum to Polkadot are located in the Snowbridge gateway contract. For this reason, it makes sense to protect these assets from irregular activity, in the gateway contract. Additionally, Polkadot Native Assets (PNAs) like DOT are minted on Ethereum, backed by assets on Asset Hub, which also need protection against irregular minting.
+
+If one considers the possible exploit shapes, they would all be protected by a circuit breaker, given the circuit breaker catches the outflow pattern:
+
+- Forged BEEFY commitment
+- MMR proof bug
+- Gateway message decode bug
+- PNA minting bug
+
+The implementation should track per-asset, net outflow over a rolling 24 hour window, both ERC-20s and Ether, and PNAs. For each asset and each class, net movement (outflow - inflow) is tracked over the window. Net flow is tracked so two-way arbitrage and market-maker activity doesn't trigger the cap and unnecessarily delay transactions. This is borrowed from Hydration's `pallet-circuit-breaker` net-volume pattern.
```
yrong: Is E→P inflow currently tracked or protected by any circuit breaker on the Polkadot side?

If the bridge were compromised—for example, due to a flaw in proof verification—a malicious actor could potentially mint unbacked wrapped assets on the other end. In that scenario, the risk comes from the bridge accepting invalid proofs rather than from any legitimate backing on Ethereum.

This is why I think bidirectional circuit breakers are necessary. Ideally:

- P→E limits should be enforced by the Ethereum Gateway circuit breaker, restricting outflows from Polkadot to Ethereum.
- E→P limits should be enforced by the Asset Hub circuit breaker, restricting the minting of bridged assets on Polkadot.

That way, even in the event of a bridge compromise, the blast radius is bounded in both directions.

vgeddes: The two different sides of the bridge have very different implementations and so I'm happy if we focus on P->E path in this initial proposal.

vgeddes (approved):

text/0167-snowbridge-circuit-breakers.md
claravanstaden: For E->P, should we not implement the same pattern on BridgeHub? I'm wary of doing a 50% implementation, would rather cover both directions?

vgeddes: Sounds good👌

Whether we do both sides now, or do them iteratively, is up to you.

yrong: > For E->P, should we not implement the same pattern on BridgeHub?

BridgeHub only verifies inbound messages — there's no asset management there. The minting/backing of bridged assets happens on Asset Hub via `pallet-assets`, so that's the natural place to enforce a limit (it's also where the value to bound actually lives). 

I'd expect this to be implemented as a part/extension of pallet-assets on Asset Hub rather than something bridge-specific. That way it protects the whole Polkadot ecosystem, not just Snowbridge — any pallet that mints or backs assets benefits from the same circuit breaker.

claravanstaden: @yrong I added a section about this in https://github.com/claravanstaden/RFCs/blob/1072fdfb4deeede41661c3da691f7c18d3ec4559/text/0167-snowbridge-circuit-breakers.md#asset-hub-circuit-breaker. Please check if you agree. :)

text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,139 @@
...
+
+If one considers the possible exploit shapes, they would all be protected by a circuit breaker, given the circuit breaker catches the outflow pattern:
+
+- Forged BEEFY commitment
+- MMR proof bug
+- Gateway message decode bug
+- PNA minting bug
+
+The implementation should track per-asset, net outflow over a rolling 24 hour window, both ERC-20s and Ether, and PNAs (Polkadot native assets, like DOT). This part of the circuit breaker is specifically for P->E transfers For each asset and each class, net movement (outflow - inflow) is tracked over the window. Net flow is tracked so two-way arbitrage and market-maker activity doesn't trigger the cap and unnecessarily delay transactions. This is borrowed from Hydration's `pallet-circuit-breaker` net-volume pattern. 
+
```
yrong, line 43: I assume Netting E→P mints against P→E burns is safe only in a single trust domain like Hydration. 

Snowbridge has two independent trust roots (Beacon for E→P, BEEFY for P→E), Netting them assumes both clients are honest — but bounding a single-client compromise is the breaker's whole purpose.

Concretely, if Beacon is compromised: attacker mints X unbacked on AH, then burns it P→E to drain the real Gateway reserve (validated by BEEFY, so it looks legit). Under `net = minted − burned`, net is 0 and the breaker never trips.

So I think each direction's cap must be gross in the quantity its own client protects.

claravanstaden: For this specific case, I believe it is catered for:

- Forge E→P mint X (Beacon) → X unbacked on Asset Hub.
- Burn X P→E → drains X from the Gateway.
- AH breaker: net = X − X = 0, doesn't trip.
- Gateway breaker: the drain is a P→E outflow of X, and the forged mint never deposited into the Gateway, so there's no E→P inflow to net against. net = X − 0 = X → trips.

PolkadotDom:
Rendered takes you to the changes screen

claravanstaden:
@PolkadotDom fixed, thanks!

text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,141 @@
...
+| --------------- |-------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **Start Date**  | 2026-05-28                                                                                                                                            |
+| **Description** | Per-asset velocity caps on the Ethereum Gateway (P→E) and Asset Hub (E→P) that automatically throttle irregular Snowbridge flows in both directions. |
+| **Authors**     | Snowbridge team                                                                                                                                       |
+
+## Summary
+
+Snowbridge does not currently have any proactive on-chain security measures in place, in case of irregular activity. This proposal suggests adding on-chain rate limiting, with the main goal being reducing artificial latency in case of an exploit, which gives the team time to halt the bridge. This RFC goes hand in hand with the reactive security measure that is introduced in [Snowbridge Emergency Pause Pallet RFC](./0166-snowbridge-emergency-pause-pallet.md).
```
PolkadotDom, line 11: did you mean 'introducing artificial latency'?

text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,141 @@
...
+- Forged BEEFY commitment
+- MMR proof bug
+- Gateway message decode bug
+- PNA minting bug
+
+The implementation should track per-asset, net outflow over a rolling 24 hour window, both ERC-20s and Ether, and PNAs (Polkadot native assets, like DOT). This part of the circuit breaker is specifically for P->E transfers For each asset and each class, net movement (outflow - inflow) is tracked over the window. Net flow is tracked so two-way arbitrage and market-maker activity doesn't trigger the cap and unnecessarily delay transactions. This is borrowed from Hydration's `pallet-circuit-breaker` net-volume pattern. 
+
+A 24 hour window is suggested, as the delay needs to be long enough for bridge operators to notice. The window matches bridges like Wormhole and LayerZero's behaviour. Assets should be tracked by denomination, not USD, so that it doesn't create reliance on oracles. Assets without a cap ignore the circuit breaker pattern, so that the tracking is opt-in by way of governance vote.
+
+The suggested cap formula is `cap = max(5x trailing-7-day-median hourly net flow, configured floor per asset)`.
```
PolkadotDom, line 46: Any consideration of using variance/stddev in this formulation? ex median + (variance of median) * 5?

text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,141 @@
...
+
+- The 5x multiplier is high enough to not trip on legitimate spikes.
+- The floor prevents low-volume-but-high-value assets from having too low a cap relative to their total locked value.
+- Both factors are governance-settable per asset; no automatic defaults at asset registration (new assets start uncapped).
+
+If the cap is tripped, asset movement is locked for a certain set time (proposed 24 hours). Other asset transfers continue as normal. Once the locked time elapses, the asset transfer continues as normal.
```
PolkadotDom, line 52: Any considerations for griefing? Also, just for clarity, what exactly trips it? There's a 'net volume last 24h's' tracker, if your transaction puts this metric over 5x the median hourly net flow it will trip?

text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,141 @@
...
+- The 5x multiplier is high enough to not trip on legitimate spikes.
+- The floor prevents low-volume-but-high-value assets from having too low a cap relative to their total locked value.
+- Both factors are governance-settable per asset; no automatic defaults at asset registration (new assets start uncapped).
+
+If the cap is tripped, asset movement is locked for a certain set time (proposed 24 hours). Other asset transfers continue as normal. Once the locked time elapses, the asset transfer continues as normal. 
+
+Specific implementation details:
+
+- E→P: Tracks inflow.
+- P→E: Tracks outflow, and checks cap. Asset transfers that would breach the cap are deferred, and the nonce is not processed. Relayers should watch and respect this lock, and resubmit the transaction when the lock lifts.
```
PolkadotDom, line 57: If relayers were to relay anyway, gateway would reject the transfers correct? A bit confused by the 'should watch and respect' here.

text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,141 @@
...
+
+When an inbound (E→P) transfer would breach the cap it is delayed. Instead of delivering it, the wrapper records the pending deposit (its asset and beneficiary) and returns success. The pallet's `on_initialize` hook later replays that deposit through the same transactor once the cap is no longer breached, completing as many held transfers as fit each block and re-checking the cap as it goes, so a backlog drains gradually rather than re-tripping all at once. Nothing is delivered until release, and release just re-runs the existing deposit path, so no delivery logic is duplicated and a transfer later judged malicious is simply dropped. This is the Asset Hub counterpart to the Gateway side, where the relayer holds and resubmits off-chain; here the held transfer is parked on-chain and completed by the runtime.
```
PolkadotDom, line 80: Assuming yes, but all the holds are supposed to work with the #166's halting, correct? Like if we halt, the transfers on hold won't go through?

PolkadotDom:
Just some questions. Great read! Thanks Clara

text/0167-snowbridge-circuit-breakers.md
yrong: > Forge E→P mint X (Beacon) → X unbacked on Asset Hub.

Should't we block this forged mint at the start?  My concern is the other path: forge mint ETH → dump into Hydration (ETH/USDC etc.) and drain the pool → optionally exit via another asset. Damage hits ecosystem LPs before any Gateway trip. Cross-asset DEX hops also aren't visible to per-asset net meters.

text/0167-snowbridge-circuit-breakers.md
yrong: Yes, defining relayer behavior in the spec doesn't really make sense. At most, it can describe common practices for well-behaved (or "legitimate") relayers. Ultimately, though, relayers are free to behave however they want—they're profit-driven.

text/0167-snowbridge-circuit-breakers.md
yrong: Another option, as mentioned before: implement this as a **common-good** module on top of `pallet-assets` / `FungibleAdapter`, so it can benefit parachains using Assets/Balances more generally. 

Inbound assets would arrive to recipient address locked (non-transferable / non-XCM-able); the lock lifts after the window, or becomes permanent if the transfer is later judged malicious.

text/0167-snowbridge-circuit-breakers.md
acatangiu:
```suggestion
Snowbridge does not currently have any proactive on-chain security measures in place, in case of irregular activity. This proposal suggests adding on-chain rate limiting, with the main goal being introducing artificial latency in case of an exploit, which gives the team time to halt the bridge. This RFC goes hand in hand with the reactive security measure that is introduced in [Snowbridge Emergency Pause Pallet RFC](./0166-snowbridge-emergency-pause-pallet.md).
```

text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,141 @@
...
+
+The downside with circuit breakers is that legitimate transactions might be delayed from time to time. The circuit breaker settings will be calibrated to minimize this as much as possible. Even if larger transactions might be delayed, this is an acceptable tradeoff for protection against exploits.
+
+## Stakeholders
+
+- **Snowbridge maintainers**, who implement this proposal and suggest initial caps.
+- **Polkadot Fellowship and OpenGov**, who vote on per-asset caps via governance.
```
acatangiu:
```suggestion
- **OpenGov**, who vote on per-asset caps via governance.
```

text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,141 @@
...
+Snowbridge's honey pot is primarily on Ethereum - all locked funds bridged from Ethereum to Polkadot are located in the Snowbridge gateway contract. For this reason, it makes sense to protect these assets from irregular activity, in the gateway contract. Additionally, Polkadot Native Assets (PNAs) like DOT are minted on Ethereum, backed by assets on Asset Hub, which also need protection against irregular minting.
+
+If one considers the possible exploit shapes, they would all be protected by a circuit breaker, given the circuit breaker catches the outflow pattern:
+
+- Forged BEEFY commitment
+- MMR proof bug
+- Gateway message decode bug
+- PNA minting bug
+
+The implementation should track per-asset, net outflow over a rolling 24 hour window, both ERC-20s and Ether, and PNAs (Polkadot native assets, like DOT). This part of the circuit breaker is specifically for P->E transfers For each asset and each class, net movement (outflow - inflow) is tracked over the window. Net flow is tracked so two-way arbitrage and market-maker activity doesn't trigger the cap and unnecessarily delay transactions. This is borrowed from Hydration's `pallet-circuit-breaker` net-volume pattern.
```
acatangiu:
```suggestion
The implementation should track per-asset, net outflow over a rolling 24 hour window, both ERC-20s and Ether, and PNAs (Polkadot native assets, like DOT). This part of the circuit breaker is specifically for P->E transfers. For each asset and each class, net movement (outflow - inflow) is tracked over the window. Net flow is tracked so two-way arbitrage and market-maker activity doesn't trigger the cap and unnecessarily delay transactions. This is borrowed from Hydration's `pallet-circuit-breaker` net-volume pattern. 
```

text/0167-snowbridge-circuit-breakers.md
acatangiu: @yrong can you describe the scenario in more detail please?

afaiu, in the current proposal there are circuit breakers in both directions, each operating on their respective minting side, so that the circuit breaker actives _before_ actual minting.

The one concern of "netting" bi-directional transfers is that when **both** directions are compromised, an attacker can fraudulently mint in both directions _without_ tripping the breaker, but seeing how this is an asymmetric bridge with different consensus light clients on each side, the risk of both sides being compromised at the same time is very small.

"netting" does however allow an attacker that found a vulnerability on one side, to piggy-back on legitimate liquidity flowing in the other direction to increase their total drain in the compromised direction before tripping the breaker.

Either way, there is no perfect solution and a compromise between the extent of security in depth vs UX and usability must be made.

text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,141 @@
...
+
+- **Snowbridge maintainers**, who implement this proposal and suggest initial caps.
+- **Polkadot Fellowship and OpenGov**, who vote on per-asset caps via governance.
+- **Snowbridge users and integrators**, who experience a tripped cap as a temporary lockdown of one asset+direction. Main stakeholder here is Hydration.
+- **Asset issuers** whose tokens are listed on Snowbridge and would have caps set against them.
+- **Relayers and the existing relayer-operated monitoring stack**, which becomes the alerting backbone for cap events.
+
+## Explanation
+
+### Gateway Circuit Breaker
```
acatangiu: for clarity of where exactly this one resides - on ethereum side in the gateway contract

```suggestion
### Gateway (Ethereum contract) Circuit Breaker
```

text/0167-snowbridge-circuit-breakers.md
acatangiu: I think this section is too opinionated about the implementation for the purpose of this RFC.
I expect we cannot fully spec out the mechanism here and implementation subtleties will come about.

I propose to keep this RFC high-level, describing the intended behaviour and not attempt to describe exactly how that behaviour will be achieved.

acatangiu (commented):
Overall looks good, some small comments to address before I fully approve.

text/0167-snowbridge-circuit-breakers.md
claravanstaden: Agreed, removed impl detail in https://github.com/polkadot-fellows/RFCs/pull/167/commits/5707ba5fe311452003868ad4e53851ca382eb90a.

text/0167-snowbridge-circuit-breakers.md
claravanstaden: I think the two options here are:

- `Gross limits`: each side only looks at their outflow, and ignores the other direction as inflow to net against. This is simpler, but the cap likely needs to be higher to prevent tripping on legimate market in and out flows.
- `Bounded netting`: more complicated, but probably a better chance at actually capturing asymmetric movements.

I checked the bridge flow data, and most of the traffic is effectively one-way, so I think switching to gross might be better, because it's simpler, and netting doesn't seem to add much for our kind of usage.

Will post the gross limits commit here once done.

text/0167-snowbridge-circuit-breakers.md
claravanstaden: Yep, corrected!

text/0167-snowbridge-circuit-breakers.md
claravanstaden: Good question. I checked the numbers against the current bridge traffic. For every asset, 81 to 100% of hours in a 7-day window have zero outflow, so the hourly median is 0. Mean + stddev would not fix it, since traffic comes in bursts and the result swings from 0 to ~39× a busy day depending on whether a whale bridged that week. An adaptive cap can also be inflated by an attacker ahead of time, and stddev would be an easy stat to manipulate. I'm leaning towards dropping the on-chain formula and making the cap a fixed per-asset governance value, proposed from long-term data and reviewed periodically. What do you think of this?

text/0167-snowbridge-circuit-breakers.md
claravanstaden: I've changed the spec to gross limits since your last review, so that changes this picture a bit.

To prevent griefing, we can introduce a refilling limit where capacity recovers linearly over a time period. @yrong @PolkadotDom what do you think of this?

text/0167-snowbridge-circuit-breakers.md
claravanstaden: Agreed, I removed the relayer impl details.

text/0167-snowbridge-circuit-breakers.md
claravanstaden: @PolkadotDom yes, halted transfers won't go through while the bridge is halted via #166, even if the 24h lock lifts in the meantime. I've added this to the RFC.

text/0167-snowbridge-circuit-breakers.md
claravanstaden: Added in https://github.com/polkadot-fellows/RFCs/pull/167/commits/c1ac357acbbc505626e54d67aa4298cb0c960d8e.

claravanstaden:
This RFC is ready to review again, with an added question about a refilling limit: [link]

I think a refilling limit might be a better pattern, since it limits the impact of a tripped cap on other users (including griefing).

For example, with cap = 5 per 24h:

Current RFC (24h lock):
- Transfer of 3: passes, used = 3.
- Another 3: 3 + 3 = 6, over 5. The breaker trips and the whole asset is locked for 24h.
- A transfer of 1 is also blocked until the lock lifts, even though it would have fit.

Refilling limit:
- Transfer of 3: passes, used = 3.
- A transfer of 1: passes, used = 4. Small transfers keep flowing as long as there's room.
- Another 3: doesn't fit, so only this transfer is held. Nothing gets locked.
- Used capacity drains back at 5 per 24h (about 1 every 5 hours). Once used has dropped enough for the held 3 to fit, it goes through.

One problem to solve is ordering. If new transfers can take capacity as it frees up, a stream of small transfers could keep a large held transfer waiting indefinitely, which is also a griefing vector.

The fix I'd propose: a held transfer reserves its capacity at the moment it's held. In the example above, holding the 3 sets used = 4 + 3 = 7. New transfers can't pass while used is over the cap, so the freed capacity goes to the held transfer first. After about 10 hours used is back to 5 and the held 3 goes through, since its amount is already counted. This gives FIFO behaviour with a single counter per asset, plus a flag per held message so the retry skips the cap check. Transfers larger than the cap itself would be rejected, so a single held transfer can't block an asset for more than 24h.

Let me know what you think.

acatangiu:
> I think a refilling limit might be a better pattern, since it limits the impact of a tripped cap on other users (including griefing).

I agree. Not just better, but actually required if the cap trips on one-way gross amount (no bidirectional netting). Griefing is very cheap as the griefer/attacker does not even need to own or "lock" the cap amount, they can just rotate the same liquidity back and forth to bring the gross up the cap for each direction.

A refilling limit puts the market back in control, since honest parties can outbid the attacker on relay fees/tips even during an active griefing attack.

claravanstaden:
@yrong @PolkadotDom @acatangiu I have updated the design to use refilling, please review.

text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,139 @@
...
+
+The limit works like a leaky bucket. Each asset has a bucket that holds one cap. A transfer adds its amount to the bucket and goes through if it fits. The bucket leaks at a steady rate and empties in 24 hours. The leak is continuous and not tied to any transfer. A transfer that has gone through is done. The bucket level is only a record of how much has gone through recently. If a transfer does not fit, only that transfer is held. The asset is not locked as a whole. Other transfers of the same asset keep going through as long as they fit. A held transfer goes through once enough has leaked out for it to fit.
+
+For example, with a cap of 5 per 24 hours: a transfer of 3 goes through (bucket at 3), a transfer of 1 goes through (bucket at 4), and a further transfer of 3 does not fit and is held. The bucket leaks 5 per 24 hours, about 1 every 5 hours. If nothing else goes through, after about 10 hours the bucket is down to 2 and the held 3 fits. The same leak sets the worst case: starting from an empty bucket, an attacker can send 5 at once and another 5 over the next 24 hours as the bucket leaks, so 10 in the first day. After that only what leaks out can be added, 5 per day.
```
acatangiu, line 37: As future work we should add a way for the `origin` or `authorize_claimer` to cancel a held transfer, and claim/unlock their funds. Thus, they may choose to cancel and retry with smaller chunks, rather than wait for their large transfer to go through.

But this can be future improvement, and not mandatory part of this RFC

text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,139 @@
...
... leave the user's funds out of sync. Instead it is delayed at the destination for `amount / cap × 24 hours` and then goes through without counting against the bucket. Large transfers of the same asset are queued: each one's release time is fixed when it enters the queue as the later of now and the previous large transfer's release time, plus its own delay. So with a cap of 5, transfers of 15, 10 and 6 arriving on the same day are released on day 3, day 5 and day 6.2, which is 5 per day. Without the queue, many transfers of just over the cap could all be released after about 24 hours. A single release is still the whole transfer, so the worst case for an asset is one cap at once and one cap per 24 hours through the bucket, plus through the queue any amount that has been visible on-chain for at least one day per cap of its size. This is intended: the breaker delays large transfers, and the emergency pause stops fraudulent ones. A large transfer does not affect ordinary transfers of the same asset, but it does delay large transfers queued behind it. A griefer can queue one very large transfer to make honest large transfers wait, but the griefer's own funds are stuck for the same time.
```
acatangiu, line 41: 👍

text/0167-snowbridge-circuit-breakers.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Gateway breaker (P→E)
+
+Snowbridge's honey pot is primarily on Ethereum - all locked funds bridged from Ethereum to Polkadot are located in the Snowbridge gateway contract. For this reason, it makes sense to protect these assets from irregular activity, in the gateway contract. Additionally, Polkadot Native Assets (PNAs) like DOT are minted on Ethereum, backed by assets on Asset Hub, which also need protection against irregular minting.
+
+If one considers the possible exploit types, they would all be protected by a circuit breaker, given the circuit breaker catches the outflow pattern:
+
+- Forged BEEFY commitment
```
acatangiu: Not really. With a forged BEEFY commitment, an attacker can upgrade the Gateway contract and bypass any security measures whatsoever.

acatangiu (approved):
Looks good now!

text/0167-snowbridge-circuit-breakers.md
claravanstaden: Addressed in https://github.com/polkadot-fellows/RFCs/pull/167/changes/7306f4df82ac19e846e98a052bbf756dcc8613ba.

text/0167-snowbridge-circuit-breakers.md
claravanstaden: True, addressed in https://github.com/polkadot-fellows/RFCs/pull/167/changes/7306f4df82ac19e846e98a052bbf756dcc8613ba.

claravanstaden:
I jumped the gun a bit, waiting for a review from @yrong and @PolkadotDom before I propose it on-chain. :)

yrong:
The bucket refilling looks good. One minor point on **held vs queued:** a transfer just under the cap gets a worse deal than one at or over it. e.g., Cap 5, bucket at 4:

- A 4.9 is held with no release time. It goes through only once the bucket drops below 0.1, and every smaller transfer that arrives first takes the leaked capacity.
- A 5.0 gets a fixed 24h release, regardless of the bucket.

So people will round up to the cap, and a held transfer can starve with no way out but governance (cancellation being future work).

Anyway, I'd treat this as an implementation detail rather than a blocker.

yrong (approved):
+1

acatangiu:
> So people will round up to the cap

and that is perfectly fine, the main purpose of being able to "audit" super-large liquidity movements is still satisfied

PolkadotDom (approved):
Really like this new direction 🙌

acatangiu:
Ref is up: https://collectives.subsquare.io/fellowship/referenda/621

PR #169, opened 2026-06-02
RFC-0169: Society V2.1

laurogripa:
This RFC proposes Society V2.1 as a bundle of fixes for issues that appeared after the V2 migration. It documents the recent no-PoI incident, the treasury exposure caused by delayed claims (what I call `the bid/claim gap`), among other things. It proposes a few possible solutions to deal with these exploits. For instance: a quorum to handle low-participation candidate approvals, better pot accounting, stronger Proof-of-Ink requirements, and adding missing pallet events. 

Beyond the immediate fixes, the RFC is an opportunity to discuss the future of the Society pallet: whether it still fits into the Proof of Personhood vision, what role it should play, how to address the current lack of utility, and the remaining question of what to do with `kick_member`.


bkchr (changes requested):
Ty for this, but this is not really a RFC.

This is more like a starting point for a discussion. IMO the best would be to post this to the forum and then go from there to a RFC.

sigurpol:
> Ty for this, but this is not really a RFC.
> 
> This is more like a starting point for a discussion. IMO the best would be to post this to the forum and then go from there to a RFC.

Agree with Basti here: this is an excellent starting point for a forum discussion.

laurogripa:
@bkchr @sigurpol 

Thanks for the guidance!

Here is the forum post: https://forum.polkadot.network/t/let-s-talk-about-kusama-society-s-future/17771

Closing this for now.

PR #172, opened 2026-06-25
RFC: Ethereum JSON-RPC compatibility standard for pallet-revive

soloking1412:
## RFC-0172: Ethereum JSON-RPC compatibility standard for `pallet-revive`

Rendered: [`text/0172-eth-rpc-compatibility-pallet-revive.md`](https://github.com/polkadot-fellows/RFCs/blob/rfc-eth-rpc-compatibility/text/0172-eth-rpc-compatibility-pallet-revive.md)

### What this proposes
`pallet-revive`'s Ethereum JSON-RPC server (`pallet-revive-eth-rpc`) exists so that unmodified Ethereum tooling works against a Polkadot system chain. Today there is no written specification of *which* behaviour is guaranteed, and several behaviours diverge silently from the de-facto reference (go-ethereum). This RFC proposes:

1. Adopting the Ethereum [`execution-apis`](https://github.com/ethereum/execution-apis) spec + go-ethereum as the **normative conformance target**.
2. **Standardising the Substrate↔Ethereum semantics** where they genuinely diverge — most importantly the mapping of Ethereum block tags onto GRANDPA finality (`safe`/`finalized` → last finalised block, `pending` → `latest`), plus EIP-1474 error codes and filter edge cases.
3. **Standardising the gas ↔ resource metering fold** (§2.4): revive meters two-dimensional `Weight` (`ref_time` + `proof_size`) plus a refundable storage deposit, so the RFC pins how those collapse into a single gas number — the binding-dimension fold plus the deposit — so `eth_estimateGas` is a true upper bound and agrees across revive chains.
4. Requiring a **conformance test suite** in CI so the guarantee is enforced rather than aspirational.

### Why now (concrete motivation)
Each of these was a real, separately-discovered divergence fixed in or reported against `polkadot-sdk`:
- `eth_feeHistory` returned the wrong reward bucket — paritytech/polkadot-sdk#12470
- `eth_getLogs` rejected the `finalized`/`safe`/`pending` block tags — paritytech/polkadot-sdk#12474
- `eth_getLogs` silently returned zero logs for empty `address`/`topics` filter lists (treated as an always-false `IN ()`) — paritytech/polkadot-sdk#12479
- error→JSON-RPC code mapping didn't follow EIP-1474 — paritytech/polkadot-sdk#11887
- `eth_estimateGas` had no specified fold from revive's two-dimensional `Weight` + storage deposit onto single-dimensional gas, so proof-size-heavy calls could be mispriced — paritytech/polkadot-sdk#11525, paritytech/polkadot-sdk#10751

The pattern (independent, unspecified, untested-against-reference behaviour) is the systemic issue this RFC addresses.

### Scope fit
Falls under the Fellowship's in-scope "standard RPCs" and "runtime public interfaces" of system-chain pallets; the repo notes node-side interface standards are the Fellowship's strongly-binding domain.

### The key decision to discuss
The one genuine design question is the block-tag → finality mapping (§2.1): is mapping `safe` to `finalized` (rather than exposing a weaker checkpoint) the right call, and is `pending` → `latest` acceptable given there's no mempool-backed speculative block? Feedback from the eth-rpc maintainers especially welcome.

### DIFF
--- a/text/0000-eth-rpc-compatibility-pallet-revive.md
+++ b/text/0000-eth-rpc-compatibility-pallet-revive.md
@@ -1,0 +1,136 @@
+# RFC-0000: Ethereum JSON-RPC compatibility standard for `pallet-revive`
+
+|                 |                                                                       |
+| --------------- | --------------------------------------------------------------------- |
+| **Start Date**  | 2026-06-25                                                            |
+| **Description** | Define a normative Ethereum JSON-RPC conformance target, Substrate↔Ethereum semantics, and a conformance test suite for `pallet-revive`'s `eth-rpc` server. |
+| **Authors**     | Maheswaran Velmurugan (@solokingm)                                    |
+
+## Summary
+
+`pallet-revive` exposes an Ethereum-compatible JSON-RPC interface (`pallet-revive-eth-rpc`) so that existing Ethereum tooling — wallets, libraries, indexers, block explorers — can talk to a Polkadot chain unchanged. Today this interface is implemented method-by-method with no written specification of *which* behaviour is guaranteed, and several behaviours diverge silently from the de-facto Ethereum reference (go-ethereum). This RFC proposes (1) adopting the Ethereum `execution-apis` specification as the normative conformance target, (2) standardising the handful of places where Substrate and Ethereum genuinely diverge — most importantly the mapping of Ethereum block tags onto Substrate's GRANDPA finality — and (3) requiring a conformance test suite in CI so the guarantee is enforced rather than aspirational.
+
+## Motivation
+
+The value proposition of `pallet-revive` is that *unmodified* Ethereum tooling works against a Polkadot chain. That promise only holds if the JSON-RPC surface behaves the way Ethereum clients expect. Three problems make this fragile today:
+
+1. **No written conformance target.** Each `eth_*` method is implemented independently; there is no document stating that the server aims to match the `execution-apis` specification and the go-ethereum reference, nor what "match" means for edge cases. As a result, divergences are discovered only when a downstream tool breaks.
+
+2. **Undocumented Substrate↔Ethereum semantics.** Some Ethereum concepts have no one-to-one Substrate equivalent. The clearest example is block tags: Ethereum's `safe`/`finalized`/`pending` are defined in terms of the beacon-chain consensus and the mempool, neither of which maps directly onto GRANDPA finality and Substrate's block lifecycle. The current code makes implicit choices (and in places rejects valid inputs) with no specification a tool author can rely on.
+
+3. **Edge-case divergences are real and recurring.** Concrete examples found and fixed while preparing this RFC:
+   - `eth_feeHistory` returned the wrong reward bucket because the cache lookup discarded the half-percentile resolution the cache was built at ([paritytech/polkadot-sdk#12470](https://github.com/paritytech/polkadot-sdk/pull/12470)).
+   - `eth_getLogs` rejected the standard block tags `finalized`/`safe`/`pending` in filter ranges with an "Unsupported tag" error, although the same tags are accepted elsewhere in the server ([#12474](https://github.com/paritytech/polkadot-sdk/pull/12474)).
+   - `eth_getLogs` produced invalid `IN ()` SQL — and therefore an error — for the valid filters `{"address": []}` and `{"topics": [[]]}`, which Ethereum clients treat as "match anything" ([#12479](https://github.com/paritytech/polkadot-sdk/pull/12479)).
+   - The mapping of internal errors to JSON-RPC error codes did not follow EIP-1474 ([#11887](https://github.com/paritytech/polkadot-sdk/pull/11887)).
+
+   Each was a small fix, but the *pattern* — independent, unspecified, untested-against-reference behaviour — is the underlying problem this RFC addresses.
+
+The requirement for the solution is: a single, citable definition of the compatibility the `eth-rpc` server provides, a small set of explicit decisions for the cases where Ethereum semantics do not map cleanly onto Substrate, and an automated way to detect regressions against that definition.
+
+This is squarely within the Fellowship's remit: the RFC scope lists "standard RPCs" and the "runtime public interfaces" of pallets used by system chains as in-scope concerns, and the repository notes that for node-side standards (such as RPC interfaces) the Fellowship's view is *strongly* binding, because all implementations "should conform to some foundational standards in order to communicate". The `eth-rpc` server is exactly such a foundational, cross-implementation interface standard.
+
+## Stakeholders
+
+- **Smart-contract developers and tooling authors** targeting `pallet-revive` (Foundry/Hardhat users, `ethers`/`viem`/`web3.js`, The Graph and other indexers, wallets such as MetaMask). They are the direct beneficiaries and the primary consumers of the guarantee.
+- **`pallet-revive` / `eth-rpc` maintainers** at Parity, who would own the conformance suite and the documented semantics.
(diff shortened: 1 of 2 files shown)


Nathy-bajo:
Hey @soloking1412, thanks for the RFC, solid proposal. Funny timing, I'd landed on the same idea and was about to send mine, so glad to fold it into yours instead of opening a competing PR.

One thing I'd add as a complementary section: the gas to resource metering model your conformance target relies on but doesn't pin down. `execution-apis`/go-ethereum assume single dimensional gas, but revive meters two dimensional `Weight` (`ref_time`+`proof_size`) plus a storage deposit. Without standardising how those fold into one gas number, `eth_estimateGas` diverges between revive chains and proof size heavy calls get mispriced, the same silent divergence your motivation flags (see https://github.com/paritytech/polkadot-sdk/issues/11525 and https://github.com/paritytech/polkadot-sdk/issues/10751).

The section would fix: 
- The fold rule, binding dimension sets gas so it's a faithful upper bound, not an under pricing average;
- How gas price/limit map to the on chain fee and deposit; 
- What `eth_estimateGas` must return. Slots in next to your error-code and block-tag sections.

soloking1412:
> Hey @soloking1412, thanks for the RFC, solid proposal. Funny timing, I'd landed on the same idea and was about to send mine, so glad to fold it into yours instead of opening a competing PR.
> 
> One thing I'd add as a complementary section: the gas to resource metering model your conformance target relies on but doesn't pin down. `execution-apis`/go-ethereum assume single dimensional gas, but revive meters two dimensional `Weight` (`ref_time`+`proof_size`) plus a storage deposit. Without standardising how those fold into one gas number, `eth_estimateGas` diverges between revive chains and proof size heavy calls get mispriced, the same silent divergence your motivation flags (see [paritytech/polkadot-sdk#11525](https://github.com/paritytech/polkadot-sdk/issues/11525) and [paritytech/polkadot-sdk#10751](https://github.com/paritytech/polkadot-sdk/issues/10751)).
> 
> The section would fix:
> 
> * The fold rule, binding dimension sets gas so it's a faithful upper bound, not an under pricing average;
> * How gas price/limit map to the on chain fee and deposit;
> * What `eth_estimateGas` must return. Slots in next to your error-code and block-tag sections.

Thanks @Nathy-bajo, this is a great addition and honestly I'm glad you flagged it here instead of spinning up a separate PR. Happy to have you on as co-author. I've already added you to the authors list, just let me know the name you'd like next to the handle.

You're right that the metering fold was the gap. I went digging through the runtime to pin it down and your read matches the code: weight_to_fee already takes the binding dimension (the max of ref_time and proof_size once each is priced, with proof_size normalised into ref-time units via the block ratio). So I've written that up as the normative rule and explicitly ruled out the averaging path (weight_to_fee_average) for anything eth-facing, since that's the one that quietly underprices proof-size-heavy calls.

The deposit was the other half. On-chain the eth budget gets split into the weight fee and the storage deposit (storage_deposit = eth_fee - tx_fee), so I've specified that eth_estimateGas has to fold the deposit into the returned gas too, otherwise a tx sent with the estimate can still fail for funds even though it "had enough gas". That matches geth's contract that the estimate won't run out.

Added it as §2.4 right after the error-code and block-tag sections like you suggested, pulled in #11525 and #10751 as the divergence evidence, and dropped in a conformance case plus an open question on whether the fold should be a hard MUST given chains can set their own coefficients. Pushing the update now, give §2.4 a look and reword anything I got wrong on your side.

Nathy-bajo:
Thanks @soloking1412, co-author works, name is Nathaniel.

Checked §2.4 against the runtime and it matches: the fold is ref_time_fee.max(proof_size_fee)inBlockRatioFee::weight_to_fee, weight_to_fee_averageis the real(ref_time + proof_size)/2path to exclude, andstorage_deposit = eth_fee - tx_feelines up withcall.rs. Nice writeup.

One sentence worth adding to Requirements, ready to paste: "The gas value presented to the client MUST be divided by the current fee multiplier (the next_fee_multiplier_reciprocalinfees.rs), since the wallet re-applies that multiplier through eth_gasPrice; this keeps gas × gasPrice equal to the on-chain fee rather than double-counting or omitting the multiplier." Without pinning that direction an implementer can double-apply or drop the multiplier.

On the open question: make it a MUST, but scope it to the fold rule (binding dimension plus deposit), not the coefficients, which stay per-chain config.

soloking1412:
> Thanks @soloking1412, co-author works, name is Nathaniel.
> 
> Checked §2.4 against the runtime and it matches: the fold is ref_time_fee.max(proof_size_fee)inBlockRatioFee::weight_to_fee, weight_to_fee_averageis the real(ref_time + proof_size)/2path to exclude, andstorage_deposit = eth_fee - tx_feelines up withcall.rs. Nice writeup.
> 
> One sentence worth adding to Requirements, ready to paste: "The gas value presented to the client MUST be divided by the current fee multiplier (the next_fee_multiplier_reciprocalinfees.rs), since the wallet re-applies that multiplier through eth_gasPrice; this keeps gas × gasPrice equal to the on-chain fee rather than double-counting or omitting the multiplier." Without pinning that direction an implementer can double-apply or drop the multiplier.
> 
> On the open question: make it a MUST, but scope it to the fold rule (binding dimension plus deposit), not the coefficients, which stay per-chain config.

Perfect, thanks Nathaniel, added you as Nathaniel (@Nathy-bajo).

Good catch on the multiplier, that's a real footgun. I checked and next_fee_multiplier_reciprocal in fees.rs exists precisely for this, its doc comment even spells out the same reasoning (divide the fee by the multiplier before handing it back as gas, since the wallet re-applies it via eth_gasPrice). I've folded your sentence into the Requirements so an implementer can't double-apply or drop it.

Also took your call on the open question: the fold rule (binding dimension + deposit) is now a hard MUST, scoped to the rule itself, with the per-dimension coefficients left as per-chain config. Dropped the "gas fold normativity" question since it's settled now.

Nathy-bajo:
@soloking1412 `§2.4` looks great, all three points landed correctly. One loose end: the PR description still predates the gas section. Two quick additions to keep it in sync:

What this proposes, add (4): standardising the gas to resource metering fold (binding dimension plus storage deposit) so eth_estimateGas is a true upper bound and agrees across revive chains.

Why now, add a bullet: eth_estimateGas had no specified fold from revive's two-dimensional Weight plus storage deposit onto single-dimensional gas, so proof-size-heavy calls could be mispriced (https://github.com/paritytech/polkadot-sdk/issues/11525, https://github.com/paritytech/polkadot-sdk/issues/10751).

soloking1412:
> @soloking1412 `§2.4` looks great, all three points landed correctly. One loose end: the PR description still predates the gas section. Two quick additions to keep it in sync:
> 
> What this proposes, add (4): standardising the gas to resource metering fold (binding dimension plus storage deposit) so eth_estimateGas is a true upper bound and agrees across revive chains.
> 
> Why now, add a bullet: eth_estimateGas had no specified fold from revive's two-dimensional Weight plus storage deposit onto single-dimensional gas, so proof-size-heavy calls could be mispriced ([paritytech/polkadot-sdk#11525](https://github.com/paritytech/polkadot-sdk/issues/11525), [paritytech/polkadot-sdk#10751](https://github.com/paritytech/polkadot-sdk/issues/10751)).

Good eye, the description was stale. Updated it: added the gas fold as proposal #4 (true upper bound, agrees across revive chains) and a "why now" bullet for the unspecified Weight+deposit→gas fold with #11525/#10751. Also fixed the #12479 bullet that still said "errored" — it's the silent-zero-logs framing now, so the whole motivation list reads consistently. Thanks for keeping it in sync.

Nathy-bajo:
@soloking1412 Thanks too for listening to my requests. The proposal looks good to go now, and raising awareness of it in the Fellowship channel is not a bad idea.

PR #173, opened 2026-07-01
Add RFC-0173: Prevent Transaction Replay via CheckCreatedAtHeight

josepot:
Proposes adding a `created_at_height` field to per-account storage and a new `CheckCreatedAtHeight` transaction extension to cryptographically bind transactions to a specific account lifecycle epoch, preventing replay attacks after account reaping and making transaction hashes safe unique identifiers.

[Rendered](https://github.com/josepot/RFCs/blob/rfc-check-created-at-height/text/0173-check-created-at-height.md)

### DIFF
--- a/text/0173-check-created-at-height.md
+++ b/text/0173-check-created-at-height.md
@@ -10,10 +10,10 @@
 
 This RFC proposes two additions to prevent transaction replay attacks that arise from Polkadot's account reaping mechanism:
 
-1. A `created_at_height` field (`u32`, block height) added to per-account storage, set to the current block height whenever an account is created (or re-created), and defaulting to zero for genesis and pre-upgrade accounts.
+1. A `created_at_height` property (`u32`, block height) associated with every account, set to the current block height whenever an account is created or re-created, and exposed via a dedicated runtime API. This RFC defines the semantics and interface of this property without prescribing its storage implementation.
 2. A new `CheckCreatedAtHeight` transaction extension whose sole value is `created_at_height` carried as an **implicit** (additional-signed) field, mixed into the transaction signature without increasing the extrinsic's encoded size.
 
-Together these changes ensure that a transaction is cryptographically bound to a specific lifetime of its sender account. If the account is reaped and subsequently re-funded, `created_at_height` changes and all previously signed transactions become permanently invalid, eliminating replay attacks and simplifying transaction-tracking APIs for DApps and indexers.
+Together these changes ensure that a signed transaction is cryptographically bound to a specific lifetime of its sender account. If the account is reaped and subsequently re-funded, `created_at_height` changes and all previously signed transactions become permanently invalid, eliminating replay attacks. As a consequence, for signed and general transactions, the transaction hash becomes a safe unique identifier, and the triple `(account_id, created_at_height, nonce)` also becomes a reliable canonical identifier: both of which simplify transaction-tracking APIs for DApps and indexers.
 
 ## Motivation
 
@@ -82,18 +82,38 @@ The `CheckNonce` extension would then carry:
 
 This would have prevented the replay vulnerability from the start. Retrofitting this structure into `CheckNonce` is not feasible without breaking the existing signed extension interface. Instead, this RFC introduces a dedicated `CheckCreatedAtHeight` extension that achieves the same cryptographic binding without modifying `CheckNonce`.
 
-### Storage Change: `created_at_height`
+### `created_at_height`: Semantics and Interface
 
-A new `created_at_height` field of type `u32` MUST be added to per-account storage. The field SHALL:
+The `created_at_height` property is a `u32` value associated with every account. Its semantics are:
 
-- Be set to the **current block height** (`frame_system::Pallet::<T>::block_number()`) whenever an account transitions from non-existent to existent in `System::Account`.
-- Default to **zero** (`0u32`) for all accounts that exist prior to the runtime upgrade that enacts this change (see storage placement options below).
+- It MUST be set to the **current block height** whenever an account transitions from non-existent to existent (on both first creation and recreation after reaping).
+- It MUST default to **zero** (`0u32`) for all accounts that exist prior to the runtime upgrade enacting this change.
+- It MUST NOT change for any reason other than account recreation.
 
-When an account is reaped and subsequently re-funded, a new `AccountInfo` entry is written with `created_at_height` set to the block height at the time of recreation. This new value differs from the value in effect when any prior transactions were signed, making those transactions permanently invalid.
+When an account is reaped and subsequently re-funded, `created_at_height` takes the value of the block in which the account is recreated. This new value differs from the value in effect when any prior transactions were signed, making those transactions permanently invalid.
 
-#### Option A: New field in `frame_system::AccountInfo`
+#### Runtime API
 
-The field is added directly to `AccountInfo`, after `data`:
+Implementations MUST expose `created_at_height` to external consumers through a dedicated runtime API:
+
+```rust
+sp_api::decl_runtime_apis! {
+    pub trait AccountCreatedAtHeightApi<AccountId: codec::Codec> {
+        /// Returns the `created_at_height` for the given account, or `0` if the account does not exist.
+        fn created_at_height(account: AccountId) -> u32;
+    }
+}
+```
+
+Consumers MUST call this API to obtain the value before constructing the signing payload. The `CheckCreatedAtHeight` extension internally reads the same value via whatever storage mechanism the implementation chooses (see Implementation Considerations below).
+
+#### Implementation Considerations
+
+This RFC does not prescribe how `created_at_height` is stored; that is an implementation detail to be resolved at the time of implementation, ideally informed by benchmarks. Three approaches are outlined below for consideration.
+
+##### Option A: New field in `frame_system::AccountInfo`
+
+Add `created_at_height` directly to `AccountInfo`, e.g. after `data`:
 
 ```rust
 pub struct AccountInfo<Nonce, AccountData> {
@@ -106,23 +126,13 @@ pub struct AccountInfo<Nonce, AccountData> {
 }
 ```
 
-**Advantage:** Semantically correct. Account lifecycle is managed entirely by `frame_system`; the field that tracks the lifecycle epoch belongs there. Future readers of the code will find the field in the natural location.
+**Advantage:** Semantically correct. Account lifecycle is managed entirely by `frame_system`; the field that tracks the lifecycle epoch belongs there.
 
-**Disadvantage:** The SCALE encoding of `AccountInfo` changes, requiring a state migration. All existing `System::Account` entries must be migrated to the new layout (inserting four zero bytes after `data`). On Polkadot, this covers millions of accounts. To avoid excessive block weight in a single block, the migration SHOULD be implemented as a lazy/on-demand migration: old-format entries are transparently decoded on read, and entries are written in the new format on any subsequent update.
+**Disadvantage:** The SCALE encoding of `AccountInfo` changes, requiring a state migration over all existing `System::Account` entries (millions of accounts on Polkadot). A lazy/on-demand migration strategy (decode old entries on read, write new format on update) reduces peak weight but leaves the intermediate state in place until all accounts are touched.
 
-#### Option B: Reinterpret `ExtraFlags` in `pallet_balances::AccountData`
+##### Option B: Reinterpret `ExtraFlags` in `pallet_balances::AccountData`
 
-The `ExtraFlags` field in `pallet_balances::AccountData` is currently a `u128` newtype with only one bit in use, bit 127 (the MSB), which is the `new_logic` sentinel introduced when the holds/freezes balance model was adopted:
-
-```rust
-pub struct ExtraFlags(pub u128);
-
-impl ExtraFlags {
-    pub const NEW_LOGIC: ExtraFlags = ExtraFlags(1 << 127);
-}
-```
-
-All bits 0–126 are zero for every existing account. This RFC proposes reinterpreting the `u128` as a structured layout without changing any stored bytes. SCALE encodes integers in little-endian order, so the mapping is:
+The `ExtraFlags` field in `pallet_balances::AccountData` is a `u128` with only bit 127 in use (the `new_logic` sentinel). All bits 0–126 are zero for every existing account. The `u128` can be reinterpreted as a structured layout without changing any stored bytes (SCALE is little-endian):
 
 ```
 bits   0– 31  : created_at_height  (u32)  - this RFC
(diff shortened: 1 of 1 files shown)


leonardocustodio:
Question: Does this have an impact on offline-signers? Let's say I have a daemon that can't read the storage value. Is the field required?

josepot:
> Question: Does this have an impact on offline-signers? Let's say I have a daemon that can't read the storage value.

Any offline signer (whether a Ledger device, Polkadot Vault, or a custom daemon) must already communicate with some agent that can read chain state. There is no way around this: without access to on-chain data, it is impossible to construct a valid transaction in the first place.

Consider what every signer already needs to fetch before signing: the current spec version, the runtime metadata, the sender's current nonce, and the current block height and block hash to anchor transaction mortality. `created_at_height` is just one more field in that same set, and it is actually the least volatile of the bunch, since it only changes when the account is reaped and re-created. The `nonce` changes on every single transaction. Also, just in case that it ain't obvious, an offline signer must have a way to know whether the last transaction that they signed was broadcasted and included, and that info must be provided by an agent that's not offline. They can't simply assume that the latest nonce that they used is the current on-chain nonce, because if they are truly offline they don't even know whether the transaction was brodacasted and/or included (it could have been/become invalid).

So in practice, the agent sitting between the offline signer and the chain already performs these lookups; it simply needs to include `created_at_height` from System::Account alongside the nonce it already fetches. The impact on offline signers is negligible.

> Is the field required?

As the RFC explains, in order to prevent replay attacks, the signature should always include the `created_at_height`, yes.

Nathy-bajo:
Strong +1 on this RFC. 

Making tx hashes safe unique identifiers is exactly the invariant needed to give hash-based archive RPCs strong lookup guarantees. In [json-rpc-interface-spec #182](https://github.com/paritytech/json-rpc-interface-spec/pull/182) I'm proposing `archive_unstable_transactionReceipt` for locating transactions by hash without downloading full blocks; without `CheckCreatedAtHeight`, that lookup is ambiguous under replay after account reaping. This RFC would let the RPC (and any hash-indexed archive tooling) be specified with unambiguous, rather than best-effort, guarantees.

leonardocustodio:
> it is impossible to construct a valid transaction in the first place.

I don't agree with this sentence, it may be more "cumbersome" to build a valid transaction but it is far from impossible. You just use the immortal transactions, which are the whole reason for this issue and RFC.

Yes, you would still need to know the current specVersion, but once again, in an offline signer that's hardcoded, it is a single value to change. The rest, currentBlockHeight, is not necessary if you are using immortal transactions; nor blockHash nor era. Also, checkMetadataHash is optional.

Being `nonce` the only thing left to handle. Considering a very simple service like a signer that only signs "transfer" transactions from exchange hot wallets, which will hardly ever fail if you can at least track the balance, I consider the word "impossible" a heavy one.

Though this is a very specific use case and thus should probably not be the reason to shape the network improvements.

I also agree that this causes a lot of confusion. I've been there myself in the beginning, and until today, it's been very hard for people to understand that transactionHashes are not unique identifiers. In that sense, I'm also in favor of a change, and that's why the question.

josepot:
> You just use the immortal transactions, which are the whole reason for this issue and RFC.

I think this is an oversimplification. Please read the RFC carefully before jumping to conclusions 🙏. These replay attacks are [also possible with mortal transactions](https://github.com/josepot/RFCs/blob/rfc-check-created-at-height/text/0173-check-created-at-height.md#why-checkmortality-does-not-fully-solve-this).

> Yes, you would still need to know the current specVersion, but once again, in an offline signer that's hardcoded

The `specVersion` really should not be hardcoded in the offline signer. It should be provided by a non-offline agent. Otherwise, that hardcoded value would need to be updated on every runtime upgrade, which is far from ideal.

> Also, checkMetadataHash is optional.

Sure, but it was created precisely to address the needs of offline signers, so they can be sure that they are safely encoding/decoding data according to the current runtime codecs declared in the metadata.

> Being `nonce` the only thing left to handle. Considering a very simple service like a signer that only signs "transfer" transactions from exchange hot wallets, which will hardly ever fail if you can at least track the balance

If you can track the balance, then we agree that there is an online agent assisting the offline signer and providing it with information. That makes this point somewhat moot, because `created_at_height` changes much less frequently than the `nonce`, right?

> Though this is a very specific use case and thus should probably not be the reason to shape the network improvements.

Indeed! That said, the RFC also acknowledges that this is the [primary integration burden](https://github.com/josepot/RFCs/blob/rfc-check-created-at-height/text/0173-check-created-at-height.md#stakeholders).

voliva:
I'm +1 as well, I think it's a nice addition that covers an important gap.

The only thing I'd like to point out is that although this will ensure signed/general transactions have unique hashes, this won't happen for unsigned transactions.

For instance, `Timestamp::set` is currently being added on every block with repeated tx hashes, as it only changes the actual timestamp every 6s. There are others in multi-block election, babe, etc. but I'm not sure if having them repeated happens often (e.g. `System::apply_authorized_upgrade` is also unsigned but realistically it would be submitted just once per unique payload)

I still think this shouldn't be a blocker, as it's still a nice addition, but it might be worth keeping in mind unsigned transaction cases for other improvements.

bkchr:
This proposes the same as https://github.com/polkadot-fellows/runtimes/issues/248, just in a different way.

josepot:
> The only thing I'd like to point out is that although this will ensure signed/general transactions have unique hashes, this won't happen for unsigned transactions.
> 
> For instance, `Timestamp::set` is currently being added on every block with repeated tx hashes, as it only changes the actual timestamp every 6s. There are others in multi-block election, babe, etc. but I'm not sure if having them repeated happens often (e.g. `System::apply_authorized_upgrade` is also unsigned but realistically it would be submitted just once per unique payload)

This is a great point, and you are absolutely right: the uniqueness guarantee only applies to signed/general transactions, not to unsigned/bare ones. I will update the RFC to make this distinction explicit. That said, I want to note that the `BlockBuilder_apply_extrinsic` hash-deduplication requirement described in the RFC already ensures that the same extrinsic cannot appear more than once within a single block regardless of whether it is signed or unsigned, which is a useful property in its own right. I will elaborate on this when I update the RFC. Thanks for flagging this, though! 🙏

> This proposes the same as [polkadot-fellows/runtimes#248](https://github.com/polkadot-fellows/runtimes/issues/248), just in a different way.

Thanks for the pointer! I hadn't come across that discussion before. The approach there is related, but there are a few reasons I favour the approach in this RFC

1. It doesn't fully prevent replay attacks. Consider an account created at block M that makes K transactions (nonces M through M+K-1), is then reaped, and re-created at block N where M < N < M+K. The new stored nonce is N, and the old transaction signed during the first lifetime with nonce N becomes immediately replayable. A high-throughput account (an exchange hot wallet, a DeFi protocol) can easily accumulate more transactions than blocks have passed since its creation. This RFC's `created_at_height` value is independent of the nonce, so all transactions from a prior lifetime are unconditionally invalidated regardless of how many were made.

2. Transaction size increases the later an account is created. The nonce is part of the explicit extrinsic payload. SCALE compact-encodes integers, so a nonce starting near zero fits in one byte, but a nonce starting at the current block height (over 20 million on Polkadot) requires four bytes. Every transaction an account ever makes would carry that overhead permanently, and the penalty grows the later the account is created, which seems like an undesirable property to bake into the protocol.

3. It conflates two distinct concepts. The nonce is a transaction-sequencing mechanism. Account lifecycle epoch is a separate concern. Encoding both into the nonce makes the semantics of the field harder to reason about and document, and closes off the space for future optimisations to either concept independently.

--------

Regarding the storage implementation options (I believe a comment on this was posted and then deleted?, but the question is worth addressing publicly anyway).

I've been thinking more about the trade-offs between the options described in the RFC and a third one I didn't initially include: Option C, a dedicated storage map, e.g. `System.AccountCreatedAt(AccountId32)` with a `Blake2_128Concat` hasher. I originally dismissed this as wasteful compared to repurposing the unused bits in `ExtraFlags`, but on reflection it has an appealing property: it requires no migration at all (absent entries simply read as 0), avoids touching either `AccountInfo` or `pallet_balances::AccountData`, and keeps the concern squarely in `frame_system`.

More broadly though, I am starting to question whether the RFC should prescribe a specific storage implementation at all. What matters for interoperability is the interface: that a `created_at_height` value is available per account via a runtime-api, that it is exposed as an implicit value via `CheckCreatedAtHeight`, and the rules governing when it changes. The storage choice is an implementation detail better settled with actual benchmarks once the interface is agreed upon. I plan to rework that section of the RFC accordingly.

Kanasjnr:
> Regarding the storage implementation options (I believe a comment on this was posted and then deleted?, but the question is worth addressing publicly anyway).
> 
Nice RFC, the implicit extension approach makes sense to me.

On the open questions
I'd go with Option A. Feels weird to put account lifecycle stuff inside ExtraFlags in balances, and Option B only really works if you're on pallet_balances. Migration is annoying but lazy decode seems fine?

For migration I'm assuming lazy is enough and you don't need to touch every account at upgrade. Curious if you're thinking background migration in on_idle for Polkadot or just leave dormant accounts on the old encoding forever.

Re the non-existent account returning 0 I think this is ok because the actual attack needs a re-fund which changes the height anyway. The reaped-but-not-funded window might still allow a sig match on old txs but there's nothing to steal? Would be good to spell that out in the RFC though, I'm not 100% sure without tracing the runtime.

Also +1 to Leonardo's offline signer question my read is the field is required (it's implicit signed data), so something online has to pass the value in. i think its worth mentioning explicitly or not?

@josepot here was my question idk i edited it but it got deleted idk 🤷

josepot:
I just updated the RFC with the received feedback, the main changes that I made are:

- Define a runtime API (`AccountCreatedAtHeightApi`) as the interface for accessing c`reated_at_height`, decoupling the RFC from any specific storage implementation.
- Restructure storage options (A, B, and new Option C) as implementation considerations rather than prescriptions.
- Clarify that the tx-hash uniqueness guarantee applies to signed/general transactions only; unsigned extrinsics are explicitly out of scope.
- Add [runtimes#248](https://github.com/polkadot-fellows/runtimes/issues/248) to Prior Art with a full tradeoff analysis.
- Update Drawbacks and Unresolved Questions to match the new framing.

Looking forward to further feedback! 🙂

text/0173-check-created-at-height.md
```diff
@@ -0,0 +1,284 @@
...
+
+### Privacy
+
+This proposal has no privacy implications. `created_at_height` is a deterministic, publicly observable property of account state and reveals no information beyond what is already available on-chain.
+
+## Performance, Ergonomics, and Compatibility
+
+### Performance
+
+`CheckCreatedAtHeight` requires one additional field access per transaction validation. In practice, `System::Account` is already fetched by `CheckNonce` during the same validation pass, so the marginal cost is a single field read on a value already in the storage cache. The overhead is negligible.
```
PolkadotDom, line 245: depends on implementation, option c for instance will be an extra read.

josepot: Indeed! It's mentioned in the "disadvantage" section of Option c.

text/0173-check-created-at-height.md
```diff
@@ -0,0 +1,284 @@
...
+### New Extension: `CheckCreatedAtHeight`
+
+A new transaction extension `CheckCreatedAtHeight` is introduced. It carries no **explicit** data (extrinsic size is unchanged) and carries `created_at_height` as a single **implicit** value:
+
+```
+Implicit = u32   (the sender's current created_at_height from storage)
+Explicit = ()    (empty, no bytes added to the extrinsic)
+```
+
+In FRAME's `TransactionExtension` trait, the implicit value is gathered at validation time by reading the sender's `created_at_height` from storage and is added into the signing payload alongside other implicit values (spec version, genesis hash, etc.). The signer MUST read the same value from storage at signing time and include it in their signed payload. If the stored value has changed since the transaction was signed, because the account was reaped and re-created, signature verification fails automatically. No explicit comparison is required in `validate()`.
```
PolkadotDom, line 170: Perhaps missing something, but how does the account data get passed into `implicit(&self)` so as to derive the `created_at_height` value?

text/0173-check-created-at-height.md
```diff
@@ -0,0 +1,284 @@
...
+  - **It conflates two distinct concepts.** The nonce is a transaction-sequencing mechanism; account lifecycle epoch is a separate concern. Encoding both into the nonce complicates reasoning about either in isolation.
+- **[polkadot-fellows/RFCs#19](https://github.com/polkadot-fellows/RFCs/issues/19)**: Discussion on light clients and downloading block bodies. Proposals in that thread (e.g. a `System.Extrinsics` per-block `StorageMap<Hash, ExtrinsicIndex>`) are complicated by the non-uniqueness of transaction hashes. This RFC resolves that precondition for signed/general transactions.
+- **[paritytech/json-rpc-interface-spec#182](https://github.com/paritytech/json-rpc-interface-spec/pull/182)**: Proposal for an `archive_unstable_transactionReceipt` JSON-RPC method. Reviewers flagged transaction hash non-uniqueness as a design obstacle; this RFC removes it for signed/general transactions.
+
+## Unresolved Questions
+
+1. **Behaviour for non-existent accounts**: `implicit()` returns `0u32` when the signer's account does not exist. Is there a scenario where this could allow a transaction to pass validation against a reaped, not-yet-recreated account?
```
PolkadotDom, line 274: For signed transactions, `validate_nonce_for_account` and fee charging would keep this from being an issue. As in accounts that aren't created can't produce valid transactions.

Perhaps worth noting the property is dependent on those checks then, and not guaranteed from this RFC itself. Also `created_at_height` needs to be set in frame-systems account creation path, not in pallet-balances, which may disqualify option B.

text/0173-check-created-at-height.md
```diff
@@ -0,0 +1,284 @@
...
+| Mortal transaction replayed within mortality window after reaping and re-funding | **Valid** | **Invalid** (`created_at_height` changed) |
+| Same transaction included twice in one block after intra-block reap and re-fund, account created in a prior block | **Valid** (nonce reset) | **Invalid** (`created_at_height` changed) |
+| Same transaction included twice in one block after intra-block reap and re-fund, account first created in the same block | **Valid** (nonce reset) | **Invalid** (runtime-level tx-hash deduplication; see below) |
+| Same transaction included twice in one block, no reaping | **Invalid** (nonce check) | **Invalid** (nonce check, unchanged) |
+
+#### Intra-Block Replay
+
+For the common case, `CheckCreatedAtHeight` fully covers intra-block replay. If a transaction T1 was signed with `created_at_height = M` (the block in which the sender's account was originally created, where M < N) and within block N the account is reaped and re-funded, the new `created_at_height` is set to N. Any attempt to re-include T1 fails immediately because the stored value N does not match the signed value M.
```
PolkadotDom, line 194: global account counter instead of `created_at_height` might solve this with less collateral damage (the deduping in apply extrinsic). Though may have to be u64 which is not ideal.

text/0173-check-created-at-height.md
```diff
@@ -0,0 +1,284 @@
...
+        fn created_at_height(account: AccountId) -> u32;
+    }
+}
+```
+
+Consumers MUST call this API to obtain the value before constructing the signing payload. The `CheckCreatedAtHeight` extension internally reads the same value via whatever storage mechanism the implementation chooses (see Implementation Considerations below).
+
+#### Implementation Considerations
+
+This RFC does not prescribe how `created_at_height` is stored; that is an implementation detail to be resolved at the time of implementation, ideally informed by benchmarks. Three approaches are outlined below for consideration.
```
PolkadotDom, line 112: I feel anything not committed to should be left out of an RFC (the options presented). Would be better to discuss in a thread then commit to it in the RFC.

text/0173-check-created-at-height.md
PolkadotDom: Yes I just mean you can't assess the performance confidently if we don't know the direction yet.

PolkadotDom:
Great read and necessary. Thanks Josep 🙌

